PR #1, opened 2020-03-12
Add cluster join related APIs

qiujian16:


clusters/v1/types.go
```diff
@@ -0,0 +1,175 @@
...
+	// Conditions applied to the request, such as approval, denial or pending.
+	Conditions []CLusterJoinRequestConditions `json:"conditions"`
+	// If request was approved, the controller will place the issued certificate here.
+	Certificate []byte `json:"certificate,omitempty"`
+}
+
+// CLusterJoinRequestConditions contains the condition info for the cluster join request
+type CLusterJoinRequestConditions struct {
+	// request approval state, currently Approved or Denied.
+	Phase ClusterJoinRequestPhase `json:"phase"`
```
qiujian16: type field and status field

clusters/v1/types.go
```diff
@@ -0,0 +1,175 @@
...
+
+	// Status represents the current status of joined spoke cluster
+	Status JoinedClusterStatus `json:"status,omitempty"`
+}
+
+// JoinedClusterSpec represents a desired configuration for the agents on the spoke cluster.
+type JoinedClusterSpec struct {
+	// AgentConfigurations is a key-value pairs that hub intends to set on agents on the
+	// spoke cluster
+	AgentConfigurations map[string]string `json:"agentConfigurations,omitempty"`
```
qiujian16: what if we have customized agents on spoke clusters?

clusters/install.go
```diff
@@ -0,0 +1,26 @@
...
+
+import (
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+
+	clustersv1 "github.com/open-cluster-management/api/clusters/v1"
+)
+
+const (
+	GroupName = "clusters.open-cluster-management.io"
```
deads2k: Don't add more types, but can you outline the types you see as existing in this group one year from now?

deads2k: openshift prefers singular, so `cluster.open-cluster-management.io`.

@pmorie @mdelder have already purchased this domain?  Versus `openclustermanagement` for instance.  Wouldn't want the wrong one.

clusters/v1/register.go
```diff
@@ -0,0 +1,28 @@
...
+	GroupVersion  = schema.GroupVersion{Group: GroupName, Version: "v1"}
+	schemeBuilder = runtime.NewSchemeBuilder(addKnownTypes, corev1.AddToScheme)
+	// Install is a function which adds this version to a scheme
+	Install = schemeBuilder.AddToScheme
+)
+
+// Adds the list of known types to api.Scheme.
+func addKnownTypes(scheme *runtime.Scheme) error {
+	scheme.AddKnownTypes(GroupVersion,
+		&JoinedCluster{},
```
deads2k: I expected to see one `JoinedCluster` resource that would be created when a client wanted to add a cluster to the hub. 
 That resource would then have status conditions that indicated whether or not the request was honored.  Was there something specific that drove two different resources?

It seems like most (all?) permissions issues with regions of spec could be handled via admission and would keep approvals co-located (CSR as a clear example).

clusters/v1/types.go
```diff
@@ -0,0 +1,175 @@
...
+	// Spec represents a desired configuration for the agent on the spoke cluster.
+	Spec JoinedClusterSpec `json:"spec"`
+
+	// Status represents the current status of joined spoke cluster
+	Status JoinedClusterStatus `json:"status,omitempty"`
+}
+
+// JoinedClusterSpec represents a desired configuration for the agents on the spoke cluster.
+type JoinedClusterSpec struct {
+	// AgentConfigurations is a key-value pairs that hub intends to set on agents on the
```
deads2k: This comment isn't super helpful. Is this something for a user-agent?  Something else?

clusters/v1/types.go
```diff
@@ -0,0 +1,175 @@
...
+	// spoke cluster
+	AgentConfigurations map[string]string `json:"agentConfigurations,omitempty"`
+}
+
+// JoinedClusterStatus represents the current status of joined spoke cluster
+type JoinedClusterStatus struct {
+	// Conditions contains the different condition statuses for this spoke cluster.
+	Conditions []ClusterCondition `json:"conditions"`
```
deads2k: Go ahead and create a real conditions type.  We have been pretty happy with https://github.com/openshift/api/blob/master/config/v1/types_cluster_operator.go#L109-L136 .  I would make the Type a string and require Reason, but otherwise happy.

clusters/v1/types.go
```diff
@@ -0,0 +1,175 @@
...
+
+	// Capacity represents the total resource capacity on the spoke cluster
```
deads2k: Indicate what this is the capacity of. Is this the total of all Capacity on all NodeStatuses?

clusters/v1/types.go
```diff
@@ -0,0 +1,175 @@
...
+	Capacity corev1.ResourceList `json:"capacity,omitempty"`
+
+	// Usage represents the total resource request consumed on the spoke cluster
+	Usage corev1.ResourceList `json:"usage,omitempty"`
```
deads2k: Indicate what this is the capacity of. Is this the total of all Capacity minus all Allocateable on all NodeStatuses?  If so, why invert the usage instead of indicating the Allocatable like a node status does?  The two are not exactly equivalent because of reservations and overcommit.

clusters/v1/types.go
```diff
@@ -0,0 +1,175 @@
...
+
+	// MasterAddress represents the apiserver URL on the spoke cluster
+	MasterAddress string `json:"masterAddress,omitempty"`
+
+	// Version represents the kubernetes version of the spoke cluster
+	Version string `json:"version,omitempty"`
+
+	// AgentVersions represents the versions of each agent in key-value pair
```
deads2k: What is an agent?  I would be very careful about using this resource to report status on logical subcomponents that would be installed.  Instead I would try to figure out what leads to having "special" sets of maintained resources instead of a common representation of "this unit will be present in the spoke" resources re-used by everyone.

clusters/v1/types.go
```diff
@@ -0,0 +1,175 @@
...
+type ClusterCondition struct {
+	// Type is the type of the cluster condition.
+	Type ClusterConditionType `json:"type"`
+
+	// Status is the status of the condition. One of True, False, Unknown.
+	Status corev1.ConditionStatus `json:"status"`
+
+	// LastHeartbeatTime is the last time this condition was updated.
+	// +optional
+	LastHeartbeatTime metav1.Time `json:"lastHeartbeatTime,omitempty"`
```
deads2k: heartbeats have proven too expensive for write load. We moved away from them upstream.

clusters/v1/types.go
```diff
@@ -0,0 +1,175 @@
...
+	ClusterOK ClusterConditionType = "OK"
+)
+
+// ClusterCondition contains condition information for a spoke cluster.
+type ClusterCondition struct {
+	// Type is the type of the cluster condition.
+	Type ClusterConditionType `json:"type"`
+
+	// Status is the status of the condition. One of True, False, Unknown.
+	Status corev1.ConditionStatus `json:"status"`
```
deads2k: don't re-use corev1 for this.

clusters/v1/types.go
deads2k: Prefer avoiding a corev1 dep and instead depend on the generic `api/resource` and create your own compatible ResourceList concept.

clusters/v1/types.go
```diff
@@ -0,0 +1,175 @@
...
+	Spec ClusterJoinRequestSpec `json:"spec"`
+
+	// Status reprensents the current status of the cluster join request
+	Status ClusterJoinRequestStatus `json:"status,omitempty"`
+}
+
+// ClusterJoinRequestSpec defines the request information to join hub
+type ClusterJoinRequestSpec struct {
+	// ClusterName is the name of the cluster
+	ClusterName string `json:"clusterName"`
```
deads2k: I expected to see....
1. a way for the spoke cluster to opt-in to being joined to the hub
2. a way for the hub to opt-in to allowing the spoke
3. status reflecting both.

A double opt-in ensures that nothing is force-added incorrectly and requests can originate from any source.  It would also make it possible to name the `ClusterJoinRequest` based on the name to avoid any duplicates and keep it as a record of who is present.

clusters/v1/types.go
```diff
@@ -0,0 +1,175 @@
...
+
+	// Base64-encoded PKCS#10 CSR data for certificate
+	Request []byte `json:"request"`
```
deads2k: Which actor handles this and why?

clusters/v1/types.go
```diff
@@ -0,0 +1,175 @@
...
+	// Request is pending
+	JoinPending ClusterJoinRequestPhase = "Pending"
+	// Request is approved and certficate is provided
+	JoinSucceeded ClusterJoinRequestPhase = "Succeeded"
+)
+
+// ClusterJoinRequestStatus defines current status of the cluster join request
+type ClusterJoinRequestStatus struct {
+	// Conditions applied to the request, such as approval, denial or pending.
+	Conditions []CLusterJoinRequestConditions `json:"conditions"`
```
deads2k: I would use a common condition type.  I've been meaning to make one in metav1 actually.

clusters/v1/types.go
```diff
@@ -0,0 +1,175 @@
...
+	// ClusterName is the name of the cluster
+	ClusterName string `json:"clusterName"`
+
+	// Base64-encoded PKCS#10 CSR data for certificate
+	Request []byte `json:"request"`
+}
+
+// ClusterJoinRequestPhase represents the current phase the cluster join request
+// is on
+type ClusterJoinRequestPhase string
```
deads2k: no phases please.  They don't represent the multiplicity of possible states well without excessive fanout.

clusters/v1/types.go
qiujian16: The agent sends this to the hub, then hub uses this to create a csr. The reason is we want agent asks for a client certificate and the hub return the certificate if the clusterjoinrequest is approved.

clusters/v1/types.go
qiujian16: The current process is agent sends a clusterjoinrequest to the hub. The hub checks if the cluster name is valid, then auto approve the request. 

Could you clarify the process that originate from hub. And the name of clusterjoinrequest is a digest of clustername + its privatekey.

clusters/v1/types.go
qiujian16: We may need the heartbeat to know it the cluster is still alive....How about having a stateless aggregated apiserver to provide custom subresources for heartbeat? any suggestions?

clusters/install.go
mdelder: We own both `open-cluster-management.io` and `open-cluster-mgmt.io` for at least the next 3 years.

clusters/v1/types.go
qiujian16: agent is a component installed as subcomponent in klusterlet. It can be search agent, topology agent or endpoint operator...We used to have a show the version of work manager and endpoint operator in cluster status. But yes..I think having another API to reflect what is maintained in the spoke cluster should be better..

clusters/v1/types.go
qiujian16: Yes, it should be Allocatable

clusters/v1/types.go
qiujian16: yes,  it is all capacity added on node statuses. I will update the comment.

clusters/v1/types.go
qiujian16: This is something similar as agent version. This is some configuration the hub want to put onto the spoke cluster, so agents on the spoke cluster can be correctly configured.

clusters/v1/register.go
qiujian16: In my mind, when a client wants to add a cluster, it needs to create a clusterjoinrequest. Then it needs to put the private key for the clusterjoinrequest to the spoke cluster. JoinedCluster is to reflect the current status of the cluster that has been joined. It can be created by user or controller.

Do you expect that joinedcluster and clusterjoinrequest is merged as one API? So its status will be like join pending -> join approved -> cluster online ?

clusters/v1/types.go
deads2k: > We may need the heartbeat to know it the cluster is still alive....How about having a stateless aggregated apiserver to provide custom subresources for heartbeat? any suggestions?

I would not couple it to a condition.  If we need to add it in the future, we can add a different field, but we know that this scales poorly for conditions.

deads2k:
Spoke on slack and bluejeans with @qiujian16.  The bits around the CSR flow are going to be redescribed with an eye towards how we handle ongoing CSR renewal (new certs) on a timeframe similar to kubelets (once a month).  This is likely to impact how the API is built and may end up leveraging CSRs directly and using approvals to control access.

clusters/v1/register.go
```diff
@@ -0,0 +1,26 @@
...
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+)
+
+var (
+	GroupName     = "cluster.open-cluster-management.io"
+	GroupVersion  = schema.GroupVersion{Group: GroupName, Version: "v1"}
+	schemeBuilder = runtime.NewSchemeBuilder(addKnownTypes, corev1.AddToScheme)
```
deads2k: remove corev1.  You no longer depend on it.

clusters/v1/register.go
deads2k: > Do you expect that joinedcluster and clusterjoinrequest is merged as one API? So its status will be like join pending -> join approved -> cluster online ?

Yes, via conditions.  Looks like this is probably done.

clusters/v1/types.go
deads2k: > Go ahead and create a real conditions type. We have been pretty happy with https://github.com/openshift/api/blob/master/config/v1/types_cluster_operator.go#L109-L136 . I would make the Type a string and require Reason, but otherwise happy.

This is still outstanding

clusters/v1/types.go
```diff
@@ -0,0 +1,114 @@
...
+	ResourceStorage ResourceName = "storage"
+	// Node, in numbers
+	ResourceNodes = "node"
+)
+
+// ResourceList defines a map for the quantity of different resources
+type ResourceList map[ResourceName]resource.Quantity
+
+// ClusterCondition contains condition information for a spoke cluster.
+type ClusterCondition struct {
```
deads2k: > Go ahead and create a real conditions type. We have been pretty happy with https://github.com/openshift/api/blob/master/config/v1/types_cluster_operator.go#L109-L136 . I would make the Type a string and require Reason, but otherwise happy.

This is still outstanding

clusters/v1/types.go
```diff
@@ -0,0 +1,114 @@
...
+
+	// Capacity represents the total resource capacity from all nodeStatuses
+	// on the spoke cluster
+	Capacity ResourceList `json:"capacity,omitempty"`
+
+	// Allocatable represents the total allocatable resources on the spoke cluster
+	Allocatable ResourceList `json:"allocatable,omitempty"`
+
+	// MasterAddress represents the apiserver URL on the spoke cluster
+	MasterAddress string `json:"masterAddress,omitempty"`
```
deads2k: This is spec information. You would be provided this by the cluster when it registered.

It should also be roughly the shape of a kubeconfig, with a URL and a CA bundle (empty means use system).

clusters/v1/types.go
```diff
@@ -0,0 +1,114 @@
...
+
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// JoinedCluster represents the current status of spoke cluster.
+type JoinedCluster struct {
+	metav1.TypeMeta `json:",inline"`
+	// Standard object's metadata.
+	// More info: https://git.k8s.io/community/contributors/devel/api-conventions.md#metadata
+	// +optional
+	metav1.ObjectMeta `json:"metadata,omitempty"`
```
deads2k: Describe what, if any, restrictions are on the name.

clusters/v1/types.go
```diff
@@ -0,0 +1,114 @@
+package v1
+
+import (
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// JoinedCluster represents the current status of spoke cluster.
```
deads2k: indicate cluster scoped or namespace scoped.  I suspect cluster scoped.

clusters/v1/types.go
```diff
@@ -0,0 +1,114 @@
...
+
+	// Spec represents a desired configuration for the agent on the spoke cluster.
+	Spec JoinedClusterSpec `json:"spec"`
+
+	// Status represents the current status of joined spoke cluster
+	Status JoinedClusterStatus `json:"status,omitempty"`
+}
+
+// JoinedClusterSpec represents a desired configuration for the agents on the spoke cluster.
+type JoinedClusterSpec struct {
```
deads2k: I expected a field here where the hub cluster-admin could choose to accept the spoke cluster.

clusters/v1/types.go
```diff
@@ -0,0 +1,114 @@
...
+
+import (
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// JoinedCluster represents the current status of spoke cluster.
+type JoinedCluster struct {
```
deads2k: This object looks like a `Spoke` to me or a `SpokeCluster`.

clusters/v1/types.go
```diff
@@ -0,0 +1,114 @@
...
+	// ClusterOK means that the cluster is "OK".
+	ClusterOK ClusterConditionType = "ClusterOK"
+	// ClusterJoined means the spoke cluster has successfully joined the hub
+	ClusterJoined ClusterConditionType = "ClusterJoined"
+	// Request to join the cluster is approved by user or controller
+	ClusterJoinApproved ClusterConditionType = "JoinApproved"
+	// Request to join the cluster is denied by user or controller
+	ClusterJoinDenied ClusterConditionType = "JoinDenied"
+	// Request to join the cluster is pending
+	ClusterJoinPending ClusterConditionType = "JoinPending"
```
deads2k: unnecessary.  this !JoinApproved && !JoinDenied.

clusters/v1/types.go
```diff
@@ -0,0 +1,114 @@
...
+// ClusterConditionType marks the kind of cluster condition being reported.
+type ClusterConditionType string
+
+const (
+	// ClusterOK means that the cluster is "OK".
+	ClusterOK ClusterConditionType = "ClusterOK"
+	// ClusterJoined means the spoke cluster has successfully joined the hub
+	ClusterJoined ClusterConditionType = "ClusterJoined"
+	// Request to join the cluster is approved by user or controller
+	ClusterJoinApproved ClusterConditionType = "JoinApproved"
```
deads2k: These should be phrased in terms of who approved.  I think this is `HubApprovedJoin`

clusters/v1/types.go
```diff
@@ -0,0 +1,114 @@
...
+	ResourceCPU ResourceName = "cpu"
+	// Memory, in bytes. (500Gi = 500GiB = 500 * 1024 * 1024 * 1024)
+	ResourceMemory ResourceName = "memory"
+	// Volume size, in bytes (e,g. 5Gi = 5GiB = 5 * 1024 * 1024 * 1024)
+	ResourceStorage ResourceName = "storage"
+	// Node, in numbers
+	ResourceNodes = "node"
+)
+
+// ResourceList defines a map for the quantity of different resources
```
deads2k: Indicate that this matches the corev1 and link to it.  It allows consumers to duck-type, without introducing an API we don't control.

docs/clusterjoinprocess.md
```diff
@@ -0,0 +1,49 @@
...
+
+Actors:
+1. cluster-admin on spoke cluster
+2. cluster-admin on hub cluster
+3. hub controller
+4. agent on spoke cluster
+
+## Cluster join initiated from spoke cluster
+1. cluster-admin on spoke cluster create bootstrap kubeconfig (connects to hub cluster) as secret and deploy the agent on spoke cluster. Cluster-admin on spoke needs to set cluster name/namespace when deploying the agent on spoke cluster.
+2. agent on spoke cluster creates CSR on hub cluster using bootstrap kubeconfig. The subject in csr is `{"Orgnization": ["open-cluster-management:cluster"], "CommonName":"open-cluster-management:clusterNamespace:clusterName"}`. The name of the csr is the digest of subject and private key, with a common prefix.
```
deads2k: Organization

docs/clusterjoinprocess.md
```diff
@@ -0,0 +1,49 @@
...
+This is to describe the various process of cluster join
+
+Actors:
+1. cluster-admin on spoke cluster
+2. cluster-admin on hub cluster
+3. hub controller
+4. agent on spoke cluster
+
+## Cluster join initiated from spoke cluster
+1. cluster-admin on spoke cluster create bootstrap kubeconfig (connects to hub cluster) as secret and deploy the agent on spoke cluster. Cluster-admin on spoke needs to set cluster name/namespace when deploying the agent on spoke cluster.
```
deads2k: break every sentence or 120ish chars, whichever is shorter so we can comment.

docs/clusterjoinprocess.md
deads2k: I think you should `system:open-cluster-management:` prefix all your names.

deads2k: Why is a `clusterNamespace` configurable and if it is configurable, how is it known on the spoke cluster at this point?

deads2k: I think you want this to identify user as the spoke agent, not the spoke cluster.  If you only identify the spoke cluster, you'll never be able to safely register the spoke multiple times.  The Organization (which maps to group), can reasonably match the spoke cluster

deads2k: I don't think the name needs to be known ahead of time.

deads2k: indicate that the CSR will specify the signer name as the kube-client one.

docs/clusterjoinprocess.md
```diff
@@ -0,0 +1,49 @@
...
+5. hub controller updates condition of `JoinedCluster` to `JoinApproved`.
+6. hub-controller creates role & rolebinding for the agent on spoke cluster.
+7. agent on spoke cluster gets certificate in CSR status, uses the certificate to create a new kubeconfig and saves it as secret.
+8. agent on spoke cluster connects to hub apiserver using the new kubeconfig.
+9. agent on spoke cluster updates conditions of `JoinedCluster` as `ClusterJoined`.
+10. agent on spoke cluster appends conditions of `JoineCluster` as `OK` and updates other fields in status of `JoinedCluster`.
+
+## Certificate renewal
+1. agent on spoke cluster detects the certificate is going to be expired.
+2. agent on spoke cluster generates a new private key and submits a new CSR to hub apiserver.
```
deads2k, line 47: specify the identity that will be used.

docs/clusterjoinprocess.md
```diff
@@ -0,0 +1,49 @@
...
+3. hub controller auto approves the csr.
```
deads2k: indicate the check which is done to decide how to do this.  Also indicate how you would eliminate the "infinite renewal" problem.

docs/clusterjoinprocess.md
```diff
@@ -0,0 +1,49 @@
...
+1. cluster-admin on spoke cluster
+2. cluster-admin on hub cluster
+3. hub controller
+4. agent on spoke cluster
+
+## Cluster join initiated from spoke cluster
+1. cluster-admin on spoke cluster create bootstrap kubeconfig (connects to hub cluster) as secret and deploy the agent on spoke cluster. Cluster-admin on spoke needs to set cluster name/namespace when deploying the agent on spoke cluster.
+2. agent on spoke cluster creates CSR on hub cluster using bootstrap kubeconfig. The subject in csr is `{"Orgnization": ["open-cluster-management:cluster"], "CommonName":"open-cluster-management:clusterNamespace:clusterName"}`. The name of the csr is the digest of subject and private key, with a common prefix.
+3. cluster-admin on hub-cluster approve the CSR.
+4. hub-controller creates `JoinedCluster` object on hub cluster and updates its condition to `JoinApproved`. The name of the object is cluster name, and its namespace is cluster namespace.
```
deads2k: This scenario is titled as "cluster join initiated by spoke".  I expected to see the JoinedCluster object created by the spoke.  How did the spoke indicate to the hub controller that it wanted creation?

deads2k: I'm surprised this is namespace scoped.  What forces made this namespace scoped instead of choosing to wire it cluster scoped to avoid collisions and use an admission webhook to prevent unexpected status updates or creating a namespaced resource that the spoke updates and a controller to mirror it the cluster scoped resource.  this appears to create a bootstrapping problem where the a namespace can be created before a double opt-in is completed.

docs/clusterjoinprocess.md
```diff
@@ -0,0 +1,49 @@
...
+5. hub-controller creates role & rolebinding for the agent on spoke cluster.
```
deads2k: which cluster is the role and rolebinding created in?  I think you're creating a role and rolebinding on the hub cluster that grants permission to the cluster namespace to the spoke agent.

docs/clusterjoinprocess.md
```diff
@@ -0,0 +1,49 @@
...
+6. agent on spoke cluster gets certificate in CSR status, uses the certificate to create a new kubeconfig and saves it as secret.
+7. agent on spoke cluster connects to hub apiserver using the new kubeconfig.
+8. agent on spoke cluster updates conditions of `JoinedCluster` as `ClusterJoined`.
+9. agent on spoke cluster appends conditions of `JoineCluster` as `OK` and updates other fields in status of `JoinedCluster`.
+
+## Cluster join initiated from hub
+1. cluster-admin on hub creates a `JoinedCluster` object, the condition of the object is `JoinPending`
+2. cluster-admin on spoke cluster deploys the agent with bootstrap kubeconfig, the name/namespace need to be mapped to `JoinedCluster` created on hub.
+3. agent on spoke cluster creates CSR on hub apiserver
+4. hub controller approves the csr
```
deads2k: how does the hub controller safely decide to approve this CSR without human intervention?

docs/clusterjoinprocess.md
```diff
@@ -0,0 +1,49 @@
...
+## Auto approve
+We need auto approve controllers in two scenarios
+1. certificate renewal
+2. when cluster-admin on hub does not want to approve manually.
+
+The auto approve controller identify whether a csr should be approve based on the following rules:
+1. Whether the Orgnization field in subject is `open-cluster-management:cluster`
+2. Whether the CommonName field has a prefix of `open-cluster-management:cluster`
+3. Whether the it is a renewal CSR request
+4. If it is not a renewl CSR request, whether the cluster name in CommonName does not exist as a `JoinedCluster` in hub server.
```
deads2k: This isn't a guarantee that the same actor did both.  I don't think this is safe to do because it allows the pre-creation of clients certificates that may later have value.

docs/clusterjoinprocess.md
qiujian16: role/rolebinding is created on hub to allow spoke agent to access the cluster namespace. Then we put work and other object in this cluster namespace for agent on spoke cluster to access.

docs/clusterjoinprocess.md
qiujian16: i am fine to put it as cluster scoped and that resolves a lot of issue. Original thinking is if it is namespace scoped, we can control which user can access which clusters, by giving user an access to that namespace. 

However, even if it is cluster scoped, we still need a cluster namespace (or that should be defined in the spec). So the hub controller should be able to control the access of the agent on the spoke cluster.

qiujian16: ok, i think the hub-controller should create a clusterrole/clusterrolebinding for the agent on spoke cluster at first, then the agent can create the `SpokeCluster` object. I will update the process here.

clusters/v1/types.go
pmorie: +1 on making type a string.

docs/clusterjoinprocess.md
pmorie: Suggestions: 

- add a level of detail on resource name/namespace coordinates

If the `JoinedCluster` is cluster-scoped, I would expect the spoke to need only the name of the cluster's resource in the hub.

> to set cluster name/namespace

I would write this as "name of the JoinedCluster resource to create in the Hub" (that's what it refers to, correct?)

docs/clusterjoinprocess.md
pmorie: I'm reading that the CN should be, given the comments above, `system:open-cluster-management:spoke-cluster-name:spoke-agent-name`. Accurate?

docs/clusterjoinprocess.md
pmorie: I also expect the spoke to create `JoinedCluster`.

Do we want the clusterrole bound to the agent's SA on the hub to be enabled to create any JoinedCluster, or should we scope the permission to only a single name or set of names?

pmorie: If there is a namespaced resource that mirrors to the cluster-scoped one, the namespaced resource can be the lower-permissioned one users are given access to view/use

qiujian16: it should only be allowed to create and update status of a single name.

qiujian16: I think the mirrored resources should be read-only. All update to the spokeCluster resources should be done on the cluster scoped API and mirror to the namespace scoped API.

clusters/v1/types.go
```diff
@@ -0,0 +1,128 @@
...
+
+// ResourceList defines a map for the quantity of different resources, the definition
+// matches the ResourceList defined in k8s.io/api/core/v1
+type ResourceList map[ResourceName]resource.Quantity
+
+// SpokeClusterStatusCondition contains condition information for a spoke cluster.
+type SpokeClusterStatusCondition struct {
+	// Type is the type of the cluster condition.
+	// +required
+	Type string `json:"type"`
```
skeeey: using `ClusterConditionType` instead of `string`

docs/clusterjoinprocess.md
qiujian16: I think what you mean is there are multiple agents (search agent/topology agent etc.) on one spoke cluster, they all submit different csr with their own identity to the hub. The organization will be the same as `"Orgnization": ["open-cluster-management:clusterName"]`?

docs/clusterjoinprocess.md
qiujian16: check step is added, but i am not sure the meaning of "infinite renewal"  problem. Would you give more explanation?

docs/clusterjoinprocess.md
qiujian16: yes, changes to cluster scoped.  And we need controller to create a namespace with the same name as spoke cluster.

clusters/v1/types.go

clusters/v1/types.go

clusters/v1/types.go
qiujian16: cluster-admin should approve the csr to accept the cluster. And the spokecluster can be created by agent on hub or cluster-admin on hub, so it may not be safe because spoke can choose to accept itself.

qiujian16:
@deads2k @pmorie i update the API and docs with the following changes.

1. resolve comments from the last review.
2. Indicate that a namespace with the same name of `SpokeCluster` will be created by hub controller.
3. Add a `ApprovePolicy` field in `SpokeCluster`, so only when cluster-admin set it to `Auto`, the join of spoke cluster will be automatically approved by hub.
4. Add another API `SpokeAgentAccessConfiguration`, such that hub controller knows how to control the access of spoke agent to the namespace on hub.

clusters/v1/doc.go
```diff
@@ -0,0 +1,4 @@
+// Package v1 contains API Schema definitions for the operators v1 API group
+// +k8s:deepcopy-gen=package,register
+// +groupName=clusters.open-cluster-management.io
```
pmorie: Let's make sure we all feel good about the apigroup name 'clusters' here. What about 'spokes' ?

With `clusters` you wind up with `clusters.Cluster`. Give it some thought.

clusters/v1/types.go
```diff
@@ -0,0 +1,203 @@
+package v1
```
pmorie, line 1: One thing that isn't present in this PR is the YAML for the CRDs that add this API to the control plane. I imagine that we would contain the go api, CRDs, webhooks, e2es, etc for the API to this repo.  I'm not suggesting we do all of these things in this initial PR; just wondering if we have the same vision.

qiujian16: yes, we should add them also in this repo.

clusters/v1/types.go
pmorie: Do we need a distinct spec and status field here? If we retain both fields, he spec and status fields should make it very clear from doc how they are different. When would you use the status field vs the spec field?

clusters/v1/types.go
```diff
@@ -0,0 +1,203 @@
...
+	// AgentName  represents the identity of the agent. The user binded to the certain role
+	// is in the format of "system:open-cluster-management:clusterName:agentName"
+	AgentName string `json:"agentName"`
+
+	// Rules holds all the PolicyRules for this agent.
+	Rules []PolicyRule `json:"rules"`
+}
+
+// PolicyRule defines the rule of policy, refer to k8s.io/api/rbac/v1
+type PolicyRule struct {
```
pmorie: IMO this should have a more specific name, like `SpokeAgentPermission`. 'Policy' is a little broad name to use for a type this API exports.

clusters/v1/doc.go
deads2k: > Let's make sure we all feel good about the apigroup name 'clusters' here. What about 'spokes' ?
> 
> With `clusters` you wind up with `clusters.Cluster`. Give it some thought.

spokeclusters.cluster.open-cluster-management.io.  You could add anything related to cluster metadata here.

deads2k: cluster is singular

clusters/v1/types.go
```diff
@@ -0,0 +1,203 @@
...
+
+import (
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// SpokeCluster represents the current status of spoke cluster.
+// SpokeCluster is cluster scoped resources whose name is globally unique.
```
deads2k: All cluster scoped resources have globally unique names.  Does this resource have a name that matches the cluster name?  Cluster UID?  A tuple of some kind?

clusters/v1/types.go
```diff
@@ -0,0 +1,203 @@
...
+	AutoApprovePolicyType ApprovePolicyType = "Auto"
+	// ManualApprovePolicyType represents hub needs human intervention to allow the
+	// join of spoke cluster.
+	ManualApprovePolicyType ApprovePolicyType = "Manual"
+)
+
+// SpokeClusterSpec represents a desired configuration for the agents on the spoke cluster.
+type SpokeClusterSpec struct {
+	// MasterAddress represents the apiserver address of the spoke cluster
+	MasterAddress KubernetesAPIServerEndpoint `json:"masterAddress,omitempty"`
```
deads2k: `SpokeClientConfig`

clusters/v1/types.go
```diff
@@ -0,0 +1,203 @@
...
+
+	// ApprovePolicy is the policy for hub to approve the join of spoke cluster.
+	// Allowd values are Auto and Manual. If set as Auto, hub will automatically
+	// approve the join of spoke cluster, otherwise, cluster-admin need to approve
+	// CSR manually.
+	ApprovePolicy ApprovePolicyType `json:"approvePolicy,omitempty"`
+}
+
+// KubernetesAPIServerEndpoint represents apiserver address of the spoke cluster
+type KubernetesAPIServerEndpoint struct {
```
deads2k: `ClientConfig`

clusters/v1/types.go
```diff
@@ -0,0 +1,203 @@
...
+	MasterAddress KubernetesAPIServerEndpoint `json:"masterAddress,omitempty"`
+
+	// ApprovePolicy is the policy for hub to approve the join of spoke cluster.
+	// Allowd values are Auto and Manual. If set as Auto, hub will automatically
+	// approve the join of spoke cluster, otherwise, cluster-admin need to approve
+	// CSR manually.
+	ApprovePolicy ApprovePolicyType `json:"approvePolicy,omitempty"`
+}
+
+// KubernetesAPIServerEndpoint represents apiserver address of the spoke cluster
```
deads2k: provides the information to securely connect to a remote server and verify its identity

clusters/v1/types.go
deads2k: This had made one shortcoming clear: how does the hub know what identity to use to contact the spoke?  Doesn't have to be immediately addressed.  One possibility is that the spoke pushes a secret containing creds into the namespace.

clusters/v1/types.go
```diff
@@ -0,0 +1,203 @@
...
+	// Node, in numbers
+	ResourceNodes = "node"
+)
+
+// ResourceList defines a map for the quantity of different resources, the definition
+// matches the ResourceList defined in k8s.io/api/core/v1
+type ResourceList map[ResourceName]resource.Quantity
+
+// SpokeClusterStatusCondition contains condition information for a spoke cluster.
+type SpokeClusterStatusCondition struct {
```
deads2k: Just `StatusCondition`

clusters/v1/types.go
deads2k: > using `ClusterConditionType` instead of `string`

No.  We want this to be generic.  I will get around to writing the "all conditions are the same" KEP upstream this week.

clusters/v1/types.go
```diff
@@ -0,0 +1,203 @@
...
+	// Status is the status of the condition. One of True, False, Unknown.
+	// +required
+	Status metav1.ConditionStatus `json:"status"`
+
+	// LastTransitionTime is the last time the condition changed from one status to another.
+	// +optional
+	LastTransitionTime metav1.Time `json:"lastTransitionTime,omitempty"`
+
+	// Reason is a (brief) reason for the condition's last status change.
+	// +optional
```
deads2k: required

clusters/v1/types.go
```diff
@@ -0,0 +1,203 @@
...
+	// Type is the type of the cluster condition.
+	// +required
+	Type string `json:"type"`
+
+	// Status is the status of the condition. One of True, False, Unknown.
+	// +required
+	Status metav1.ConditionStatus `json:"status"`
+
+	// LastTransitionTime is the last time the condition changed from one status to another.
+	// +optional
```
deads2k: required

clusters/v1/types.go
```diff
@@ -0,0 +1,203 @@
...
+
+	// Capacity represents the total resource capacity from all nodeStatuses
+	// on the spoke cluster
+	Capacity ResourceList `json:"capacity,omitempty"`
+
+	// Allocatable represents the total allocatable resources on the spoke cluster
+	Allocatable ResourceList `json:"allocatable,omitempty"`
+
+	// MasterAddress represents the apiserver URL on the spoke cluster
+	MasterAddress string `json:"masterAddress,omitempty"`
```
deads2k: This cruft should be removed.  I think you only need it on spec.

clusters/v1/types.go
```diff
@@ -0,0 +1,203 @@
...
+
+	// Version represents the kubernetes version of the spoke cluster
+	Version string `json:"version,omitempty"`
```
deads2k: I suspect you'll grow to include other versions, things like the klusterlet version.

Please make a `Versions struct` that contains `Kubernetes string` for starters and a TODO comment about adding the spoke agent (klusterlet?) version.

clusters/v1/types.go
```diff
@@ -0,0 +1,203 @@
...
+// SpokeClusterSpec represents a desired configuration for the agents on the spoke cluster.
+type SpokeClusterSpec struct {
+	// MasterAddress represents the apiserver address of the spoke cluster
+	MasterAddress KubernetesAPIServerEndpoint `json:"masterAddress,omitempty"`
+
+	// ApprovePolicy is the policy for hub to approve the join of spoke cluster.
+	// Allowd values are Auto and Manual. If set as Auto, hub will automatically
+	// approve the join of spoke cluster, otherwise, cluster-admin need to approve
+	// CSR manually.
+	ApprovePolicy ApprovePolicyType `json:"approvePolicy,omitempty"`
```
deads2k: I don't think we should allow an auto policy to start.  let's remove this.

clusters/v1/types.go
```diff
@@ -0,0 +1,203 @@
...
+	// MasterAddress represents the apiserver URL on the spoke cluster
+	MasterAddress string `json:"masterAddress,omitempty"`
+
+	// Version represents the kubernetes version of the spoke cluster
+	Version string `json:"version,omitempty"`
+}
+
+const (
+	// ClusterOK means that the cluster is "OK".
+	ClusterOK string = "ClusterOK"
```
deads2k: `SpokeClusterConditionOk = ClusterOk` and similar for all of these

clusters/v1/types.go
```diff
@@ -0,0 +1,203 @@
...
+	// ClusterJoined means the spoke cluster has successfully joined the hub
+	ClusterJoined string = "ClusterJoined"
+	// ClusterJoinApproved means the request to join the cluster is approved by user or controller
+	ClusterJoinApproved string = "HubApprovedJoin"
+	// ClusterJoinDenied means the request to join the cluster is denied by user or controller
+	ClusterJoinDenied string = "JoinDenied"
```
deads2k: is this "HubJoinDenied"?

clusters/v1/types.go
```diff
@@ -0,0 +1,203 @@
...
+
+	// Version represents the kubernetes version of the spoke cluster
+	Version string `json:"version,omitempty"`
+}
+
+const (
+	// ClusterOK means that the cluster is "OK".
+	ClusterOK string = "ClusterOK"
+	// ClusterJoined means the spoke cluster has successfully joined the hub
+	ClusterJoined string = "ClusterJoined"
```
deads2k: "SpokeClusterJoined" ?

clusters/v1/types.go
```diff
@@ -0,0 +1,203 @@
...
+
+const (
+	// CPU, in cores. (500m = .5 cores)
+	ResourceCPU ResourceName = "cpu"
+	// Memory, in bytes. (500Gi = 500GiB = 500 * 1024 * 1024 * 1024)
+	ResourceMemory ResourceName = "memory"
+	// Volume size, in bytes (e,g. 5Gi = 5GiB = 5 * 1024 * 1024 * 1024)
+	ResourceStorage ResourceName = "storage"
+	// Node, in numbers
+	ResourceNodes = "node"
```
deads2k: This is novel. Please remove for now and we'll re-assess later

clusters/v1/types.go
```diff
@@ -0,0 +1,203 @@
...
+// ResourceName is the name identifying various resources in a ResourceList.
+type ResourceName string
+
+const (
+	// CPU, in cores. (500m = .5 cores)
+	ResourceCPU ResourceName = "cpu"
+	// Memory, in bytes. (500Gi = 500GiB = 500 * 1024 * 1024 * 1024)
+	ResourceMemory ResourceName = "memory"
+	// Volume size, in bytes (e,g. 5Gi = 5GiB = 5 * 1024 * 1024 * 1024)
+	ResourceStorage ResourceName = "storage"
```
deads2k: do nodes report this today?  I just don't know.

docs/clusterjoinprocess.md
```diff
@@ -0,0 +1,91 @@
...
+3. hub controller
+4. agent on spoke cluster
+
+Some rules on cluster join:
+- The name of the clustr must be golbally unique on hub and conforms to dns label format.
+
+## Cluster join initiated from spoke cluster
+1. cluster-admin on spoke cluster create bootstrap kubeconfig (connects to hub cluster) as secret and deploy the agent
+on spoke cluster.
+  - Cluster-admin on spoke needs to set cluster name when deploying the agent on spoke cluster.
```
deads2k: This looks potentially conflicty.  Especially if you join to multiple hubs.  In openshift at least, clusters have a UID.  We should recommend that as a procedure and probably make a spoke resource in `cluster.open-cluster-managemenent.io` to represent it in spokes.

docs/clusterjoinprocess.md
```diff
@@ -0,0 +1,91 @@
...
+2. agent on spoke cluster creates CSR on hub cluster using bootstrap kubeconfig.
+  - The subject in csr is
+`{"Organization": ["system:open-cluster-management:clusterName"], "CommonName":"system:open-cluster
+management:clusterName:agentName"}`.
+  - The name of the csr is the digest of subject and private key, with a common prefix.
+  CSR will specify the signer name as the kube-client one.
+3. cluster-admin on hub-cluster approve the CSR.
+4. hub-controller creates a clusterrolebinding on the hub with the identity of
+`system:open-cluster-management:clusterName:agentName`
+   - Allows the create and status update of `SpokeCluster`
```
deads2k: Creating a SpokeCluster should not be considered highly privileged.  You cannot control what names can be created via RBAC and you have no way to establish proof of ownership of the cluster

docs/clusterjoinprocess.md
```diff
@@ -0,0 +1,91 @@
...
+5. hub-controller creates a namespace as the name of cluster on hub cluster if it does not exist.
+e.g. if the the name of cluster is `cluster1`, the hub controller creates a namespace of `cluster1`
+with a label of `open-cluster-management.io/cluster=cluster1`.
```
deads2k: What is the future purpose of this label?

docs/clusterjoinprocess.md
```diff
@@ -0,0 +1,91 @@
...
+`{"Organization": ["system:open-cluster-management:clusterName"], "CommonName":"system:open-cluster
+management:clusterName:agentName"}`.
+  - The name of the csr is the digest of subject and private key, with a common prefix.
+  CSR will specify the signer name as the kube-client one.
+3. cluster-admin on hub-cluster approve the CSR.
+4. hub-controller creates a clusterrolebinding on the hub with the identity of
+`system:open-cluster-management:clusterName:agentName`
+   - Allows the create and status update of `SpokeCluster`
+5. hub-controller creates a namespace as the name of cluster on hub cluster if it does not exist.
+e.g. if the the name of cluster is `cluster1`, the hub controller creates a namespace of `cluster1`
```
deads2k: The namespace logically needs to include the agentName, right?

docs/clusterjoinprocess.md
```diff
@@ -0,0 +1,91 @@
...
+on spoke cluster.
+  - Cluster-admin on spoke needs to set cluster name when deploying the agent on spoke cluster.
+  - The credential in bootstrap kubeconfig needs to have the permission to create/watch csr.
+2. agent on spoke cluster creates CSR on hub cluster using bootstrap kubeconfig.
+  - The subject in csr is
+`{"Organization": ["system:open-cluster-management:clusterName"], "CommonName":"system:open-cluster
+management:clusterName:agentName"}`.
+  - The name of the csr is the digest of subject and private key, with a common prefix.
+  CSR will specify the signer name as the kube-client one.
+3. cluster-admin on hub-cluster approve the CSR.
```
pmorie: I would expect the status to update when the join was approved; currently it's afterward (step 7)

docs/clusterjoinprocess.md
deads2k: > The namespace logically needs to include the agentName, right?

I can also live with an outcome that says "spokes can only join each hub once, but may join multiple hubs".  But be specific and clear about the impacts.

docs/clusterjoinprocess.md
```diff
@@ -0,0 +1,91 @@
...
+4. hub-controller creates a clusterrolebinding on the hub with the identity of
+`system:open-cluster-management:clusterName:agentName`
+   - Allows the create and status update of `SpokeCluster`
+5. hub-controller creates a namespace as the name of cluster on hub cluster if it does not exist.
+e.g. if the the name of cluster is `cluster1`, the hub controller creates a namespace of `cluster1`
+with a label of `open-cluster-management.io/cluster=cluster1`.
+6. hub-controller creates role/rolebinding on the cluster namespace on the hub
+  - Allow the access of agent on spoke cluster to the namespace.
+  - The role is retrived from `SpokeAgentAccessConfig` based on agent name.
```
deads2k: I don't think you need flexibility for this role.

docs/clusterjoinprocess.md
```diff
@@ -0,0 +1,91 @@
...
+1. cluster-admin on spoke cluster
+2. cluster-admin on hub cluster
+3. hub controller
+4. agent on spoke cluster
+
+Some rules on cluster join:
+- The name of the clustr must be golbally unique on hub and conforms to dns label format.
+
+## Cluster join initiated from spoke cluster
+1. cluster-admin on spoke cluster create bootstrap kubeconfig (connects to hub cluster) as secret and deploy the agent
```
pmorie: Let's be more specific on the APIs that drive this in the hub. In our review, we mentioned driving this part of the join flow via a 'bootstrap' resource that controlled the setup flow in the hub (creation of the kubeconfig, etc).

docs/clusterjoinprocess.md
```diff
@@ -0,0 +1,91 @@
...
+  - Cluster-admin on spoke needs to set cluster name when deploying the agent on spoke cluster.
+  - The credential in bootstrap kubeconfig needs to have the permission to create/watch csr.
+2. agent on spoke cluster creates CSR on hub cluster using bootstrap kubeconfig.
+  - The subject in csr is
+`{"Organization": ["system:open-cluster-management:clusterName"], "CommonName":"system:open-cluster
+management:clusterName:agentName"}`.
+  - The name of the csr is the digest of subject and private key, with a common prefix.
+  CSR will specify the signer name as the kube-client one.
+3. cluster-admin on hub-cluster approve the CSR.
+4. hub-controller creates a clusterrolebinding on the hub with the identity of
```
pmorie: I personally was imagining that 4-6 would be driven by a bootstrap resource.

clusters/v1/types.go
qiujian16: This is previously used to report PV usage on the spoke cluster, but it is not quite accurate. I will remove this.

clusters/v1/doc.go
qiujian16: We do rename it as `SpokeCluster`, so it will be `cluster.SpokeCluster`. Does it look better?

docs/clusterjoinprocess.md
qiujian16: We have not considered the spoke agent to join multiple hubs. Is it for the purpose of multi-tenancy? 

Agree that it may have name conflict. By default we can read openshift cluster UID, or generate a UID as a cluster name, and allow overriding the cluster name when starting agent on the spoke cluster. Does it make sense?

docs/clusterjoinprocess.md
qiujian16: We can let hub to create the empty `SpokeCluster`, since the hub can get the clusterName from csr, only allowing update and status update from agent on spoke. Does it sound right?

docs/clusterjoinprocess.md
qiujian16: The purpose is to filtering what namespace is used as cluster namespace in the future.

docs/clusterjoinprocess.md
qiujian16: To my understanding, the agent here represents the different agent connects to the hub. Like search agent/work agent/topology agent etc., right?

We can let those agent to watch the same cluster namespace, but give them different roles to access the hub.

docs/clusterjoinprocess.md
qiujian16: OK. since it will involve another bootstrap API, i will describe the procedure and links to the current implementation. We will need another PR to refactor that bootstrap API in this repo later.

docs/clusterjoinprocess.md
```diff
@@ -0,0 +1,91 @@
...
+4. agent on spoke cluster
+
+Some rules on cluster join:
+- The name of the clustr must be golbally unique on hub and conforms to dns label format.
+
+## Cluster join initiated from spoke cluster
+1. cluster-admin on spoke cluster create bootstrap kubeconfig (connects to hub cluster) as secret and deploy the agent
+on spoke cluster.
+  - Cluster-admin on spoke needs to set cluster name when deploying the agent on spoke cluster.
+  - The credential in bootstrap kubeconfig needs to have the permission to create/watch csr.
```
deads2k: I think the bootstrap cred should also be able to create the SpokeCluster resource.

clusters/v1/types.go
```diff
@@ -0,0 +1,203 @@
...
+import (
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// SpokeCluster represents the current status of spoke cluster.
+// SpokeCluster is cluster scoped resources whose name is globally unique.
+type SpokeCluster struct {
```
deads2k, line 17: Describe at a high level the double opt-in.  This is effectively:

 1. hub cluster-admin approves a CSR for the spoke cluster's UID and agent.
 2. spoke cluster-admin creates the credentials to provide to the hub cluster and once the hub cluster creates the namespace, the final act is for the spoke agent to push the hub's credentials to use against the spoke's kube-apiserver

docs/clusterjoinprocess.md
deads2k: > We have not considered the spoke agent to join multiple hubs. Is it for the purpose of multi-tenancy?

I think there will be multiple hubs.  One for an ops team across all clusters and another for an app team across some clusters as a for instance.

deads2k: > Agree that it may have name conflict. By default we can read openshift cluster UID, or generate a UID as a cluster name, and allow overriding the cluster name when starting agent on the spoke cluster. Does it make sense?

I think a resource for the purpose makes sense.  On openshift, we'll fill it in via the existing cluster UID.  On other platforms, creation can auto-generate it via admission.

docs/clusterjoinprocess.md
deads2k: > The purpose is to filtering what namespace is used as cluster namespace in the future.

If the namespace matches the cluster UID, then you already have a perfect mapping for a GET

deads2k:
I think we've made it a long way.  I'd like to see a new API PR created from this that starts by simply creating a `SpokeCluster` resource with Spec and Status and with the StatusCondition struct.  The spec would be empty, the Status only contains conditions.  We'll merge that minimal API to get the CI working and tested on this repo.

docs/clusterjoinprocess.md
pmorie: Agree

docs/clusterjoinprocess.md
qiujian16: you are right, remove it

docs/clusterjoinprocess.md
qiujian16: bootstrap resources on hub will deploy the agent on the spoke. The 4-6 should happen only when a the spoke cluster is joined and approved by the hub.

docs/clusterjoinprocess.md
qiujian16: rm this

docs/clusterjoinprocess.md
qiujian16: done

docs/clusterjoinprocess.md
qiujian16: right, changed the sequence here.

docs/clusterjoinprocess.md
qiujian16: done

docs/clusterjoinprocess.md
qiujian16: added a process "Cluster join initiated from bootstrap resources on hub" separately.

clusters/v1/types.go
qiujian16: it is still not clear about the usage of this API. Removed it for now.

clusters/v1/types.go
qiujian16: We do not have a use case on hub talks to spoke apiserver using kubeconfig yet. Why do we need this creds of spoke cluster pushes to hub? One possibility is to fetch log from spoke apiserver?

deads2k: > We do not have a use case on hub talks to spoke apiserver using kubeconfig yet. Why do we need this creds of spoke cluster pushes to hub? One possibility is to fetch log from spoke apiserver?

It seems a little weird to have the connection information then, but I'll go with it for now

docs/clusterjoinprocess.md
deads2k: > We can let hub to create the empty `SpokeCluster`, since the hub can get the clusterName from csr, only allowing update and status update from agent on spoke. Does it sound right?

I would have it be spoke initiated and/or external actor initiated (hub admin).  I would not expect automated creation by a hub controller.

clusters/v1/types.go
```diff
@@ -0,0 +1,247 @@
...
+
+	// Version represents the kubernetes version of the spoke cluster
+	Version SpokeVersion `json:"version,omitempty"`
+}
+
+// SpokeVersion represents the Kubernetes version of spoke cluster
+// TODO add spoke agent versions
+type SpokeVersion struct {
+	// Kubernetes is the kubernetes version of spoke cluster
+	Kubernetes string `json:"version,omitempty"`
```
deads2k: json tag should be kubernetes
this should be optional

clusters/v1/types.go
```diff
@@ -0,0 +1,247 @@
...
+
+	// Status represents the current status of joined spoke cluster
+	Status SpokeClusterStatus `json:"status,omitempty"`
+}
+
+// SpokeClusterSpec provides the information to securely connect to a remote server
+// and verify its identity
+type SpokeClusterSpec struct {
+	// SpokeClientConfig represents the apiserver address of the spoke cluster
+	SpokeClientConfig ClientConfig `json:"masterAddress,omitempty"`
```
deads2k: json tag should be spokeClientConfig
should be optional - EDIT "NOT" optional

clusters/v1/types.go
```diff
@@ -0,0 +1,247 @@
...
+}
+
+// ClientConfig represents apiserver address of the spoke cluster
+// TODO include credential to connect to spoke cluster kube-apiserver
+type ClientConfig struct {
+	// URL is the url of apiserver endpoint of the spoke cluster
+	URL string `json:"url"`
```
deads2k, line 45: this one is required

clusters/v1/types.go
```diff
@@ -0,0 +1,247 @@
...
+	// CABundle is the ca bundle to connect to apiserver of the spoke cluster, system
+	// ca is used if it is not set.
+	CABundle []byte `json:"caBundle,omitempty"`
```
deads2k, line 49: this should be optional and explain that if not set, the system certs are used.

clusters/v1/types.go
```diff
@@ -0,0 +1,247 @@
...
+
+// SpokeVersion represents the Kubernetes version of spoke cluster
+// TODO add spoke agent versions
+type SpokeVersion struct {
+	// Kubernetes is the kubernetes version of spoke cluster
+	Kubernetes string `json:"version,omitempty"`
+}
+
+const (
+	// SpokeClusterConditionOK means that the cluster is "OK".
```
deads2k: let's snip this and you can re-add later with a better explanation of what it means.

clusters/v1/types.go
```diff
@@ -0,0 +1,247 @@
...
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// SpokeBootstrapConfiguration defines the desired configuration to deploy agent on spoke
+// SpokeBootstrapConfiguration is namespace scoped resource and created on hub.
+// The status.ClusterBootStrapInfo in SpokeBootstrapConfiguration refers to a secret that
+// stores the deployment files including:
+// - secret with bootstrap kubeconfig
+// - crd and deployment spec of spoke agent operator
+// - cr to deploy spoke agent
+// The cluster-admin of apoke agent should fetch the secret and apply it on spoke cluster.
```
deads2k: typo spoke agent

clusters/v1/types.go
```diff
@@ -0,0 +1,247 @@
...
+type SpokeBootstrapConfiguration struct {
```
deads2k: Let's pull this API out of this PR so we can land the SpokeCluster

docs/clusterjoinprocess.md
```diff
@@ -0,0 +1,73 @@
...
+This is to describe the various process of cluster join
+
+Actors:
+1. cluster-admin on spoke cluster
+2. cluster-admin on hub cluster
+3. hub controller
+4. agent on spoke cluster
+
+Some rules on cluster join:
+- The name of the clustr must be golbally unique on hub and conforms to dns label format.
```
deads2k: typos: cluster and globally

clusters/v1/types.go
mdelder: Here this is changing the nomenclature to "Spoke" whereas all other references are using "Managed Endpoint" or "Managed Cluster". I would like to keep these consistent.

clusters/v1/types.go
mdelder: I'm expecting there's our klusterlet management agent and optionally additional agents that perform other vertical functions (like integration with monitoring, metering, etc). I would not expect that we have customized agents for the klusterlet use cases though.

clusters/v1/types.go
mdelder: We had previously discussed using a messaging layer to capture heartbeats. @deads2k is your feedback about heart beats in general or just collecting that information into `etcd`?

qiujian16: we can also consider using `Lease` API just as kubelet does.

deads2k:
I like this.  we'll hone it from here

/lgtm

PR #3, opened 2020-03-23
Add hubacceptsclient field

qiujian16:


clusters/v1/types.go
```diff
@@ -35,6 +35,12 @@ type SpokeClusterSpec struct {
 	// SpokeClientConfig represents the apiserver address of the spoke cluster
 	// +optional
 	SpokeClientConfig ClientConfig `json:"spokeClientConfig,omitempty"`
+
+	// HubAcceptsClient reprsents that hub accepts the join of spoke agent.
+	// Its default value is false, and can only be upated when the user on hub
+	// has an RBAC rule to UPDATE on the virtual subresource of spokecluster/accept.
+	// +required
+	HubAcceptsClient bool `json:"HubAcceptsClient"`
```
deads2k: `acceptSpokeCluster`

clusters/v1/types.go
```diff
@@ -35,6 +35,12 @@ type SpokeClusterSpec struct {
 	// SpokeClientConfig represents the apiserver address of the spoke cluster
 	// +optional
 	SpokeClientConfig ClientConfig `json:"spokeClientConfig,omitempty"`
+
+	// HubAcceptsClient reprsents that hub accepts the join of spoke agent.
+	// Its default value is false, and can only be upated when the user on hub
+	// has an RBAC rule to UPDATE on the virtual subresource of spokecluster/accept.
```
deads2k: spokeclusters/accept (the URLs are plural)

clusters/v1/types.go
```diff
@@ -35,6 +35,12 @@ type SpokeClusterSpec struct {
 	// SpokeClientConfig represents the apiserver address of the spoke cluster
 	// +optional
 	SpokeClientConfig ClientConfig `json:"spokeClientConfig,omitempty"`
+
+	// HubAcceptsClient reprsents that hub accepts the join of spoke agent.
+	// Its default value is false, and can only be upated when the user on hub
```
deads2k: "can only be set to true"

It could be done on create if the creating user has sufficient power.

clusters/v1/types.go
```diff
@@ -35,6 +35,12 @@ type SpokeClusterSpec struct {
 	// SpokeClientConfig represents the apiserver address of the spoke cluster
 	// +optional
 	SpokeClientConfig ClientConfig `json:"spokeClientConfig,omitempty"`
+
+	// HubAcceptsClient reprsents that hub accepts the join of spoke agent.
```
deads2k: tell me what happens when this is true and when it is false.  I think it's like this

When set to true, a namespace is created on the hub that represents the spokecluster and grants permission to the spoke cluster.
When set to false, the namespace that represents the hub is deleted.

docs/clusterjoinprocess.md
```diff
@@ -29,13 +29,16 @@ management:clusterName:agentName"}`.
 5. hub-controller creates a clusterrolebinding on the hub with the identity of
 `system:open-cluster-management:clusterName:agentName`
    - Allows status update of `SpokeCluster`
-6. hub-controller updates condition of `SpokeCluster` to `HubApprovedJoin`.
-7. hub-controller creates a namespace as the name of cluster on hub cluster if it does not exist.
+6. cluster-admin on hub update `spec.hubAcceptsClient` to `true`.
+  - Only user on hub who has the rbac permision to update subresource of `spokecluster/accept`
```
deads2k: spokeclusters/accept

clusters/v1/types.go
deads2k: update godoc on the SpokeCluster type to reflect this.

clusters/v1/types.go
qiujian16: comments added.

clusters/v1/types.go
qiujian16: that is correct, comments updated.

clusters/v1/types.go
qiujian16: more comments are added on SpokeCluster type.

clusters/v1/types.go
```diff
@@ -10,9 +10,11 @@ import (
...
 // SpokeCluster is cluster scoped resources. The name is the cluster UID.
 // The cluster join follows the double opt-in proceess:
-// 1. agent on spoke cluster creates CSR on hub with cluster UID and agent name
-// 2. cluster admin on hub approves the CSR for the spoke's cluster UID and agent name
-// 3. cluster admin on spoke creates credential. Once hub creates the cluster namespace,
+// 1. agent on spoke cluster creates CSR on hub with cluster UID and agent name.
+// 2. agent on spoke cluster creates spokecluster on hub.
+// 3. cluster admin on hub approves the CSR for the spoke's cluster UID and agent name.
+// 4. cluster admin set spec.acceptSpokeCluster of spokecluster to true.
+// 5. cluster admin on spoke creates credential. Once hub creates the cluster namespace,
```
deads2k: what is this credential used for?  Is it related to any other credential?  What does the credential have the power to do.

pmorie: Do we want to give treatment to behavior of admission controllers that reject unauthorized writes to `spec.acceptSpokeCluster` / what RBAC permission is checked / etc?

pmorie: Nevermind, I see that is accounted for. Well done

clusters/v1/types.go
```diff
@@ -35,6 +37,17 @@ type SpokeClusterSpec struct {
 	// SpokeClientConfig represents the apiserver address of the spoke cluster
 	// +optional
 	SpokeClientConfig ClientConfig `json:"spokeClientConfig,omitempty"`
+
+	// AcceptSpokeCluster reprsents that hub accepts the join of spoke agent.
+	// Its default value is false, and can only be set true when the user on hub
+	// has an RBAC rule to UPDATE on the virtual subresource of spokeclusters/accept.
+	// When the vaule is set true, a namespace is created on hub representing the
+	// spoke cluster, also role/rolebinding is created on the namespace to grant the
```
pmorie: This is a good place to talk about the name of the namespace and its relationship to `SpokeCluster`. What happens if there is a collision with an existing namespace?

deads2k:
just the question, lgtm otherwise.

clusters/v1/types.go
```diff
@@ -35,6 +37,17 @@ type SpokeClusterSpec struct {
...
+	// permision of access from agent on spoke.
+	// When the value is set false, the namespace representing the spoke cluster is
+	// deleted.
+	// +required
+	AcceptSpokeCluster bool `json:"acceptSpokeCluster"`
```
pmorie: One thing I noticed is that here we use `accept`, but else where we use `approve`, eg:

```go
SpokeClusterClusterConditionJoinApproved string = "HubApprovedJoin"
```

Seems like this field should be called 'ApproveSpokeCluster` to be consistent.

clusters/v1/types.go
deads2k: > This is a good place to talk about the name of the namespace and its relationship to `SpokeCluster`. What happens if there is a collision with an existing namespace?

You mean if a UID collides?  I would expect the namespace to be made to be "right"

clusters/v1/types.go
deads2k: I would have gone the other way: "HubAcceptedSpoke".  I don't feel strongly enough to argue over it though.

clusters/v1/types.go
qiujian16: We currently do not have a use case for how hub uses the kubeconfig credential of spoke. I think it can be used to get the pod log.

clusters/v1/types.go
qiujian16: i will go with accept...seems clearer

pmorie: SGTM

PR #5, opened 2020-03-25
Build infra

pmorie:
Adds basic build infrastructure, required to enable prow


pmorie:
Auto-merging since we need this stuff in place for prow

PR #8, opened 2020-03-26
Add basic readme

pmorie:
...and test CI :)


deads2k:
interesting deps for future us.

let's see about prow
/lgtm

PR #11, opened 2020-04-03
using plural cluster in api group

skeeey:
Our generated client cannot list the `spokecluster` due to

```
Failed to list *v1.SpokeCluster: the server could not find the requested resource (get spokeclusters.cluster.open-cluster-management.io)
```

this because we register the api with a singular cluster in api group (spokeclusters.**cluster**.open-cluster-management.io), I think we should use the plural form

### DIFF
--- a/clusters/install.go
+++ b/clusters/install.go
@@ -8,7 +8,7 @@ import (
 )
 
 const (
-	GroupName = "cluster.open-cluster-management.io"
+	GroupName = "clusters.open-cluster-management.io"
 )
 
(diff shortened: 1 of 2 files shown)


pmorie:
/hold

I have confused myself a bit, thinking about this now.

pmorie:
I see why this is confusing now. The name of the package/directory should be `cluster` instead of `clusters`. For reference, notice that we use singular package names (with the exception of `apps) in openshift/api.

PR #12, opened 2020-04-03
Add work API

qiujian16:


clusters/v1/types.go
```diff
@@ -148,3 +149,140 @@ type SpokeClusterList struct {
...
+	// +optional
+	RemoteGetter WorkRemoteGetter `json:"remoteGetter,omitempty" protobuf:"bytes,4,opt,name=remoteGetter"`
+}
+
+// WorkType represents the ty
+type WorkType string
+
+const (
+	// ObjectsWorkType means that the work is to deploy objects on the spoke cluster.
+	ObjectsWorkType WorkType = "objects"
```
TheRealHaoLiu: the behavior of this WorkType is not immediately apparent to me

TheRealHaoLiu: i really like how the syncset API allow us to define multiple resources in 1 syncset 
https://github.com/openshift/hive/blob/master/pkg/apis/hive/v1/syncset_types.go

clusters/v1/register.go
```diff
@@ -32,6 +32,8 @@ func addKnownTypes(scheme *runtime.Scheme) error {
 	scheme.AddKnownTypes(GroupVersion,
 		&SpokeCluster{},
 		&SpokeClusterList{},
+		&Work{},
```
deads2k: I see this as existing in different API group.  `apps.open-cluster-management.io` I think

clusters/v1/types.go
```diff
@@ -148,3 +149,140 @@ type SpokeClusterList struct {
...
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// Work represents a workload that hub wants to deploy on the spoke cluster. A
+// workload is defined as a set of kubernetes resources. Work can also represent
+// a patch that hub wants to apply to the existing kubernetes resources on the
+// spoke cluster.
+// Work must be created in the cluster namespace on the hub, so that agent on the
+// corresponding spoke cluster can access this work resource and deploy on the spoke
+// cluster.
+type Work struct {
```
deads2k: `SpokeWorkload`

clusters/v1/types.go
deads2k: `Manifests` I think.  Discriminators are CamelCase

clusters/v1/types.go
```diff
@@ -148,3 +149,140 @@ type SpokeClusterList struct {
...
+}
+
+// WorkType represents the ty
+type WorkType string
+
+const (
+	// ObjectsWorkType means that the work is to deploy objects on the spoke cluster.
+	ObjectsWorkType WorkType = "objects"
+	// PatchesWorkType means that the work is to apply patch on the spoke cluster.
+	PatchesWorkType WorkType = "patches"
```
deads2k: I'm -1 on this because it implies something else creating the thing that we're patching

clusters/v1/types.go
```diff
@@ -148,3 +149,140 @@ type SpokeClusterList struct {
...
+	// RemoteGetterWorkType means that spoke agent fetches resources from remote sources
+	// to deploy on spoke cluster.
+	RemoteGetterWorkType WorkType = "remote"
```
deads2k: I would defer adding this.

clusters/v1/types.go
```diff
@@ -148,3 +149,140 @@ type SpokeClusterList struct {
...
+	// +optional
+	Status WorkStatus `json:"status,omitempty" protobuf:"bytes,3,opt,name=status"`
+}
+
+// WorkSpec represents a desired configuration of work to be deployed on the spoke cluster.
+// +union
+type WorkSpec struct {
+	// +unionDiscriminator
+	// Type defines whether objects or patches should applied on the spoke cluster
+	Type WorkType `json:"type,omitempty" protobuf:"bytes,1,opt,name=type"`
```
deads2k: workload types would contain this inside of ` template` struct. Just in case we add another thing later.

clusters/v1/types.go
```diff
@@ -148,3 +149,140 @@ type SpokeClusterList struct {
...
+
+	// Objects represents a list of kuberenetes resources to be deployed on the spoke cluster.
+	// +optional
+	Objects []runtime.RawExtension `json:"Objects,omitempty" protobuf:"bytes,2,rep,name=objects"`
```
deads2k: `manifests` I'd think

clusters/v1/types.go
```diff
@@ -148,3 +149,140 @@ type SpokeClusterList struct {
...
+
+// WorkStatus represents the current status of work
+type WorkStatus struct {
+	// Conditions contains the different condition statuses for this work.
+	Conditions []StatusCondition `json:"conditions" protobuf:"bytes,1,rep,name=conditions"`
+
+	// ObjectsConditions represents the conditions of the resources deployed or patched on
+	// spoke cluster. The agent on spoke cluster syncs the condition from spoke to the hub.
+	// +optional
+	ObjectConditions []WorkObjectCondition `json:"objectConditions,omitempty" protobuf:"bytes,2,rep,name=objectConditions"`
```
deads2k: I think you end up with a union type in here too that is congruent to your spec template.

clusters/v1/types.go
```diff
@@ -148,3 +149,140 @@ type SpokeClusterList struct {
...
+	// +required
+	Group string `json:"group" protobuf:"bytes,1,opt,name=group"`
+
+	// Version is the version of the kubernetes resource
+	// +required
+	Version string `json:"version" protobuf:"bytes,2,opt,name=version"`
+
+	// Kind is the kind of the kubernetes resource
+	// +required
+	Kind string `json:"kind" protobuf:"bytes,3,opt,name=kind"`
```
deads2k: Also provide the resource.  In order to submit this, you must have done a RESTMapping and the result can be observed here.

clusters/v1/types.go
```diff
@@ -148,3 +149,140 @@ type SpokeClusterList struct {
...
+	// +required
+	ResourceMeta WorkResourceMeta `json:"resourceMeta" protobuf:"bytes,2,opt,name=resourceMeta"`
+
+	// Patch defines the string to be patched on the kubernetes resource
+	// +required
+	Patch string `json:"patch" protobuf:"bytes,3,opt,name=patch"`
+}
+
+// WorkResourceMeta represents the gvk, name and namespace of a resoure
+type WorkResourceMeta struct {
```
deads2k: you should have a required integer for ordering.

clusters/v1/types.go
```diff
@@ -148,3 +149,140 @@ type SpokeClusterList struct {
...
+	// Group is the API Group of the kubernetes resource
+	// +required
```
deads2k: these are logically required, however failures can produce empty strings.

clusters/v1/types.go
```diff
@@ -148,3 +149,140 @@ type SpokeClusterList struct {
...
+	ObjectConditions []WorkObjectCondition `json:"objectConditions,omitempty" protobuf:"bytes,2,rep,name=objectConditions"`
+}
+
+// WorkObjectCondition represents the conditions of the resources deployed or patched on
+type WorkObjectCondition struct {
+	// ResourceMeta represents the gvk, name and namespace of a resoure
+	// +required
+	ResourceMeta WorkResourceMeta `json:"resourceMeta" protobuf:"bytes,1,opt,name=resourceMeta"`
+
+	// Conditions represents the conditions of this resource on spoke cluster
```
deads2k: In this doc here, enumerate your expected conditions.

deads2k: > In this doc here, enumerate your expected conditions.

I see
 1. Applied
 2. Available
 3. Progressing
 4. Degraded

clusters/v1/types.go
```diff
@@ -148,3 +149,140 @@ type SpokeClusterList struct {
...
+
+// WorkRemoteGetter represents a getter on spoke agent to get resources from remote sources
+type WorkRemoteGetter struct {
+	// URL represents a url link where agent on spoke gets resouces to deploy on spoke
+	URL string `json:"url,omitempty" protobuf:"bytes,1,opt,name=url"`
+}
+
+// WorkStatus represents the current status of work
+type WorkStatus struct {
+	// Conditions contains the different condition statuses for this work.
```
deads2k: In this doc here, enumerate your expected conditions.

I see
 1. Applied
 2. Available
 3. Progressing
 4. Degraded

clusters/v1/types.go
qiujian16: ok, removed

clusters/v1/register.go
qiujian16: done

clusters/v1/types.go

clusters/v1/types.go

clusters/v1/types.go

clusters/v1/types.go

qiujian16: changes to manifests. we also want to put multiple resource in one API

clusters/v1/types.go
qiujian16: I think we need some way to change the existing resources on the spoke cluster.  Maybe we should have a separate `SpokePatch` API to do that?

clusters/v1/types.go

clusters/v1/types.go
qiujian16: would you explain more about why we need ordering here?

Makefile
```diff
@@ -19,6 +19,7 @@ GO_LD_FLAGS:=
 # $3 - manifests
 # $4 - output
 $(call add-crd-gen,cluster,./cluster/v1,./cluster/v1,./cluster/v1)
+$(call add-crd-gen,apps,./apps/v1,./apps/v1,./apps/v1)
```
pmorie: I have a strong preference to call this group `work`

apps/v1/types.go
```diff
@@ -0,0 +1,194 @@
...
+// +genclient
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+// +kubebuilder:subresource:status
+
+// SpokeWorkload represents a workload that hub wants to deploy on the spoke cluster.
+// A workload is defined as a set of kubernetes resources.
+// SpokeWorkload must be created in the cluster namespace on the hub, so that agent on the
+// corresponding spoke cluster can access this resource and deploy on the spoke
+// cluster.
+type SpokeWorkload struct {
```
pmorie: I don't think we necessarily need to connect this in name with the `spoke` concept. How about just `Workload`?

apps/v1/types.go
```diff
@@ -0,0 +1,194 @@
...
+	// +optional
+	WorkloadCondition SpokeWorkloadCondition `json:"workloadCondition,omitempty" protobuf:"bytes,2,rep,name=workloadCondition"`
+}
+
+// SpokeWorkloadCondition represents the conditions of the workload deployed on spoke cluster
+// +union
+type SpokeWorkloadCondition struct {
+	// +unionDiscriminator
+	// Type defines the type of workload applied on the spoke cluster
+	Type SpokeWorkloadType `json:"type,omitempty" protobuf:"bytes,1,opt,name=type"`
```
pmorie: Does the type need to be present here?

pmorie (commented):
A few initial thoughts

clusters/v1/types.go
deads2k: > would you explain more about why we need ordering here?

for failure cases, you may not have metadata like name and groupversionkind.  In those cases, the ordinal will help.  Similarly, you have duplicates and the ordinal removes doubt

apps/v1/types.go
deads2k: > I don't think we necessarily need to connect this in name with the `spoke` concept. How about just `Workload`?

You will end up needing to distinguish this namespace scoped resource from a cluster-scoped WorkloadTemplate of sorts.  I guess it would be WorkloadTemplate, but the names  have to distinct.  Do you have a naming suggestion for that?

WorkloadTemplate + WorkloadBinding produces Workload?  I could buy that

apps/v1/types.go
deads2k: > Does the type need to be present here?

Yes.  It is the discriminator.  If we fear a future (whether we have plans now to build it or not) where we have a WorkloadType of `OLMInstance` or `HelmChart` or other sugar, we need a discriminator.  If all versions have a discriminator, then even if they cannot deserialize or interpret the entire struct, they can deserialize enough to know that they should leave it alone.

Lacking that, you cannot have API growth in this area.  While I don't want to go down that path anytime soon, I'm not certain enough that we will never need it to drop the discriminator.

apps/v1/types.go
```diff
@@ -0,0 +1,194 @@
...
+// SpokeWorkloadType represents the type of workload deployed on spoke
+type SpokeWorkloadType string
+
+const (
+	// ManifestsWorkType means that the work is to deploy manifests on the spoke cluster.
+	ManifestsWorkType SpokeWorkloadType = "Manifests"
+)
+
+// WorkResourceMeta represents the gvk, gvr, name and namespace of a resoure
+type WorkResourceMeta struct {
```
deads2k: I think you need an ordinal that matches the order of the items in the spec to handle cases of corruption (no group, version, kind) and cases of duplicates, or other unexpected cases.

apps/v1/types.go
```diff
@@ -0,0 +1,194 @@
...
+	// Group is the API Group of the kubernetes resource
+	// +required
```
deads2k: Core types don't have groups.  Think pods

apps/v1/types.go
```diff
@@ -0,0 +1,194 @@
...
+	Group string `json:"group,omitempty" protobuf:"bytes,1,opt,name=group"`
+
+	// Version is the version of the kubernetes resource
+	// +required
```
deads2k: Since this type is supposed to reflect reality (status), I think reality insists that these fields be optional to handle the error cases where there is an inability to parse information.

apps/v1/types.go
```diff
@@ -0,0 +1,194 @@
...
+	Version string `json:"version,omitempty" protobuf:"bytes,2,opt,name=version"`
+
+	// Kind is the kind of the kubernetes resource
+	// +required
+	Kind string `json:"kind,omitempty" protobuf:"bytes,3,opt,name=kind"`
```
deads2k: Throughout: remove the proto tags.

EDIT I take that back.  You have them, leave them.  We just have to review very carefully

deads2k: > Throughout: remove the proto tags.

I take that back.  You have them, leave them.  We just have to review very carefully

apps/v1/types.go
```diff
@@ -0,0 +1,194 @@
...
+// SpokeWorkloadTemplate represents the resource workload to be deployed on spoke cluster
+// +union
+type SpokeWorkloadTemplate struct {
+	// +unionDiscriminator
+	// Type defines type of the workload deployed on the spoke cluster
+	Type SpokeWorkloadType `json:"type,omitempty" protobuf:"bytes,1,opt,name=type"`
+
+	// Manifests represents a list of kuberenetes resources to be deployed on the spoke cluster.
+	// +optional
+	Manifests []runtime.RawExtension `json:"manifests,omitempty" protobuf:"bytes,2,rep,name=objects"`
```
deads2k: prototag needs fixing

apps/v1/types.go
```diff
@@ -0,0 +1,194 @@
...
+type SpokeWorkloadSpec struct {
+	// Workload represents the resource workload to be deployed on spoke cluster
+	Workload SpokeWorkloadTemplate `json:"workload,omitempty" protobuf:"bytes,1,opt,name=workload"`
+}
+
+// SpokeWorkloadTemplate represents the resource workload to be deployed on spoke cluster
+// +union
+type SpokeWorkloadTemplate struct {
+	// +unionDiscriminator
+	// Type defines type of the workload deployed on the spoke cluster
```
deads2k: indicate required

apps/v1/types.go
```diff
@@ -0,0 +1,194 @@
...
+	Type SpokeWorkloadType `json:"type,omitempty" protobuf:"bytes,1,opt,name=type"`
```
deads2k: cannot omit

apps/v1/types.go
```diff
@@ -0,0 +1,194 @@
...
+	// Name is the namespace of the kubernetes resource
+	// +optional
+	Namespace string `json:"namespace,omitempty" protobuf:"bytes,6,opt,name=namespace"`
+}
+
+// SpokeWorkloadStatus represents the current status of spoke workload
+type SpokeWorkloadStatus struct {
+	// Conditions contains the different condition statuses for this work.
+	// Valid condition types are:
+	// 1. WorkloadApplied represents workload is applied successfully on spoke cluster.
```
deads2k: I wouldn't prefix these with `Workload`.  These names happen to match with our existing vocabulary for clusteroperators

apps/v1/types.go
```diff
@@ -0,0 +1,194 @@
...
+	// 2. WorkloadProgressing represents workload is being applied on spoke cluster.
+	// 3. WorkloadAvailable represents workload exists on the spoke cluster.
+	// 4. WorkloadDegraded represents workload is destroyed unexpectedly on spoke cluster.
```
deads2k: destroyed isn't quite right.  Try cribbing the meaning from https://github.com/openshift/api/blob/master/config/v1/types_cluster_operator.go#L142-L168  Just a lot shorter

apps/v1/types.go
```diff
@@ -0,0 +1,194 @@
...
+	Conditions []StatusCondition `json:"conditions" protobuf:"bytes,1,rep,name=conditions"`
+
+	// WorkloadCondition represents the conditions of the workload deployed on
+	// spoke cluster. The agent on spoke cluster syncs the condition from spoke to the hub.
+	// +optional
+	WorkloadCondition SpokeWorkloadCondition `json:"workloadCondition,omitempty" protobuf:"bytes,2,rep,name=workloadCondition"`
```
deads2k: `workloadStatus` I think

apps/v1/types.go
deads2k: Just to clarify, even in status you need consumers to be able to distinguish and unions are assumed to be encapsulated.  You can't have a discriminator in another field.

apps/v1/types.go
```diff
@@ -0,0 +1,194 @@
...
+// +union
+type SpokeWorkloadCondition struct {
+	// +unionDiscriminator
+	// Type defines the type of workload applied on the spoke cluster
+	Type SpokeWorkloadType `json:"type,omitempty" protobuf:"bytes,1,opt,name=type"`
+
+	// Manifests represents the condition of manifests deployed on spoke cluster.
+	// Valid condition types are:
+	// 1. Applied represents the resource is applied successfully on spoke cluster.
+	// 2. Available represents the resource exists on the spoke cluster.
```
deads2k: I think this will need progressing too.

deads2k:
I like them.

deads2k:
I did not make back through with a fine-tooth comb today, but I am quite happy with my skimming.  I think you can feel confident that you're close enough to build on top of this in a PR branch.

work/v1/types.go
```diff
@@ -0,0 +1,206 @@
...
+	Resource ResourceTemplate `json:"resource,omitempty" protobuf:"bytes,1,opt,name=resource"`
+}
+
+// ResourceTemplate represents the resource workload to be deployed on spoke cluster
+// +union
+type ResourceTemplate struct {
+	// +unionDiscriminator
+	// Type defines type of the workload deployed on the spoke cluster
+	// +required
+	Type WorkloadType `json:"type" protobuf:"bytes,1,opt,name=type"`
```
pmorie: I am thinking that a type field for the this struct is not the exact right way to go - i could see `manifests` not being portable to all types of work. Instead, I would suggest a different flavor of union type like so:

```go
+union
type ResourceTemplate struct {
    Manifest *ManifestWorkload `json:"manifest"`
}

type ManifestWorkload struct {
    Manifests []runtime.RawExtension `json:"manifests"`
}
```

As we add workload types, we can add fields to `ResourceTemplate` that fully encapsulate the different kinds of work.

work/v1/types.go
```diff
@@ -0,0 +1,206 @@
...
+	// +optional
+	Manifests []runtime.RawExtension `json:"manifests,omitempty" protobuf:"bytes,2,rep,name=manifests"`
+}
+
+// WorkloadType represents the type of workload deployed on spoke
+type WorkloadType string
+
+const (
+	// ManifestsWorkType means that the work is to deploy manifests on the spoke cluster.
+	ManifestsWorkType WorkloadType = "Manifests"
```
pmorie: This const won't be necessary with changes to `ResourceTemplate`

work/v1/types.go
```diff
@@ -0,0 +1,206 @@
...
+)
+
+// WorkloadResourceMeta represents the gvk, gvr, name and namespace of a resoure
+type WorkloadResourceMeta struct {
```
pmorie: Is this really `ManifestResourceMeta` ?

work/v1/types.go
```diff
@@ -0,0 +1,206 @@
...
+	// ResourceStatus represents the status of each resource in workload deployed on
+	// spoke cluster. The agent on spoke cluster syncs the condition from spoke to the hub.
+	// +optional
+	ResourceStatus WorkloadResourceStatus `json:"resourceStatus,omitempty" protobuf:"bytes,2,rep,name=resourceStatus"`
+}
+
+// WorkloadResourceStatus represents the status of each resource in workload deployed on
+// spoke cluster
+// +union
+type WorkloadResourceStatus struct {
```
pmorie: Similar refactor here to `ResourceTemplate`.

pmorie (commented):
Left some suggestions about `ResourceTemplate`, curious to know your thoughts @deads2k

pmorie:
I am asking myself now whether we should be using union types, or just have a simple `ManifestWork` type, and add different types of work in different API resources as we build them out. My gut is telling me union typing might be premature.

WDYT @deads2k @qiujian16

pmorie:
After some additional thought, I think that we should make the API resource in this PR focused around manifests exclusively and drop the union-type dimension. I think this has the best chance of aligning with a future upstream k8s work API as well as yielding a simpler API to understand, program to, and implement. We can support different flavors of work each in their own API resources as we build them out, which will keep controllers simpler, more focused, and allow for more parallel development (and graduation).

So, I would suggest:

- Call this resource `ManifestWork`
- Drop the union types and type fields entirely

deads2k:
> I am asking myself now whether we should be using union types, or just have a simple `ManifestWork` type, and add different types of work in different API resources as we build them out. My gut is telling me union typing might be premature.
> 
> WDYT @deads2k @qiujian16

`ManifestWork` sounds reasonable to me.

docs/clusterjoinprocess.md
```diff
@@ -21,15 +21,14 @@ and deploy the agent on spoke cluster.
...
-management:clusterName:agentName"}`.
+`{"Organization": ["system:open-cluster-management:clusterName"], "CommonName":"system:open-cluster-management:clusterName:agentName"}`.
   - The name of the csr is the digest of subject and private key, with a common prefix.
   CSR will specify the signer name as the kube-client one.
 4. cluster-admin on hub-cluster approve the CSR.
 5. hub-controller creates a clusterrolebinding on the hub with the identity of
 `system:open-cluster-management:clusterName:agentName`
    - Allows status update of `SpokeCluster`
-6. cluster-admin on hub update `spec.accepeSpokeCluster` to `true`.
+6. cluster-admin on hub update `spec.acceptSpokeCluster` to `true`.
```
pmorie, line 31: :+1:

docs/spokeworkload.md
```diff
@@ -0,0 +1,23 @@
+# SpokeWorkload API
+
```
pmorie, line 2: We'll want to change this a little but I am not a believer in doc blocking development, we can tune this md file up in a follow-on.

hack/update-codegen.sh
```diff
@@ -28,3 +28,12 @@ for group in cluster; do
     --go-header-file ${SCRIPT_ROOT}/hack/boilerplate.txt \
     ${verify}
 done
+
+for group in work; do
```
pmorie, line 32: you can just add `work` after `cluster` on line 23

pmorie:
This is looking very solid

/approve

hack/update-codegen.sh
deads2k: > you can just add `work` after `cluster` on line 23

followup is acceptable, but yes

work/v1/doc.go
```diff
@@ -0,0 +1,9 @@
+// Package v1 contains API Schema definitions for the operators v1 API group
+// +k8s:deepcopy-gen=package,register
+// +k8s:conversion-gen=github.com/open-cluster-management/api/apps
```
deads2k, line 3: `api/work`

you'll find it later

work/v1/types.go
```diff
@@ -0,0 +1,186 @@
...
+type ManifestWorkSpec struct {
+	// Workload represents the manifest workload to be deployed on spoke cluster
+	Workload ManifestsTemplate `json:"workload,omitempty" protobuf:"bytes,1,opt,name=workload"`
+}
+
+// ManifestsTemplate represents the manifest workload to be deployed on spoke cluster
+type ManifestsTemplate struct {
+	// Manifests represents a list of kuberenetes resources to be deployed on the spoke cluster.
+	// +optional
+	Manifests []runtime.RawExtension `json:"manifests,omitempty" protobuf:"bytes,2,rep,name=manifests"`
```
deads2k, line 62: it's weird to start at "bytes,2"  if you change it really fast you can change it post-merge

PR #18, opened 2020-04-26
spoke cluster health check

skeeey:
introduce a new condition `SpokeClusterConditionAPIServerAvailable` to describe whether the kube-apiserver of the spoke cluster is available


cluster/v1/types.go
```diff
@@ -96,6 +96,9 @@ const (
 	// SpokeClusterConditionHubDenied means the request to join the cluster is denied by
 	// cluster-admin on hub
 	SpokeClusterConditionHubDenied string = "HubDeniedSpoke"
+	// SpokeClusterConditionAPIServerAvailable means the kube-apiserver of the spoke cluster is
+	// available
+	SpokeClusterConditionAPIServerAvailable string = "SpokeClusterAPIServerAvailable"
```
pmorie: Is this really testing whether the API is available, or whether the spoke agent is running?

skeeey: I plant to use `healthz` API to check the apiserver status, I think if the `healthz` API returns ok, I can consider the apiserver is available, and I check the apiserver status in agent, so I think if I can check the status, the spoke agent should be running

docs/spokeclusterhealthcheck.md
```diff
@@ -0,0 +1,25 @@
...
+Actors
+
+- agent on spoke cluster
+- hub controller on hub cluster
+
+## Health check
+
+1. after a spoke cluster is accepted by hub cluster admin, the agent on spoke cluster will
+create a [`Lease`](https://github.com/kubernetes/api/blob/master/coordination/v1/types.go#L27)
+on its namespace on hub cluster, then the agent will periodically (by default 10s)
```
deads2k: I think we need to talk about scale and how you plan to manage this.  If you have 5000 spokes, what you're describing here would be 500 qps doing nothing but heartbeating.  As with nodes, this presents a serious problem. The upstream KEP is here: https://github.com/kubernetes/enhancements/blob/master/keps/sig-node/0009-node-heartbeat.md .  A smaller resource (like coordination) helps to some degree because you aren't fanning out every watcher with large objects, but that is a lot of QPS for an idle system.

If you want to do this, I think it's important to describe what it will be used to do, how it is better or different than metrics, and then provide some meaningful way to manage the scale, QPS, and managing/changing update frequency from the hub to attempt to prevent overrun.

pmorie: +1 to David's comments.

> provide some meaningful way to manage the scale, QPS, and managing/changing update frequency from the hub to attempt to prevent overrun.

A knob is critical for this, also using Jitter on the frequency of the update. We should be able to tell users how to plan for a specific QPS number.

@deads2k what do you think is a reasonable qps to allocate for heartbeating?

skeeey: @deads2k and @pmorie, sorry for late

for QPS, if we have a reasonable QPS (or we can allow user set it), I think we can use `Lease.LeaseDurationSeconds` to coordinate the QPS

1. hub controller gets the count of spoke clusters and calculate the average lease duration time with a reasonable QPS and updates the `Lease` object for each spoke cluster , e.g. if the expected QPS is 100 and there are 5000 clusters, we set the `Lease.LeaseDurationSeconds` to 50s
2. spoke agent calculates its lease renew interval time by `Lease.LeaseDurationSeconds`, refer to kubelet implementation, we give a renew interval fraction (e.g. 0.25), we use `Lease.LeaseDurationSeconds * Fraction` to get the renew interval time, for above example, the renew interval time will be 12.5s

> how it is better or different than metrics

for migration case, we want to automatically migrate the workloads among the spoke clusters when some spoke clusters are unavailable,  if we use metrics, the hub controller need periodically poll all of spoke clusters to know whether a spoke cluster is available, I think this is an inefficient way, if we have a status to present whether a spoke cluster is available, the hub controller can list them on hub cluster, I think this is more efficient (except the spoke cluster api health, we may need to know the status of other spoke agents, so we may change the `SpokeClusterConditionAPIServerAvailable` to `SpokeClusterConditionAvailable`)

What's your opinions?

qiujian16: should we consider adding a `LeaseDuration`  field in SpokeCluster API? So cluster-admin on hub is able to change that, and spoke agent can adjust the lease accordingly.

I think we need the condition on spokecluster also because reporting metrics will require additional request path from hub to spoke (or spoke to hub). With this condition, we can at least know whether registration/work agent is running with the minimum deployment.

docs/spokeclusterhealthcheck.md
```diff
@@ -0,0 +1,60 @@
...
+on spoke cluster, a new condition `SpokeClusterConditionAvailable` will be introduced to describe
+whether a spoke cluster is available
+
+If the kube-apiserver is health and the registration/work agent is running with the minimum
+deployment on a spoke cluster, this spoke cluster will be considered available
+
+The Kubernetes Metrics API may be used to get the status of spoke cluster kube-apiserver or spoke
+registration/work agent, but Metrics API has some limits for our cases
+
+- The hub may not access the spoke cluster, so we cannot use Kubernetes Metrics API to
```
deads2k: there is a push model

docs/spokeclusterhealthcheck.md
```diff
@@ -0,0 +1,60 @@
...
+get spoke cluster metrics from hub cluster
+- Reporting metrics will require additional request path from hub to spoke (or spoke to hub)
```
deads2k: you're already talking about additional requests

docs/spokeclusterhealthcheck.md
```diff
@@ -0,0 +1,60 @@
...
+1. after a spoke cluster is accepted by hub cluster admin, the agent on spoke cluster will
+create a [`Lease`](https://github.com/kubernetes/api/blob/master/coordination/v1/types.go#L27)
+on its namespace on hub cluster, then the agent will periodically (by default 10s, the period time
+should be adjusted, see the scale problem section)
+    - update the `RenewTime` of `Lease` to keep available, and
+    - check the spoke cluster kube-apiserver health by `healthz` API, and check the
+    registration/work agent is running with the minimum deployment, if all of them are
+    available the agent make sure the status of `SpokeClusterConditionAvailable`
+    condition is `true`, otherwise, the agent updates the status to `false`
+2. hub controller lists the spoke clusters on hub cluster periodically (by default 5s), for
```
deads2k: an informer ends up taking many watch events, but the smearing may help.  I guess I would at least try an informer to start because it's "normal".

You probably want to label these leases to avoid selecting leases for nodes.

docs/spokeclusterhealthcheck.md
```diff
@@ -0,0 +1,60 @@
...
+each accepted spoke cluster, the controller get its `Lease` from its namespace and check
+wether the `Lease` is constantly updated in a duration (by default 40s), if not, the
```
deads2k: 40s seems really tight to update. if we round up to 60s
1. 60 spokes is 1qps
2. 600 spokes 10 qps
3. 6000 spokes 100 qps

How responsive do you need this to be?   Seems like o(hour) is good enough for most production use-cases and won't crush a server.

docs/spokeclusterhealthcheck.md
```diff
@@ -0,0 +1,60 @@
...
+controller updates the status of this spoke cluster `SpokeClusterConditionAvailable`
+condition to `unknown`
+
+## The scale problem
+
+If there are many spoke clusters, we will face there are a lot of QPS for an idle system due to
+spoke lease update request, to solve this problem, we need a way to control the frequency of
+lease update, we propose two alternative ways
+
+- Adding `LeaseDuration` field in `SpokeCluster` API, the hub cluster admin is able to update it
```
deads2k: This could work

deads2k:
In the end, I think we should consider what subset of a problem this solves for us.  Is it truly valuable to know the spoke kube-apiserver is readyz==ok?

I worry that something like this would grow into desiring a projection of things like clusteroperators or the status of other hub/spoke communicator.  To my mind, that is an anti-goal, those extensions bits should never be included in a core API like this.

If we can agree that this will never expand further and perhaps we even start another repo to project  clusteroperator.status back into the spoke namespace, I think that would be useful.  I may even try to do that separately because it would require building out a way to provide creds and exercise manifestwork.

/approve

docs/spokeclusterhealthcheck.md
pmorie: I think 5 minutes is the lower bound for me on update period - I could believe 30 minutes or an hour would be entirely sufficient for many use cases.

pmorie:
> If we can agree that this will never expand further and perhaps we even start another repo to project clusteroperator.status back into the spoke namespace, I think that would be useful.

Agree. I also do not want us to tack extra status fields onto this API that relate to arbitrary other components. I think we are describing an API principle here: the status of SpokeCluster reflects only the status pertaining to that cluster's registration and agent.

Should we formalize that concept in the doc for SpokeCluster?

cluster/v1/types.go
```diff
@@ -53,6 +53,11 @@ type SpokeClusterSpec struct {
 	// deleted.
 	// +required
 	HubAcceptsClient bool `json:"hubAcceptsClient" protobuf:"bytes,2,opt,name=hubAcceptsClient"`
+
+	// LeaseDurationSeconds is used to coordinate the lease update time of spoke agents.
+	// If it is empty, the spoke agent will update its lease per 10s
```
qiujian16: maybe 1 min by default, if we think the default offline time is 5 min.

docs/spokeclusterhealthcheck.md
```diff
@@ -0,0 +1,60 @@
...
+according to system QPS, and spoke agent can adjust the lease update time accordingly
+- Dynamically adjust the QPS for each spoke cluster, suppose we have a reasonable QPS, we may use
```
qiujian16: i think we can remove this option.

docs/spokeclusterhealthcheck.md
```diff
@@ -0,0 +1,60 @@
...
+each accepted spoke cluster, the controller get its `Lease` from its namespace and check
+wether the `Lease` is constantly updated in a duration (by default 5min), if not, the
+controller updates the status of this spoke cluster `SpokeClusterConditionAvailable`
+condition to `unknown`
+
+## The scale problem
+
+If there are many spoke clusters, we will face there are a lot of QPS for an idle system due to
+spoke lease update request, to solve this problem, we need a way to control the frequency of
+lease update, we propose two alternative ways
```
qiujian16: we will only use lease method

docs/spokeclusterhealthcheck.md
```diff
@@ -0,0 +1,60 @@
...
+    - update the `RenewTime` of `Lease` to keep available, and
+    - check the spoke cluster kube-apiserver health by `healthz` API, and check the
+    registration/work agent is running with the minimum deployment, if all of them are
+    available the agent make sure the status of `SpokeClusterConditionAvailable`
+    condition is `true`, otherwise, the agent updates the status to `false`
+2. hub controller watches the spoke clusters on hub cluster with its informer, for
+each accepted spoke cluster, the controller get its `Lease` from its namespace and check
+wether the `Lease` is constantly updated in a duration (by default 5min), if not, the
+controller updates the status of this spoke cluster `SpokeClusterConditionAvailable`
+condition to `unknown`
```
qiujian16, line 41: Specify the scope here will be registration agent only. Other agent despite of registration/work should add additional customized conditontype here.

cluster/v1/types.go
```diff
@@ -53,6 +53,11 @@ type SpokeClusterSpec struct {
 	// deleted.
 	// +required
 	HubAcceptsClient bool `json:"hubAcceptsClient" protobuf:"bytes,2,opt,name=hubAcceptsClient"`
+
+	// LeaseDurationSeconds is used to coordinate the lease update time of spoke agents.
+	// If it is empty, the spoke agent will update its lease per 1 minute
```
deads2k: If you allow both nil and zero you have to explain the difference between them.  If there is no difference, don't use a pointer.

For this particular field, I don't think I see value in allowing a zero.

deads2k:
Please update the specification for the int before merge.  I'll leave lgtm with @qiujian16 

/assign @qiujian16

PR #19, opened 2020-04-28
Add nucleus operator API

qiujian16:


hack/update-codegen.sh
```diff
@@ -37,3 +37,12 @@ for group in work; do
     --go-header-file ${SCRIPT_ROOT}/hack/boilerplate.txt \
     ${verify}
 done
+
+for group in nucleus; do
```
pmorie: make line 24 in this file say `for group in cluster work nucleus; do` and remove the duplicate blocks

nucleus/v1/types.go
```diff
@@ -0,0 +1,187 @@
...
+	// Status represents the current status of nucleus hub
+	// +optional
+	Status NucleusHubStatus `json:"status,omitempty" protobuf:"bytes,3,opt,name=status"`
+}
+
+// NucleusHubSpec represents a desired deployment configuration of nucleus hub.
+type NucleusHubSpec struct {
+	// Components represents the components to be installed for nucleus hub.
+	// +required
+	Components NucleusHubComponents `json:"components" protobuf:"bytes,1,opt,name=components"`
```
deads2k: We should not imply that there are components.  I don't think this infrastructure should use component style language or people will think that they add components.  I don't think we will add any components.  Instead of components, we need to make the deployment across spokes really hum using a local OLM to manage installed components and provide a solid pattern for other operators to use to provide hub/spoke mechanics if they need them.

nucleus/v1/types.go
```diff
@@ -0,0 +1,187 @@
...
+
+// RegistrationComponent represents the desired configuration of registration controller installed on hub.
+type RegistrationComponent struct {
+	CommonComponentConfig `json:",inline" protobuf:"bytes,1,opt,name=commonComponentConfig"`
+}
+
+// CommonComponentConfig represents the common configuration of the component.
+type CommonComponentConfig struct {
+	// Replicas is the number of desired replicas
+	Replicas int32 `json:"replica,omitempty" protobuf:"varint,1,opt,name=replicas"`
```
deads2k: this looks questionable to me and as a concept will prevent a proper HPA from functioning.

nucleus/v1/types.go
```diff
@@ -0,0 +1,187 @@
...
+	// Image is the image name of the component. If not set, default imagee is used.
+	// +optional
+	Image string `json:"image,omitempty" protobuf:"bytes,2,opt,name=image"`
+
+	// ImagePullPolicy is the policy to pull image.
+	// +optional
+	ImagePullPolicy corev1.PullPolicy `json:"pullPolicy,omitempty" protobuf:"bytes,3,opt,name=imagePullPolicy"`
```
deads2k: we haven't found a need to provide this capability in openshift.  Just use IfNotPresent.

nucleus/v1/types.go
```diff
@@ -0,0 +1,187 @@
...
+
+	// ResourceRequirements represents the resource requirement of the component.
+	// +optional
+	ResourceRequirements corev1.ResourceRequirements `json:"resourceRequirements,omitempty" protobuf:"bytes,4,opt,name=resourceRequirements"`
```
deads2k: We haven't yet found a need in openshift even with all our operators.  If this comes up, we can add it, but I made the mistake of trying to add this early and regretted it.

nucleus/v1/types.go
```diff
@@ -0,0 +1,187 @@
...
+
+// NucleusAgentSpec represents the desired deployment configuratioin of nucleus agent.
+type NucleusAgentSpec struct {
+	// Components represents the agent components to be installed ono spoke cluster.
+	// +optional
+	Components NucleusAgentComponents `json:"components,omitempty" protobuf:"bytes,1,opt,name=components"`
+}
+
+// NucleusAgentComponents represents the desireed configuration of nucleus agent.
+type NucleusAgentComponents struct {
```
deads2k: same comment about not having components.

nucleus/v1/types.go
```diff
@@ -0,0 +1,187 @@
...
+
+// RegistrationAgentComponent reprensents the desired configuration of registration
+// agent on spoke cluster.
+type RegistrationAgentComponent struct {
+	CommonComponentConfig `json:",inline" protobuf:"bytes,3,opt,name=commonComponentConfig"`
+
+	// BootstrapKubeconfigSecret is the kubeconfig of hub to bootstrap the process that
+	// agent joins hub.
+	// +required
+	BootstrapKubeconfigSecret string `json:"bootsrapKubeconfigSecret" protobuf:"bytes,1,opt,name=bootstrapKubeconfigSecret"`
```
deads2k: place this directly in spec

nucleus/v1/types.go
```diff
@@ -0,0 +1,187 @@
...
+
+	// ClusterName is the name of the spoke cluster to be created on hub.
+	// The spoke agent generates a random name if it is not set.
```
deads2k: or discovers the appropriate cluster name on openshift.

nucleus/v1/types.go
```diff
@@ -0,0 +1,187 @@
...
+
+	// Spec represents the desired deployment configuratioin of nucleus agent.
+	Spec NucleusAgentSpec `json:"spec,omitempty" protobuf:"bytes,1,opt,name=spec"`
+
+	// Status represents the current status of nucleus agent.
+	Status NucleusAgentStatus `json:"status,omitempty" protobuf:"bytes,2,opt,name=status"`
+}
+
+// NucleusAgentSpec represents the desired deployment configuratioin of nucleus agent.
+type NucleusAgentSpec struct {
```
deads2k: to allow connections to multiple hubs, you must have a namespace.  Choose some name restrictions it (prefix) and a default value, but I suspect an optional namespace will be important.

qiujian16: I think that means the the api (also nucleushub api) should be cluster scoped. And nucleushub api should be singleton?

nucleus/v1/types.go
qiujian16: i am not sure we need hpa here, but I agree with you that by default we should have 3 replicas with leader elect.

nucleus/v1/types.go
pmorie: +1 on avoiding characterization as components. The API here should be _just_ what is necessary to deploy the most essential pieces of the system, which currently would be `registration` and `work`.

I think you can just inline `Registration` and add `Work` as a sibling.

nucleus/v1/types.go
pmorie: I think we can remove this for now and explore exact needs for this concern in the future, let's just use leader election with 3 replicas to start

nucleus/v1/types.go
pmorie: I don't expect that this will need to ripple out to the control plane.

nucleus/v1/types.go
```diff
@@ -0,0 +1,187 @@
...
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// +genclient
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+// +kubebuilder:subresource:status
+
+// NucleusHub represents a deployment of nucleus hub component.
+type NucleusHub struct {
```
pmorie: Let's finalize the name before we merge this PR. 'nucleus' is a term I found convenient to illustrate the concept but not a name I would want in production.

Some ideas:

- ControlPlane
- Hub
- HubCore
- ControlPlaneCore

When we think about the name, think about something that would be good to layer product onto in a different resource(s), rather than a place where we will grow a lot of knobs for product. The intent should be that this resource will stay small and deal only with the most essential parts of the system, and contain no product decisions.

nucleus/v1/types.go
pmorie: I think that means that we should:

- pick a default namespace
- restrict names to be the default or an optional suffix on the default
- allow specification of the namespace in the API

I'm not sure my @deads2k model could predict whether this would yield a namespaced resource or not. I would _guess_ cluster scoped.

nucleus/v1/types.go
```diff
@@ -0,0 +1,187 @@
...
+	Registration RegistrationAgentComponent `json:"registration" protobuf:"bytes,1,opt,name=registration"`
+
+	// Work represents the desired configuration of work agent.
+	// +optional
+	Work WorkComponent `json:"work,omitempty" protobuf:"bytes,2,opt,name=work"`
+}
+
+// RegistrationAgentComponent reprensents the desired configuration of registration
+// agent on spoke cluster.
+type RegistrationAgentComponent struct {
```
pmorie: I would call this `RegistrationAgentSpec` and inline the common fields

pmorie (commented):
A couple feedbacks, let's decide on a descriptive, not overly cute name instead of 'nucleus'

nucleus/v1/types.go
```diff
@@ -0,0 +1,178 @@
...
+	Status HubCoreStatus `json:"status,omitempty" protobuf:"bytes,3,opt,name=status"`
+}
+
+// HubCoreSpec represents a desired deployment configuration of nucleus hub.
+type HubCoreSpec struct {
+	// Namespace is the namespace to deploy the agent.
+	// The namespace must have a prefix of "open-cluster-management-", and if it is not set,
+	// the namespace of "open-cluster-management" is used to deploy agent.
+	// +optional
+	Namespace string `json:"namespace,omitempty" protobuf:"bytes,1,opt,name=namespace"`
```
deads2k: a hub cluster can only be a single hub because the namespaces collide.  I don't think this one needs a namespace, we'll simply choose one for it.  Sorry if I put my previous comment in teh wrong spot.

nucleus/v1/types.go
```diff
@@ -0,0 +1,178 @@
...
+
+	// Registration represents the desired configuration of registration controller installed on hub.
+	// +required
+	Registration RegistrationHubSpec `json:"registration" protobuf:"bytes,2,opt,name=registration"`
```
deads2k: Let's just put `registrationImagePullSpec` right here.

nucleus/v1/types.go
```diff
@@ -0,0 +1,178 @@
...
+}
+
+// HubCoreStatus represents the current status of nucleus hub.
+type HubCoreStatus struct {
+	// Conditions contain the different condition statuses for this work.
```
deads2k: rephrase into Available, Degraded, Progressing please.

nucleus/v1/types.go
```diff
@@ -0,0 +1,178 @@
...
+}
+
+// +genclient
+// +genclient:nonNamespaced
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+// +kubebuilder:subresource:status
+// +kubebuilder:resource:scope=Cluster
+
+// AgentCore represents a deployment of nucleus agent core on spoke cluster.
+type AgentCore struct {
```
deads2k: To be consistent, this would be a SpokeCore, right?

nucleus/v1/types.go
```diff
@@ -0,0 +1,178 @@
...
+	// Spec represents the desired deployment configuratioin of nucleus agent.
+	Spec AgentCoreSpec `json:"spec,omitempty" protobuf:"bytes,2,opt,name=spec"`
+
+	// Status represents the current status of nucleus agent.
+	Status AgentCoreStatus `json:"status,omitempty" protobuf:"bytes,3,opt,name=status"`
+}
+
+// AgentCoreSpec represents the desired deployment configuratioin of nucleus agent.
+type AgentCoreSpec struct {
+	// Namespace is the namespace to deploy the agent.
```
deads2k, line 82: yeah, this is what I meant, thanks.

nucleus/v1/types.go
```diff
@@ -0,0 +1,178 @@
...
+	// The namespace must have a prefix of "open-cluster-management-", and if it is not set,
+	// the namespace of "open-cluster-management" is used to deploy agent.
+	// +optional
+	Namespace string `json:"namespace,omitempty" protobuf:"bytes,1,opt,name=namespace"`
+
+	// Registration represents the desired configuration of registration agent.
+	// +required
+	Registration RegistrationAgentSpec `json:"registration" protobuf:"bytes,2,opt,name=registration"`
```
deads2k: `registrationImagePullSpec` and `workImagePullSpec` please

nucleus/v1/types.go
```diff
@@ -0,0 +1,178 @@
...
+
+	// Work represents the desired configuration of work agent.
+	// +optional
+	Work WorkAgentSpeec `json:"work,omitempty" protobuf:"bytes,3,opt,name=work"`
+
+	// BootstrapKubeconfigSecret is the kubeconfig of hub to bootstrap the process that
+	// agent joins hub.
+	// +required
+	BootstrapKubeconfigSecret string `json:"bootsrapKubeconfigSecret" protobuf:"bytes,4,opt,name=bootstrapKubeconfigSecret"`
```
deads2k: I don't think we need to configure this.  I think we can simply require it to be in a fixed location for our operator.  Say `bootstrap-hub-kubeconfig`

nucleus/v1/types.go
```diff
@@ -0,0 +1,178 @@
...
+// RegistrationAgentSpec reprensents the desired configuration of registration
+// agent on spoke cluster.
+type RegistrationAgentSpec struct {
+	CommonComponentConfig `json:",inline" protobuf:"bytes,1,opt,name=commonComponentConfig"`
+}
+
+// AgentCoreStatus represents the current status of nucleus agent.
+type AgentCoreStatus struct {
+	// Conditions contain the different condition statuses for this work.
+	// Valid condition types are:
```
deads2k, line 105: Available, Progressing, Degraded

deads2k:
looking pretty close to me.

cluster/install.go
```diff
@@ -1,26 +0,0 @@
...
-	"k8s.io/apimachinery/pkg/runtime/schema"
-
-	clusterv1 "github.com/open-cluster-management/api/cluster/v1"
-)
-
-const (
-	GroupName = "cluster.open-cluster-management.io"
-)
-
-var (
```
pmorie, line 14: Was this meant to be deleted?

nucleus/v1/register.go
```diff
@@ -0,0 +1,39 @@
+package v1
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+)
+
+var (
+	GroupName     = "nucleus.open-cluster-management.io"
```
pmorie, line 10: Should we change the name of this API group?

nucleus/v1/types.go
```diff
@@ -0,0 +1,166 @@
...
+	// Status represents the current status of nucleus hub
+	// +optional
+	Status HubCoreStatus `json:"status,omitempty" protobuf:"bytes,3,opt,name=status"`
+}
+
+// HubCoreSpec represents a desired deployment configuration of nucleus hub.
+type HubCoreSpec struct {
+	// Registration represents the desired configuration of registration controller installed on hub.
+	// +required
+	Registration RegistrationImagePullSec `json:"registration" protobuf:"bytes,1,opt,name=registration"`
```
pmorie: s/Sec/Spec

nucleus/v1/types.go
```diff
@@ -0,0 +1,166 @@
...
+	// +optional
+	metav1.ListMeta `json:"metadata,omitempty" protobuf:"bytes,1,opt,name=metadata"`
+
+	// Items is a list of nucleus hub.
+	Items []HubCore `json:"items" protobuf:"bytes,2,rep,name=items"`
+}
+
+// RegistrationImagePullSec represents the desired image configuration of registration to be installed.
+type RegistrationImagePullSec struct {
+	CommonComponentConfig `json:",inline" protobuf:"bytes,1,opt,name=commonComponentConfig"`
```
pmorie: Inline the `image` field into `RegistrationImagePullSpec`

nucleus/v1/types.go
```diff
@@ -0,0 +1,166 @@
...
+
+// CommonComponentConfig represents the common configuration of the component.
+type CommonComponentConfig struct {
+	// Image is the image name of the component. If not set, default imagee is used.
+	// +optional
+	Image string `json:"image,omitempty" protobuf:"bytes,1,opt,name=image"`
+}
+
+// WorkAgentImagePullSec represents the desired configuration of work agent.
+type WorkAgentImagePullSec struct {
```
pmorie: Inline the image field

pmorie (commented):
Looking good, left a couple minor comments.

nucleus/v1/types.go
```diff
@@ -0,0 +1,166 @@
...
+	// Spec represents a desired deployment configuration of nucleus hub
+	Spec HubCoreSpec `json:"spec" protobuf:"bytes,2,opt,name=spec"`
+
+	// Status represents the current status of nucleus hub
+	// +optional
+	Status HubCoreStatus `json:"status,omitempty" protobuf:"bytes,3,opt,name=status"`
+}
+
+// HubCoreSpec represents a desired deployment configuration of nucleus hub.
+type HubCoreSpec struct {
```
deads2k, line 28:
```go
type HubCoreSpec struct {
    RegistrationImagePullSpec string
}
```

nucleus/v1/types.go
```diff
@@ -0,0 +1,166 @@
...
+
+	// Spec represents the desired deployment configuratioin of nucleus agent.
+	Spec SpokeCoreSpec `json:"spec,omitempty" protobuf:"bytes,2,opt,name=spec"`
+
+	// Status represents the current status of nucleus agent.
+	Status SpokeCoreStatus `json:"status,omitempty" protobuf:"bytes,3,opt,name=status"`
+}
+
+// SpokeCoreSpec represents the desired deployment configuratioin of nucleus spoke agent.
+type SpokeCoreSpec struct {
```
deads2k, line 81:
```go
type SpokeCoreSpec struct {
    ClusterName string
    Namespace string
    RegistrationImagePullSpec string
    WorkImagePullSpec string
}
```

cluster/install.go
qiujian16: yes, I think it is not necessary? The Install is cluster/v1/register.go. I do not think we need another func here.

nucleus/v1/types.go
```diff
@@ -0,0 +1,149 @@
...
+)
+
+// +genclient
+// +genclient:nonNamespaced
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+// +kubebuilder:subresource:status
+// +kubebuilder:resource:scope=Cluster
+
+// HubCore represents a deployment of nucleus hub core component.
+// HubCore will only deployed in open-cluster-managemeent namespace.
```
deads2k: nit: management

nucleus/v1/types.go
```diff
@@ -0,0 +1,149 @@
...
+	// the namespace of "open-cluster-management" is used to deploy agent.
+	// +optional
+	Namespace string `json:"namespace,omitempty" protobuf:"bytes,1,opt,name=namespace"`
+
+	// RegistrationImagePullSpec represents the desired image configuration of registration agent.
+	// +required
+	RegistrationImagePullSpec string `json:"registrationImagePullSpec" protobuf:"bytes,2,opt,name=registrationImagePullSpec"`
+
+	// WorkImagePullSpec represents the desired image configuration of work agent.
+	// +optional
```
deads2k: minor: required

deads2k:
I have followups I want quickly, but I'm willing to merge this to unblock progress.

/lgtm
/hold

holding for pmorie to have a say.

PR #20, opened 2020-05-05
to fix trivia in version checking

sdminonne:
@pmorie as far as I understand this is needed to have something working


sdminonne:
Sorry, realized afterwards code is vendored. This https://github.com/openshift/build-machinery-go/pull/15 is (hopefully) better off

PR #22, opened 2020-05-09
Add sever url in agentcore

qiujian16:
Need this to be set from API


nucleus/v1/types.go
```diff
@@ -97,6 +97,22 @@ type SpokeCoreSpec struct {
 	// The spoke agent generates a random name if it is not set, or discovers the appropriate cluster name on openshift.
 	// +optional
 	ClusterName string `json:"clusterName,omitempty" protobuf:"bytes,4,opt,name=clusterName"`
+
+	// ExternalAPIServer represents the apiserver url and ca bundle that is accessible externally
+	// +optional
+	ExternalAPIServer ServerURL `json:"externalSeverURL,omitempty" protobuf:"bytes,5,opt,name=externalSeverURL"`
```
deads2k: I'm thinking this should be a slice to accommodate multiples and you should explain what empty slice means.

pmorie: Typo: `externalServerURL`

Agree that this should be a slice.

qiujian16: update to slice, also change the `spokeClientConfig` on spoke cluster to slice.

qiujian16: update to slice, also change the `spokeClientConfig` on spoke cluster to slice.

PR #23, opened 2020-05-11
Enable preserveUnknowField

qiujian16:
This is to set preserveUnknowField in work api to true, since we
have runtime.RawExtension in the spec.


deads2k:
ask @damemi and/or @sttts if this is the right way to do this.  It doesn't look right to me because I would expect the `preserveUnknown` to be on *only* the fields that are un-schemad and this looks to be on everything

/hold

sttts:
> since we have runtime.RawExtension in the spec.

Are these RawExtension real objects with `TypeMeta` and `ObjectMeta`? For that purpose we have `x-kubernetes-embeded-resource: true`. For true raw JSON data, there is `x-kubernetes-preserve-unknown-fields: true`.

qiujian16:
These are real objects with `TypeMeta` and `ObjectMeta`

sttts:
Here is an example: https://github.com/kubernetes-sigs/controller-tools/pull/345/files#diff-862d6e06a6705fcf1b7782d96f1e003aR128

work/v1/0000_00_work.open-cluster-management.io_manifestworks.crd.yaml
```diff
@@ -43,14 +43,18 @@ spec:
...
+              items:
+                description: ManifestsTemplate represents the manifest workload to
+                  be deployed on spoke cluster
+                type: object
+                properties:
+                  manifest:
+                    description: Manifest represents a resource to be deployed on
+                      the spoke
                     type: object
+                    x-kubernetes-preserve-unknown-fields: true
```
sttts, line 56: fyi, this means you can put in any object type (pods, services, builds, anything...). If you know this type more concretely, you could also specify the schema here instead of `x-kubernetes-preserve-unknown-fields: true`.

deads2k: > fyi, this means you can put in any object type (pods, services, builds, anything...). If you know this type more concretely, you could also specify the schema here instead of `x-kubernetes-preserve-unknown-fields: true`.

it really is an anything field.

work/v1/types.go
```diff
@@ -52,14 +52,16 @@ type ManifestWork struct {
...
 	// Workload represents the manifest workload to be deployed on spoke cluster
-	Workload ManifestsTemplate `json:"workload,omitempty" protobuf:"bytes,1,opt,name=workload"`
+	Workload []ManifestsTemplate `json:"workload,omitempty" protobuf:"bytes,1,opt,name=workload"`
 }
 
 // ManifestsTemplate represents the manifest workload to be deployed on spoke cluster
 type ManifestsTemplate struct {
-	// Manifests represents a list of kuberenetes resources to be deployed on the spoke cluster.
-	// +optional
-	Manifests []runtime.RawExtension `json:"manifests,omitempty" protobuf:"bytes,2,rep,name=manifests"`
```
deads2k, line 62: @sttts can we keep this a slice and have each member of the slice unchecked?

sttts: I believe you have to type alias `runtime.RawExtension` with controller-tools and annotate the new type with `+kubebuilder:validation:EmbeddedResource` and `+kubebuilder:pruning:PreserveUnknownFields`.

sttts:
```Golang
Manifests []Manifest `json:"manifests,omitempty" protobuf:"bytes,2,rep,name=manifests"`

// +kubebuilder:validation:EmbeddedResource
// +kubebuilder:pruning:PreserveUnknownFields
type Manifest runtime.RawExtension
```

qiujian16: it seems not work, the generated crd has a field like
```
manifest:
     description: Manifests represents a list of kuberenetes resources
          to be deployed on the spoke cluster.
      type: array
      items:
         $ref: '#/definitions/k8s.io~1apimachinery~1pkg~1runtime~0RawExtension'
```

qiujian16: inlining RawExtension works. Please help to check if it is a valid pattern.

PR #25, opened 2020-05-20
Update nucleus default ns

qiujian16:


nucleus/v1/types.go
```diff
@@ -81,7 +81,7 @@ type SpokeCore struct {
 type SpokeCoreSpec struct {
 	// Namespace is the namespace to deploy the agent.
 	// The namespace must have a prefix of "open-cluster-management-", and if it is not set,
-	// the namespace of "open-cluster-management" is used to deploy agent.
+	// the namespace of "open-cluster-management-core" is used to deploy agent.
```
deads2k: sorry, i should have noticed this before. If we do this, then cleanup becomes more complicated (cannot simply delete the namespace).  One more change.  Can you do `open-cluster-management-{spoke,hub}`?

qiujian16: i think we cannot delete open-cluster-management-spoke namespace, because it needs to be created in advance, and a bootstrap-hub-kubeconfig secret needs to be created in this namespace. But yes for open-cluster-management-hub.

PR #26, opened 2020-05-22
add tracking for created resources related to manifest work

deads2k:
Tracking created resources separately from currently managed ones allows us to independently calculate when what we created doesn't match spec and have the current field for reflecting individual creation status (which may fail on metadata gathering) match well.

I think the alternative is adding an admission plugin to ensure well-formedness and then using the status field in a way that isn't parallel to spec.


work/v1/types.go
```diff
@@ -100,6 +100,30 @@ type ManifestResourceMeta struct {
 	Namespace string `json:"namespace" protobuf:"bytes,7,opt,name=namespace"`
 }
 
+// CreatedManifestResourceMeta represents the gvr, name and namespace of a resource
```
pmorie, line 103: I'd end these two lines with a period, they're sentences.

pmorie:
You need to go mod vendor, I think

work/v1/types.go
```diff
@@ -115,6 +139,14 @@ type ManifestWorkStatus struct {
 	// spoke cluster. The agent on spoke cluster syncs the condition from spoke to the hub.
 	// +optional
 	ResourceStatus ManifestResourceStatus `json:"resourceStatus,omitempty" protobuf:"bytes,2,rep,name=resourceStatus"`
+
+	// CreatedResources represents every resource that has been created for this ManifestWork and has not yet been deleted.
```
qiujian16, line 143: will `CreatedResources` include the resource that  already  exist in the spoke cluster, so manifestwork is defined to update that resource rather than create it?

deads2k: > will `CreatedResources` include the resource that already exist in the spoke cluster, so manifestwork is defined to update that resource rather than create it?

Well considered question.  I think so, yes.  On a successful mutation call, the `CreatedResources` will be updated.  Can you think of a better name or a better way to phrase the doc?

pmorie: `TrackedResources` ?

PR #30, opened 2020-06-01
Update s/SpokeCluster/ManagedCluster/g

mdelder:
+ Update cluster/v1 API to be consistent with ManagedCluster naming.
+ Minor edits to documentation


operator/v1/types.go
```diff
@@ -93,37 +93,37 @@ type KlusterletSpec struct {
...
-	// If it is set empty, spoke cluster has no externally accessible url that hub cluster can visit.
+	// If it is set empty, managed cluster has no externally accessible url that hub cluster can visit.
 	// +optional
 	ExternalServerURLs []ServerURL `json:"externalServerURLs,omitempty" protobuf:"bytes,5,opt,name=externalServerURLs"`
 }
 
 // ServerURL represents the apiserver url and ca bundle that is accessible externally
 type ServerURL struct {
-	// URL is the url of apiserver endpoint of the spoke cluster.
+	// URL is the url of apiserver endpoint of the manageed cluster.
```
pmorie: typo: managed

work/v1/types.go
```diff
@@ -125,40 +125,40 @@ type AppliedManifestResourceMeta struct {
 	Namespace string `json:"namespace" protobuf:"bytes,5,opt,name=namespace"`
 }
 
-// ManifestWorkStatus represents the current status of spoke manifest workload
+// ManifestWorkStatus represents the current status of managed manifest workload
```
pmorie: Want to make sure this is intentional, reads a little awkwardly. `managed cluster`?

operator/v1/types.go
```diff
@@ -81,7 +81,7 @@ type Klusterlet struct {
 type KlusterletSpec struct {
 	// Namespace is the namespace to deploy the agent.
 	// The namespace must have a prefix of "open-cluster-management-", and if it is not set,
-	// the namespace of "open-cluster-management-spoke" is used to deploy agent.
+	// the namespace of "open-cluster-management-agent" is used to deploy agent.
```
pmorie, line 84: reminder to plumb this into the registration operator repo when you get there

operator/v1/types.go
```diff
@@ -93,37 +93,37 @@ type KlusterletSpec struct {
 	// +required
 	WorkImagePullSpec string `json:"workImagePullSpec,omitempty" protobuf:"bytes,3,opt,name=workImagePullSpec"`
 
-	// ClusterName is the name of the spoke cluster to be created on hub.
-	// The spoke agent generates a random name if it is not set, or discovers the appropriate cluster name on openshift.
+	// ClusterName is the name of the managed cluster to be created on hub.
+	// The klusterlet agent generates a random name if it is not set, or discovers the appropriate cluster name on openshift.
```
pmorie: We should decide now whether we want to conventionally capitalize `klusterlet` or not.

pmorie (commented):
Found a couple nits

operator/v1/types.go
mdelder: Good call. I opted for `Klusterlet`.

cluster/v1/types.go
```diff
@@ -9,107 +9,108 @@ import (
...
 }
 
-// SpokeClusterSpec provides the information to securely connect to a remote server
+// ManagedClusterSpec provides the information to securely connect to a remote server
 // and verify its identity.
-type SpokeClusterSpec struct {
-	// SpokeClientConfigs represents a list of the apiserver address of the spoke cluster.
-	// If it is empty, spoke cluster has no accessible address to be visited from hub.
+type ManagedClusterSpec struct {
+	// ManagedClientConfigs represents a list of the apiserver address of the managed cluster.
```
qiujian16: ManagedClusterClientConfig?

itdove:
@mdelder  FYI This changes will request cycles in the rcm-controller because:
1) The CRD changes and the type.go change... so need to adapt the code.
2) In the code to be coherent we use the word "spoke" at many places (from Go variables to Go files) and so all need to be adapted to have a clear and compressible code.

mdelder:
@itdove Agreed. There is a PR in flight to adopt the API updates in the `nucleus` operator (where the repo will also be renamed).

pmorie:
> In the code to be coherent we use the word "spoke" at many places (from Go variables to Go files) and so all need to be adapted to have a clear and compressible code.

@itdove IMO there is no need to change comments, I think the meaning will remain clear.

itdove:
I guess also issues must be open for each squad to make this change and also for the doc team. For me it is not very nice to have different names for the same thing, it just drive to confusion. Also because it is not impossible that in the future that name will change again ;-)

mdelder:
@qiujian16 Updated `ManagedClientConfig` to `ManagedClusterClientConfig`. Also caught there the CRD filename had been missed.

I believe this PR is ready to be merged, barring any additional feedback from reviewers.

qiujian16:
/assign pmorie
/assign deads2k

/lgtm 
leave to @pmorie @deads2k for any more comments.

pmorie:
> I guess also issues must be open for each squad to make this change and also for the doc team. For me it is not very nice to have different names for the same thing, it just drive to confusion. 

I totally get it. I think that 'spoke' is a meaningful var / param name to use, whatever the name of the API resource, and I wouldn't feel like it was wrong to see those terms in code. I'll probably continue to say 'spoke' in conversation because it is specific and shorter than the API resource name.

> Also because it is not impossible that in the future that name will change again ;-)

Totally possible, all the more reason to avoid exhaustively rooting it out in places like code where it's fine as-is.

mdelder:
@pmorie Are you okay to approve this PR?

PR #31, opened 2020-06-04
Add generation in statuses

qiujian16:
This is to follow the same pattern in openshift api


operator/v1/types.go
```diff
@@ -33,14 +33,41 @@ type ClusterManagerSpec struct {
...
+
 	// Conditions contain the different condition statuses for this ClusterManager.
 	// Valid condition types are:
 	// Applied: components in hub are applied.
 	// Available: components in hub are available and ready to serve.
 	// Progressing: components in hub are in a transitioning state.
 	// Degraded: components in hub do not match the desired configuration and only provide
 	// degraded service.
-	Conditions []StatusCondition `json:"conditions" protobuf:"bytes,1,opt,name=conditions"`
+	Conditions []StatusCondition `json:"conditions" protobuf:"bytes,2,opt,name=conditions"`
```
deads2k: you cannot change the proto tag serialization order (that int after bytes)

operator/v1/types.go
```diff
@@ -33,14 +33,41 @@ type ClusterManagerSpec struct {
 
 // ClusterManagerStatus represents the current status of the registration and work distribution controllers running on the hub.
 type ClusterManagerStatus struct {
+	// ObservedGeneration is the last generation change you've dealt with
+	// +optional
+	ObservedGeneration int64 `json:"observedGeneration,omitempty" protobuf:"bytes,1,opt,name=observedGeneration"`
```
deads2k: this one wouldn't be bytes

deads2k: > this one wouldn't be bytes

varint I think

deads2k:
I'm fine with the change, but you cannot break your protobuf serialization.  You may want to consider simply removing it since CRDs cannot make use of it.

I would also suggest adding `.status.relatedResources` so that `oc adm inspect` works.

operator/v1/types.go
pmorie: We should just remove proto for now.

operator/v1/types.go
```diff
@@ -41,6 +45,65 @@ type ClusterManagerStatus struct {
...
+
+	// namespace is where the thing you're tracking is
+	// +optional
+	Namespace string `json:"namespace"`
+
+	// name is the name of the thing you're tracking
+	// +required
+	Name string `json:"name"`
+
+	// lastGeneration is the last generation of the controller involved
```
pmorie: I'm not sure what this is supposed to mean, exactly. Is it the last generation the controller observed, or last generation the controller created?

qiujian16: This is the last generation the controller created. Change to lastCreatedGeneration?

pmorie: I think the name is okay, let's just make sure the doc says that it's the last generation created by the controller.

PR #34, opened 2020-06-22
Add applied manifestwork API

qiujian16:


work/v1/types.go
```diff
@@ -225,3 +225,68 @@ type ManifestWorkList struct {
...
+// AppliedManifestWorkStatus represents the current status of AppliedManifestWork
+type AppliedManifestWorkStatus struct {
+	// AppliedResources represents a list of resources defined within the manifestwork that are applied.
+	// Only resources with valid GroupVersionResource, namespace, and name are suitable.
+	// An item in this slice is deleted when there is no mapped manifest in manifestwork.Spec or by finalizer.
+	// The resource relating to the item will also be removed from managed cluster.
+	// The deleted resource may still be present until the finalizers for that resource are finished.
+	// However, the resource will not be undeleted, so it can be removed from this list and eventual consistency is preserved.
+	// +optional
+	AppliedResources []AppliedManifestResourceMeta `json:"appliedResources,omitempty"`
```
elgnay, line 274: After we introduce `AppliedManifestWork`, applied resources of ManifestWork are stored in two different places: `ManifestWork.Status.AppliedResrouces` and `AppliedManifestWork.Status.AppliedResrouces`. Changes of this slice may come from `ManifestWork` when applying manifests, or from `AppliedManifestWork` when deleting applied resources. It's complicated to sync them with each other. So I suggest we only keep one copy of this slice in `AppliedManifestWork.Status`.

work/v1/types.go
```diff
@@ -225,3 +225,68 @@ type ManifestWorkList struct {
...
+	Spec AppliedManifestWorkSpec `json:"spec,omitempty"`
+
+	// Status represents the current status of AppliedManifestWork
+	// +optional
+	Status AppliedManifestWorkStatus `json:"status,omitempty"`
+}
+
+// AppliedManifestWorkSpec represents the desired configuration of AppliedManifestWork
+type AppliedManifestWorkSpec struct {
+	// HubID represents the ID of the hub that the agent fetches ManifestWork. It could
```
pmorie: I think we want to be more specific about what this is - where does this correspond to a type defined elsewhere in this repo?

work/v1/types.go
```diff
@@ -225,3 +225,68 @@ type ManifestWorkList struct {
...
+	// be a hash of hub kubernetes apiserver url.
+	// +required
+	HubID string `json:"hubID"`
+
+	// ManifestWorkName represents the name of the related manifestwork on hub.
+	// +required
+	ManifestWorkName string `json:"manifestWorkName"`
+
+	// ClusterName represents the name of the cluster on hub where the related manifestwork
```
pmorie, line 259: Is this semantically the cluster name, or the namespace corresponding to the cluster?

work/v1/types.go
pmorie: That's a good point and I'm still thinking about whether we need to keep the information in a single place, or whether we should clearly define the precedence of AppliedManifestWork and ManifestWork.Status

deads2k: I think this is a good point.  Since we only act on the information on the spoke, I think it's a good point.  It does mean that the update is non-atomic, but I think that's ok.

work/v1/types.go
deads2k: > I think we want to be more specific about what this is - where does this correspond to a type defined elsewhere in this repo?

Perhaps we can avoid it by declaring it a hub hash created by hashing the *first* hub kube apiserver URL.

work/v1/types.go
```diff
@@ -225,3 +225,68 @@ type ManifestWorkList struct {
...
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// AppliedManifestWork represents an applied manifestwork on managed cluster. It is placed
+// on managed cluster. An AppliedManifestWork links to a manifestwork on a hub recording resources
+// deployed in the managed cluster.
+// When the agent is removed from managed cluster, cluster-admin on managed cluster
+// can delete appliedmanifestwork to remove resources deployed by the agent.
+type AppliedManifestWork struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
```
deads2k, line 238: I'd like the name to be forced to include the hub hash so that we can support multiple hubs.  I'd also like to see a force label so we can efficiently select.

work/v1/types.go
```diff
@@ -225,3 +225,68 @@ type ManifestWorkList struct {
...
+
+	// Status represents the current status of AppliedManifestWork
+	// +optional
+	Status AppliedManifestWorkStatus `json:"status,omitempty"`
+}
+
+// AppliedManifestWorkSpec represents the desired configuration of AppliedManifestWork
+type AppliedManifestWorkSpec struct {
+	// HubID represents the ID of the hub that the agent fetches ManifestWork. It could
+	// be a hash of hub kubernetes apiserver url.
```
deads2k: indicate that it is a hash to ensure no one can use this as a backdoor API to get the hub apiserver URL.

work/v1/types.go
deads2k: Perhaps, .metadata.name == `<hubhash>-<manifestwork name>`

work/v1/types.go
```diff
@@ -225,3 +225,68 @@ type ManifestWorkList struct {
...
+	HubID string `json:"hubID"`
+
+	// ManifestWorkName represents the name of the related manifestwork on hub.
+	// +required
+	ManifestWorkName string `json:"manifestWorkName"`
+
+	// ClusterName represents the name of the cluster on hub where the related manifestwork
+	// is placed.
+	// +required
+	ClusterName string `json:"clusterNamee"`
```
deads2k, line 262: this seems unnecessary since it will be the same for every instance in the cluster.

qiujian16: I am not sure if there is a case that there are multiple agents on one spoke that register to a single hub?

work/v1/types.go
qiujian16: it is semantically the cluster name, the name of the namespace corresponding to the cluster has the same value.

work/v1/types.go
qiujian16: Will a generate name with manifestwork name as prefix works also? Since we already have hubhash in the spec.
labels makes sense.

deads2k: > Will a generate name with manifestwork name as prefix works also? Since we already have hubhash in the spec.
> labels makes sense.

That makes it possible to accidentally create two appliedmanifestworks for the same underlying manifestwork.  We don't want that to be possible.

work/v1/types.go
deads2k: > I am not sure if there is a case that there are multiple agents on one spoke that register to a single hub?

I don't think that will behave properly at an API level since you cannot predict which will service you.  Let's start without it.

work/v1/types.go
```diff
@@ -231,3 +222,68 @@ type ManifestWorkList struct {
...
+	HubHash string `json:"hubHash"`
+
+	// ManifestWorkName represents the name of the related manifestwork on hub.
+	// +required
+	ManifestWorkName string `json:"manifestWorkName"`
+
+	// ClusterName represents the name of the cluster on hub where the related manifestwork
+	// is placed.
+	// +required
+	ClusterName string `json:"clusterNamee"`
```
deads2k, line 262: also, the json tag is typoed

deads2k:
I'd like the name to be semantically described pretty tightly to avoid accidents and I think clustername should be removed, but since I'm about to go on a long weekend, I'll tag so that you can make progress.


/lgtm
/approve

PR #35, opened 2020-06-24
Replace DeletionTimestamp with UID in AppliedManifestResourceMeta

elgnay:


qiujian16:
/hold
until we figure out  whether criticalresource works as expected.

PR #36, opened 2020-06-29
Add server print for managed cluster

qiujian16:
fixes https://github.com/open-cluster-management/backlog/issues/3125

### DIFF
--- a/cluster/v1/0000_00_clusters.open-cluster-management.io_managedclusters.crd.yaml
+++ b/cluster/v1/0000_00_clusters.open-cluster-management.io_managedclusters.crd.yaml
@@ -3,13 +3,29 @@ kind: CustomResourceDefinition
 metadata:
   name: managedclusters.cluster.open-cluster-management.io
 spec:
+  additionalPrinterColumns:
+  - JSONPath: .spec.hubAcceptsClient
+    name: Hub Accepted
+    type: boolean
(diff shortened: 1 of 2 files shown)


qiujian16:
output will be like this

```
[centos@yzw open-cluster-management]$ kubectl get managedcluster
NAME       HUB ACCEPTED   MANAGED CLUSTER URLS   JOINED   AVAILABLE   AGE
cluster1   true           https://localhost      True     True        16m
```

cluster/v1/0000_00_clusters.open-cluster-management.io_managedclusters.crd.yaml
```diff
@@ -3,13 +3,29 @@ kind: CustomResourceDefinition
 metadata:
   name: managedclusters.cluster.open-cluster-management.io
 spec:
+  additionalPrinterColumns:
+  - JSONPath: .spec.hubAcceptsClient
+    name: Hub Accepted
+    type: boolean
+  - JSONPath: .spec.managedClusterClientConfigs[*].url
```
pmorie, line 10: (I believe `additionalPrinterColumns` is an ordered field that dictates the ordering of these fields in CLI, if not, please correct me).

We should think carefully about the ordering of these fields; I would probably have the condition-based fields, then this field, then age.

pmorie:
/approve
/lgtm

we can tweak column ordering in the future if necessary

PR #37, opened 2020-07-03
Update appliedmanifestwork

qiujian16:
followup of pr 34

### DIFF
--- a/work/v1/types.go
+++ b/work/v1/types.go
@@ -233,6 +233,8 @@ type ManifestWorkList struct {
 // deployed in the managed cluster.
 // When the agent is removed from managed cluster, cluster-admin on managed cluster
 // can delete appliedmanifestwork to remove resources deployed by the agent.
+// The name of the appliedmanifestwork must be in the format of
+// {hash of hub's first kube-apiserver url}-{manifestwork name}
 type AppliedManifestWork struct {
 	metav1.TypeMeta   `json:",inline"`
 	metav1.ObjectMeta `json:"metadata,omitempty"`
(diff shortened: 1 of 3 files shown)


work/v1/types.go
```diff
@@ -233,6 +233,8 @@ type ManifestWorkList struct {
 // deployed in the managed cluster.
 // When the agent is removed from managed cluster, cluster-admin on managed cluster
 // can delete appliedmanifestwork to remove resources deployed by the agent.
+// The namee of the appliedmanifestwork must be in the format of
+// {hash of hub's first kube-apiserver url}-{manifestwork name}
```
pmorie, line 237: typo: namee :)

pmorie: The comment raises an interesting question I hadn't thought of before... does this effectively mean that the first kube-apiserver URL must be immutable?

qiujian16: good point, what we want is a unique immutable ID to identify hub. For OCP,  maybe we can also use the cluster id of hub cluster?

deads2k:
> The comment raises an interesting question I hadn't thought of before... does this effectively mean that the first kube-apiserver URL must be immutable?

> good point, what we want is a unique immutable ID to identify hub. For OCP, maybe we can also use the cluster id of hub cluster?

I'm open to alternative suggestions of the hubhash.   Some ideas
1. hub apiserver URL - this has problems with changing the URLs
2. hub ID hash - I don't think we actually track a concept of hub id
3. hash of spoke namespace running work - since this is the namespace that manages a set, this may be our closest approximation. I think it also enables adoption  use-cases and hub apiserver location changes.

/lgtm
/approve
/hold

the hold is just to make sure you read and consider this comment. Feel free to release after making a decision.

qiujian16:
> hash of spoke namespace running work - since this is the namespace that manages a set, this may be our closest
> approximation. I think it also enables adoption use-cases and hub apiserver location changes.

Using namespace implies that each work agent needs to be run in different namespaces, and it might be not appropriate to use hubhash as the field name? I think what we need is a way to identify different agents rather different hubs. So it might be call it `AgentIdentifier` rather than `HubHash`? Also should we hash klusterlet.Name which also can be used to identify different agents.

deads2k:
> klusterlet.Name which also can be used to identify different agents.

sure, this seems fine.  Get the doc updated and this is ready to merge

/hold cancel

PR #49, opened 2020-07-17
open-cluster-management/backup#3613 Desc rev

cadawson:
Description review.  Note: I am not sure what "gvr" is in this. We should spell it out, unless it is very common. Based on a later review, is this "group, version, resource"?


cadawson:
Redone in a collective PR, #66 .

PR #50, opened 2020-07-17
open-cluster-management/backlog#3613 Desc rev

cadawson:
Description review.  Note: I am not sure what "gvr" is in this. We should spell it out, unless it is very common. It looks like it might be "group, version, resource", maybe?


cadawson:
Redone in a collective PR, #66 .

PR #55, opened 2020-08-18
Add ManagedClusterSet API

elgnay:


cluster/v1/types.go
```diff
@@ -170,3 +170,44 @@ type ManagedClusterList struct {
...
+	// Clusters represents a set of managed clusters in the cluster set
+	Clusters ClusterSelector `json:"clusterSelector,omitempty"`
+}
+
+// ClusterSelector represents a selector to select a set of managed clusters
+type ClusterSelector struct {
+	// ClusterNames represents a list of cluster name
+	ClusterNames []string `json:"clusterNames,omitempty"`
+
+	// LabelSelector represents a label selector to select cluster by label
```
qiujian16: +optional

I think we need to doc what is meaning when labelSelector is empty

cluster/v1/0000_01_clusters.open-cluster-management.io_managedclustersets.crd.yaml
```diff
@@ -0,0 +1,103 @@
...
+    controller-gen.kubebuilder.io/version: v0.2.5
+  creationTimestamp: null
+  name: managedclustersets.cluster.open-cluster-management.io
+spec:
+  group: cluster.open-cluster-management.io
+  names:
+    kind: ManagedClusterSet
+    listKind: ManagedClusterSetList
+    plural: managedclustersets
+    singular: managedclusterset
```
qiujian16: should be clusterscoped

cluster/v1/types.go
```diff
@@ -170,3 +170,44 @@ type ManagedClusterList struct {
...
+// ManagedClusterSetSpec describes the attributes of the desired managed clusters
+type ManagedClusterSetSpec struct {
+	// Clusters represents a set of managed clusters in the cluster set
+	Clusters ClusterSelector `json:"clusterSelector,omitempty"`
+}
+
+// ClusterSelector represents a selector to select a set of managed clusters
+type ClusterSelector struct {
+	// ClusterNames represents a list of cluster name
+	ClusterNames []string `json:"clusterNames,omitempty"`
```
qiujian16: +optional

cluster/v1/types.go
```diff
@@ -170,3 +170,44 @@ type ManagedClusterList struct {
...
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	// Spec defines the attributes of the desired managed clusters
+	Spec ManagedClusterSetSpec `json:"spec"`
+}
+
+// ManagedClusterSetSpec describes the attributes of the desired managed clusters
+type ManagedClusterSetSpec struct {
+	// Clusters represents a set of managed clusters in the cluster set
+	Clusters ClusterSelector `json:"clusterSelector,omitempty"`
```
qiujian16: I think the variable name here should better be ClusterSelector

cluster/v1/0000_01_clusters.open-cluster-management.io_managedclustersets.crd.yaml
elgnay: Fixed

cluster/v1/types.go
elgnay: Fixed

cluster/v1/types.go
elgnay: Updated.

cluster/v1/types.go
elgnay: Comment added.

cluster/v1/types.go
```diff
@@ -170,3 +170,52 @@ type ManagedClusterList struct {
...
+	// Spec defines the attributes of the desired managed clusters
+	Spec ManagedClusterSetSpec `json:"spec"`
+}
+
+// ManagedClusterSetSpec describes the attributes of the desired managed clusters
+type ManagedClusterSetSpec struct {
+	// ClusterSelector represents a selector to select a set of managed clusters
+	// If empty, the cluster set will include all managed clusters
+	// +optional
+	ClusterSelector ClusterSelector `json:"clusterSelector,omitempty"`
```
deads2k: I think you want to allow a slice of these so that ORs are easy to express.

cluster/v1/types.go
```diff
@@ -170,3 +170,52 @@ type ManagedClusterList struct {
...
 	Items []ManagedCluster `json:"items"`
 }
+
+// +genclient
+// +genclient:nonNamespaced
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+// +kubebuilder:resource:scope="Cluster"
+
+// ManagedClusterSet is a set of managed clusters.
+type ManagedClusterSet struct {
```
deads2k: I recommend providing a status with conditions so you could report things like whether the label selector was invalid and perhaps the number of clusters selected.

cluster/v1/types.go
elgnay: Updated.

cluster/v1/types.go
elgnay: Added.

cluster/v1alpha1/types.go
```diff
@@ -0,0 +1,100 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// +genclient
+// +genclient:nonNamespaced
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+// +kubebuilder:subresource:status
+// +kubebuilder:resource:scope="Cluster"
+
+// ManagedClusterSet is a set of managed clusters.
```
pmorie: What properties can we describe here that will be useful to know?

cluster/v1alpha1/types.go
```diff
@@ -0,0 +1,100 @@
...
+
+	// Spec defines the attributes of the desired managed clusters
+	Spec ManagedClusterSetSpec `json:"spec"`
+
+	// Status represents the current status of the cluster set
+	// +optional
+	Status ManagedClusterSetStatus `json:"status,omitempty"`
+}
+
+// ManagedClusterSetSpec describes the attributes of the desired managed clusters
```
pmorie: nit (and you don't have to change this now, I know @cadawson is working on this repo and he seems like he might be interested in these details): should we say ManagedCluster instead of "managed cluster"?

cluster/v1alpha1/types.go
```diff
@@ -0,0 +1,100 @@
...
+	// LabelSelector represents a label selector to select cluster by label
+	// +optional
+	LabelSelector *metav1.LabelSelector `json:"labelSelector,omitempty"`
+}
+
+// ManagedClusterSetStatus represents the current status of the cluster set.
+type ManagedClusterSetStatus struct {
+	// ClusterNames contains names of all the selected clusters.
+	// +optional
+	ClusterNames []string `json:"clusterNames,omitempty"`
```
pmorie: I wonder if we really want to reflect this in status. On the surface it absolutely makes sense, I wonder about how it would scale in storage:

- 1MB upper limit on etcd records means a soft upper limit on number of clusters that can be represented in the entire object
- What happens if you select too many clusters to write into the status?

cluster/v1alpha1/types.go
```diff
@@ -0,0 +1,100 @@
...
+
+	// Message is a human-readable message indicating details about the last status change.
+	// +required
+	Message string `json:"message"`
+}
+
+const (
+	// ManagedClusterSetConditionValid means the managed cluster set, especially
+	// the cluster selectors, are valid.
+	ManagedClusterSetConditionValid string = "ClusterSetValid"
```
pmorie: I'm not exactly sure what this condition means - if it's the sense of API validation, why wouldn't we just reject invalid selectors in admission?

qiujian16: i am thinking like if no clusters is selected by the labelSelector in the spec, it is invalid.

cluster/v1alpha1/types.go
qiujian16: Maybe we could add things like this
ManagedClusterSet defines a group of ManagedCluster that user's workload can run on. A workload can be defined to deployed on a ManagedClusterSet, which mean
1. the workload can run on any ManagedCluster in the ManagedClusterSet 
2. the workload cannot run on any ManagedCluster outside the ManagedClusterSet
3. the service exposed by the workload can be shared in any ManagedCluster in the ManagedClusterSet

elgnay: Comments updated.

cluster/v1alpha1/types.go
elgnay: Updated.

cluster/v1alpha1/types.go
elgnay: Removed from status.

cluster/v1alpha1/types.go
elgnay: Renamed ManagedClusterSetConditionValid to ManagedClusterSetConditionSelected

pmorie: > i am thinking like if no clusters is selected by the labelSelector in the spec, it is invalid.

Interesting, I'm not sure i would call this 'invalid', but you might want to surface it in the API. I might call it 'empty'. This could be an entirely valid state to be in though. Maybe you're draining a cluster group for some duration and planning to move clusters back into it?

elgnay: Renamed it one more time. `ManagedClusterSetConditionEmpty string = "ClusterSetEmpty"`

cluster/v1alpha1/types.go
```diff
@@ -0,0 +1,101 @@
...
+
+	// LabelSelector represents a label selector to select cluster by label
+	// +optional
+	LabelSelector *metav1.LabelSelector `json:"labelSelector,omitempty"`
+}
+
+// ManagedClusterSetStatus represents the current status of the ManagedClusterSet.
+type ManagedClusterSetStatus struct {
+	// Conditions contains the different condition statuses for this ManagedClusterSet.
+	Conditions []StatusCondition `json:"conditions"`
```
deads2k, line 57: `metav1.Condition` came in kube 1.19

cluster/v1alpha1/types.go
```diff
@@ -0,0 +1,101 @@
...
+
+	// Message is a human-readable message indicating details about the last status change.
+	// +required
+	Message string `json:"message"`
+}
+
+const (
+	// ManagedClusterSetConditionEmpty means no ManagedCluster is included in the
+	// ManagedClusterSet.
+	ManagedClusterSetConditionEmpty string = "ClusterSetEmpty"
```
deads2k, line 86: Like paul, I wonder about the utility of this vs representing how many it selected as an integer and condition saying you matched.  But I can go with it for now.

PR #56, opened 2020-08-20
Add ManagedClusterSetBinding API

elgnay:


cluster/v1alpha1/types.go
```diff
@@ -0,0 +1,40 @@
...
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// +genclient
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+// +kubebuilder:resource:scope="Namespaced"
+
+// ManagedClusterSetBinding projects a managed cluster set into a certain namespace.
```
pmorie: Good start at the description, I do think that we should enrich this with information about how users can expect to use these resources. Do they act the same as ManagedClusters? What do they allow you to do? I know we're still sorting these properties out, but let's record the key ones here (in subsequent PRs) as we do.

cluster/v1alpha1/types.go
```diff
@@ -0,0 +1,43 @@
...
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	// Spec defines the attributes of ManagedClusterSetBinding.
+	Spec ManagedClusterSetBindingSpec `json:"spec"`
+}
+
+// ManagedClusterSetBindingSpec defines the attributes of ManagedClusterSetBinding.
+type ManagedClusterSetBindingSpec struct {
+	// ClusterSet is the name of the ManagedClusterSet to bind.
```
deads2k: you need more doc around what dangling references do, whether there are any native restrictions against the clusterset name (can anyone bind any clusterset).  Whether instance names are required to match the clusterset name.

deads2k:
/approve

The doc needs improvement.  I'll leaving tagging lgtm with @qiujian16 

/assign @qiujian16

cluster/v1alpha1/types.go
elgnay: Doc added

cluster/v1alpha1/types.go
```diff
@@ -99,3 +99,44 @@ type ManagedClusterSetList struct {
...
+
+	// Spec defines the attributes of ManagedClusterSetBinding.
+	Spec ManagedClusterSetBindingSpec `json:"spec"`
+}
+
+// ManagedClusterSetBindingSpec defines the attributes of ManagedClusterSetBinding.
+type ManagedClusterSetBindingSpec struct {
+	// ClusterSet is the name of the ManagedClusterSet to bind. The instance name of the
+	// ManagedClusterSetBinding doesn't have to match the target ClusterSet. User is
+	// allowed to bind a ManagedClusterSet if they have an RBAC rule to GET on the
```
qiujian16: I would say like ```User is allowed to set or update this field if they have an RBAC rule...```

elgnay: Updated

PR #57, opened 2020-08-20
add managedclusteraddon.addon.open-cluster-management.io

TheRealHaoLiu:


addon/v1/types.go
```diff
@@ -0,0 +1,178 @@
...
+}
+
+// ManagedClusterAddonStatus provides information about the status of the operator.
+// +k8s:deepcopy-gen=true
+type ManagedClusterAddonStatus struct {
+	// conditions describes the state of the operator's managed and monitored components.
+	// +patchMergeKey=type
+	// +patchStrategy=merge
+	// +optional
+	Conditions []AddonStatusCondition `json:"conditions,omitempty"  patchStrategy:"merge" patchMergeKey:"type"`
```
deads2k: kube 1.19 has `metav1.Condition` as a type.  I recommend using that.

addon/v1/types.go
```diff
@@ -0,0 +1,178 @@
...
+
+	// relatedObjects is a reference to the configuration resource that drives the operator
+	// +optional
+	RelatedObject ObjectReference `json:"relatedObject,omitempty"`
```
deads2k: I suggest a slice

addon/v1/types.go
```diff
@@ -0,0 +1,178 @@
...
+
+	// LatestRelease indicates the latest available release for the addon
+	LatestRelease Release `json:"availableUpdate,omitempty"`
```
deads2k: Why release instead of Version?  I think openshift is pretty consistent in the use of the word Version.

addon/v1/types.go
```diff
@@ -0,0 +1,178 @@
...
+
+	// CurrentRelease ...
```
deads2k: you should provide guidance about how to handle "what version is this" during an upgrade when multiple versions are involved.  The rule changed recently in openshift to be, "update the currentVersion **after** you are fully rolled out"

addon/v1/types.go
deads2k: > Why release instead of Version? I think openshift is pretty consistent in the use of the word Version.

nm, I see below.

addon/v1/types.go
```diff
@@ -0,0 +1,178 @@
...
+	Version string `json:"version"`
+
+	// imageManifest is list of images will be used for the specific version
+	// this will be use for backward compatability (i.e hub upgraded but agent have not)
+	// +required
+	ImageManifest []Image `json:"imageManifest"`
+}
+
+type Image struct {
+	Name  string `json: "name"`
```
deads2k: The distinction between name and Image needs to be described.  You mean ImagePullSpec perhaps?  Is it required to be a SHA?

addon/v1/types.go
```diff
@@ -0,0 +1,178 @@
...
+
+	// status holds the information about the state of an operator.  It is consistent with status information across
+	// the Kubernetes ecosystem.
+	// +optional
+	Status ManagedClusterAddonStatus `json:"status"`
+}
+
+// ManagedClusterAddonSpec is empty for now, but you could imagine holding information like "pause".
+type ManagedClusterAddonSpec struct {
+	// UpdateApproved ...
```
deads2k: Can you describe this field in more detail?

addon/v1/types.go
```diff
@@ -0,0 +1,178 @@
...
+	UpdateApproved bool `json:"updateApproved"`
```
deads2k: Nothing is a bool.  `UpdateApprovalStatus` as a string with enumerated values.  "Approved", "AutomaticApproval", "AwaitingApproval"

addon/v1/types.go
```diff
@@ -0,0 +1,178 @@
...
+const (
+	// Available indicates that the operand (eg: openshift-apiserver for the
+	// openshift-apiserver-operator), is functional and available in the cluster.
+	Available AddonStatusConditionType = "Available"
+
+	// Progressing indicates that the operator is actively rolling out new code,
+	// propagating config changes, or otherwise moving from one steady state to
+	// another.  Operators should not report progressing when they are reconciling
+	// a previously known state.
+	Progressing AddonStatusConditionType = "Progressing"
```
deads2k: Given the combination of clusteroperator and clusterversion here, I would expect to see an `Upgrading` condition type too.

addon/v1/types.go
TheRealHaoLiu: in the product currently there's a "image-manifest" that contain the information that maps a fixed key to a image. long story...

so the motivation behind this part is that we can record the images used in the current version and when hub upgrade and the hub's image-manifest for the previous version is ripped away we can still manage the agent

TheRealHaoLiu: https://github.com/open-cluster-management/pipeline/blob/2.1-integration/snapshots/manifest-2020-08-24-20-16-08-2.1.0.json

this is one example of such manifest

we use the image-key to lookup the image information that we try to find

TheRealHaoLiu: lets say an agent uses multiple images and image name may change from release to release but the key will be consistent

i was originally imagine that if the agent upgraded if the name changes we can still use the constant key to map this

addon/v1/types.go
TheRealHaoLiu: i think this project is still using 1.18

addon/v1/types.go
TheRealHaoLiu: changed the name to suggest a strong association and also change the field to required

addon/v1/types.go
TheRealHaoLiu: i changed the name to LatestVersion

addon/v1/types.go
TheRealHaoLiu: added detail description

addon/v1/types.go
TheRealHaoLiu: change the fields to ImageKey and ImagePullSpec

addon/v1/types.go
TheRealHaoLiu: added details

addon/v1/types.go
TheRealHaoLiu: is it ok if it remain a boolean, this is in the spec, it's the same way that `hubAcceptClient` does it

addon/v1/managedclusteraddon_types.go
```diff
@@ -0,0 +1,181 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	runtime "k8s.io/apimachinery/pkg/runtime"
+)
+
+// +genclient
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// ManagedClusterAddOn is the Custom Resource object which holds the current state
+// of an operator. This object is used by operators to convey their state to
+// the rest of the cluster.
```
qiujian16: you may want  to doc which namespace this api resides, it seems to me that it should reside in cluster namespace?

addon/v1/managedclusteraddon_types.go
```diff
@@ -0,0 +1,181 @@
...
+}
+
+// ManagedClusterAddOnSpec is empty for now, but you could imagine holding information like "pause".
+type ManagedClusterAddOnSpec struct {
+	// UpdateApproved represents that user has approved the update of the particular addon for a specific
+	// the managed cluster on the hub.
+	// The default value is false, it can only be set to true when the latestVersion and currentVersion
+	// is on different version.
+	// When the value is set true, the controller/operator that manages the addon will proceed to update the addon.
+	// The value will be set to false by mutating webhook when latestVersion matches currentVersion.
```
qiujian16, line 35: Does it mean it dangles based on the status? The field seems to me a permission from user. I would have a separate status field (maybe conditions) to reflect the matching between latestVersion and currentVersion

addon/v1/managedclusteraddon_types.go
```diff
@@ -0,0 +1,181 @@
...
+	// +required
+	Group string `json:"group"`
+	// resource of the referent.
+	// +kubebuilder:validation:Required
+	// +required
+	Resource string `json:"resource"`
+	// name of the referent.
+	// +kubebuilder:validation:Required
+	// +required
+	Name string `json:"name"`
```
qiujian16, line 82: i think you also need namespace here?

addon/v1/managedclusteraddon_types.go
```diff
@@ -0,0 +1,181 @@
...
+	// CurrentVersion indicates the current Version of the addon
+	// During the agent update process this field will be set to same value as latestVersion after the update has been completed
+	CurrentVersion Release `json:"currentVersion,omitempty"`
+
+	// extension contains any additional status information specific to the
+	// operator which owns this status object.
+	// +nullable
+	// +optional
+	// +kubebuilder:pruning:PreserveUnknownFields
+	Extension runtime.RawExtension `json:"extension"`
```
qiujian16: not quite clear on what the usage of this field is

addon/v1/managedclusteraddon_types.go
```diff
@@ -0,0 +1,181 @@
...
+}
+
+// RelatedImage represents information for one of the images that the agent uses its associated key.
+// +k8s:deepcopy-gen=true
+type RelatedImage struct {
+	// ImageKey is the unique identifier to link the image to specific deployment for the agent
+	// this will be use for backward compatability, i.e if the hub updated but agent have not the agent
+	// controller can use this information to continue to manage the agent.
+	// +required
+	ImageKey string `json:"imageKey"`
```
qiujian16, line 135: should this be optional? since it is for backward compatible.

addon/v1/managedclusteraddon_types.go
TheRealHaoLiu: yes, it should reside in cluster namespace i will add it

addon/v1/managedclusteraddon_types.go
TheRealHaoLiu: like `status.updateAvailable`? would it be boolean?

addon/v1/managedclusteraddon_types.go
TheRealHaoLiu: in the current design this is only for referring to resource in the same namespace

addon/v1/managedclusteraddon_types.go
TheRealHaoLiu: i will remove that

addon/v1/managedclusteraddon_types.go
TheRealHaoLiu: so the situation that im trying to address here is like, if a user deleted the manifestwork that deployed the agent the controller will still have the information needed to recreate. the manifestwork

addon/v1/managedclusteraddon_types.go
qiujian16: to confirm, it is the same namespace of klusterletAddon API?

addon/v1/managedclusteraddon_types.go
```diff
@@ -0,0 +1,181 @@
...
+	LatestVersion Release `json:"latestVersion,omitempty"`
+
+	// CurrentVersion indicates the current Version of the addon
+	// During the agent update process this field will be set to same value as latestVersion after the update has been completed
+	// +optional
+	CurrentVersion Release `json:"currentVersion,omitempty"`
+
+	// UpdateAvailable indicates if there is an version update that is available for the addon
+	// it will be set to false if currentVersion.version != latestVersion.version
+	// it will be set to true if currentVersion.version == latestVersion.version
```
leena-jawale, line 64: Should this be `false if currentVersion.version == latestVersion.version` and `true if currentVersion.version != latestVersion.version`

TheRealHaoLiu:
closing this PR and open a new one https://github.com/open-cluster-management/api/pull/59

TheRealHaoLiu:
closing and switch to https://github.com/open-cluster-management/api/pull/59

PR #58, opened 2020-08-20
Merging API description updates

cadawson:


qiujian16:
@cadawson  I think maybe it needs to run `make verify`  locally,  and push the changes. It seems crd manifests are not updated.

PR #59, opened 2020-08-25
ClusterManagementAddon and ManagedClusterAddOn API

TheRealHaoLiu:


addon/v1alpha1/0000_00_addon.open-cluster-management.io_clustermanageraddons.crd.yaml
```diff
@@ -0,0 +1,70 @@
+apiVersion: apiextensions.k8s.io/v1beta1
```
pmorie: Does the skew of supported versions support v1 CRDs?

addon/v1alpha1/types_clustermanageraddon.go
```diff
@@ -0,0 +1,63 @@
...
+)
+
+// +genclient
+// +genclient:nonNamespaced
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+// +kubebuilder:subresource:status
+// +kubebuilder:resource:scope="Cluster"
+
+// ClusterManagerAddOn represents the registration of an addon
+// It allows the user to discover which addon is available for the cluster manager.
```
pmorie: What operator cr in this API would this align with?

addon/v1alpha1/types_clustermanageraddon.go
```diff
@@ -0,0 +1,63 @@
...
+// It provide a linkage from ClusterManagerAddOn to ManagedClusterAddOn
```
pmorie: I’m not exactly sure what this sentence means because it defines the type in terms of itself

addon/v1alpha1/types_clustermanageraddon.go
```diff
@@ -0,0 +1,63 @@
...
+// The name of the ClusterManagerAddOn resource will be the use for the namespace scoped ManagedClusterAddon resource
```
pmorie: Used for

addon/v1alpha1/types_clustermanageraddon.go
```diff
@@ -0,0 +1,63 @@
...
+	// +optional
+	Status ClusterManagerAddOnStatus `json:"status,omitempty"`
+}
+
+// ClusterManagerAddOnSpec provides the information of addon CustomResourceDefination, related ManagedClusterAddon
+// names, and addon information we will use for UI
+type ClusterManagerAddOnSpec struct {
+	// DisplayName represents the name that will be displayed.
+	// +required
+	DisplayName string `json:"displayName"`
```
pmorie: There’s a conventional annotation that UIs use for this property

addon/v1alpha1/types_clustermanageraddon.go
```diff
@@ -0,0 +1,63 @@
...
+
+	// Description represents the detailed description of the addon.
+	// +optional
+	Description string `json:"description"`
```
pmorie: Likewise, there’s an annotation that’s conventional in the ecosystem for ui descriptions

addon/v1alpha1/types_clustermanageraddon.go
```diff
@@ -0,0 +1,63 @@
...
+
+	// AddOnConfigCRD is the name for the CRD name of the addon CRs, which is used for addon configurations.
+	// It will inform the user which resource the add-on controller/operator will use to control the addon.
+	// +required
```
pmorie: Do we need this field if it’s name aligned with another cluster scoped resource?

addon/v1alpha1/types_managedclusteraddon.go
```diff
@@ -0,0 +1,181 @@
+package v1alpha1
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// +genclient
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// ManagedClusterAddOn is the Custom Resource object which holds the current state
```
pmorie, line 14: Managed... holds

addon/v1alpha1/types_managedclusteraddon.go
```diff
@@ -0,0 +1,181 @@
...
+// of an addon. This object is used by addon operators to convey their state.
```
pmorie: Of an addon on the managed cluster

addon/v1alpha1/types_managedclusteraddon.go
```diff
@@ -0,0 +1,181 @@
...
+
+	// status holds the information about the state of an operator.  It is consistent with status information across
+	// the Kubernetes ecosystem.
+	// +optional
+	Status ManagedClusterAddOnStatus `json:"status"`
+}
+
+// ManagedClusterAddOnSpec is empty for now, but you could imagine holding information like "pause".
+type ManagedClusterAddOnSpec struct {
+	// UpdateApproved represents that user has approved the update of the particular addon for a specific
```
pmorie: The spec is not empty :)

addon/v1alpha1/types_managedclusteraddon.go
```diff
@@ -0,0 +1,181 @@
...
+// ManagedClusterAddOnStatus provides information about the status of the operator.
+// +k8s:deepcopy-gen=true
+type ManagedClusterAddOnStatus struct {
+	// conditions describe the state of the operator's managed and monitored components.
+	// +patchMergeKey=type
+	// +patchStrategy=merge
+	// +optional
+	Conditions []AddOnStatusCondition `json:"conditions,omitempty"  patchStrategy:"merge" patchMergeKey:"type"`
+
+	// addonResource is a reference  to an object that contain the configuration of the addon
```
pmorie: In what scope?

pmorie (commented):
Some initial comments

pmorie (commented):
Some initial comments

addon/v1alpha1/0000_00_addon.open-cluster-management.io_clustermanageraddons.crd.yaml
TheRealHaoLiu: it does not we do still support 4.3

addon/v1alpha1/types_clustermanageraddon.go
TheRealHaoLiu: can you point to an example for this

addon/v1alpha1/0000_00_addon.open-cluster-management.io_clustermanageraddons.crd.yaml
pmorie: ack

addon/v1alpha1/types_clustermanageraddon.go
TheRealHaoLiu: the desire to put this in the spec itself rather than the in the annotation is beause we figure it would be more CLI friendly where we can print the information with server print

addon/v1alpha1/types_clustermanageraddon.go
pmorie: ...asking because I think this type needs to be contextualized with other types in the API here. It's not clear exactly what "the cluster manager" is specifically in this context.

addon/v1alpha1/types_clustermanageraddon.go
TheRealHaoLiu: i will update the description, i can see why its misleading

TheRealHaoLiu: the motivation of this field is that one addon config CRD may control multiple addons 
so that for each addon the addon config CRD controls it will create an managedclusteraddon

the relationship is not 1:1

addon/v1alpha1/types_managedclusteraddon.go
TheRealHaoLiu: updating the spec to remove all of the update related fields

addon/v1alpha1/types_managedclusteraddon.go
TheRealHaoLiu: will add to description to clarify

addon/v1alpha1/types_clustermanageraddon.go
```diff
@@ -0,0 +1,63 @@
...
+// +genclient:nonNamespaced
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+// +kubebuilder:subresource:status
+// +kubebuilder:resource:scope="Cluster"
+
+// ClusterManagementAddOn represents the registration of an add-on to the cluster manager.
+// This resource allows the user to discover which add-on is available for the cluster manager and
+// also provides metadata information about the add-on.
+// The resource also provides a linkage to ManagedClusterAddOn, the name of the ClusterManagementAddOn
+// resource will be use for the namespace scoped ManagedClusterAddOn resource
```
swopebe: will be used (line 17)

addon/v1alpha1/types_clustermanageraddon.go
```diff
@@ -0,0 +1,63 @@
...
+// ClusterManagementAddOn is a cluster scoped resource.
+type ClusterManagementAddOn struct {
```
swopebe: cluster-scoped

addon/v1alpha1/types_clustermanageraddon.go
```diff
@@ -0,0 +1,63 @@
...
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	// Spec represents a desired configuration for the agent on the cluster manager add-on
+	Spec ClusterManagementAddOnSpec `json:"spec"`
```
swopebe: period at the end

addon/v1alpha1/types_clustermanageraddon.go
```diff
@@ -0,0 +1,63 @@
...
+
+	// Status represents the current status of cluster manager add-on
+	// +optional
+	Status ClusterManagementAddOnStatus `json:"status,omitempty"`
+}
+
+// ClusterManagementAddOnSpec provides the information of add-on CustomResourceDefinition.
+type ClusterManagementAddOnSpec struct {
+	// DisplayName represents the name that will be displayed.
+	// +required
```
swopebe: should required be Required? line 34

addon/v1alpha1/types_clustermanageraddon.go
```diff
@@ -0,0 +1,63 @@
...
+	DisplayName string `json:"displayName"`
+
+	// Description represents the detailed description of the add-on.
+	// +optional
+	Description string `json:"description"`
+
+	// AddOnConfigCRD is a reference to the name of the CRD that driving the configuration of the add-on
```
swopebe: that configures?

addon/v1alpha1/types_clustermanageraddon.go
```diff
@@ -0,0 +1,63 @@
...
+	// Note: there may be a case where a single add-on config CRD controls multiple related add-ons,
```
swopebe: There (capital)

addon/v1alpha1/types_managedclusteraddon.go
```diff
@@ -0,0 +1,131 @@
...
+	// +required
+	AddOnResource ObjectReference `json:"addOnResource"`
+}
+
+// ObjectReference contains enough information to let you inspect or modify the referred object.
+type ObjectReference struct {
+	// group of the referent.
+	// +kubebuilder:validation:Required
+	// +required
+	Group string `json:"group"`
```
qiujian16, line 69: Version also?

addon/v1alpha1/types_clustermanagementaddon.go
```diff
@@ -0,0 +1,63 @@
...
+	// DisplayName represents the name that will be displayed.
+	// +required
+	DisplayName string `json:"displayName"`
+
+	// Description represents the detailed description of the add-on.
+	// +optional
+	Description string `json:"description"`
+
+	// AddOnConfigCRD is a reference to the name of the CRD that configures the add-on
+	// Note: There may be a case where a single add-on config CRD controls multiple related add-ons,
```
qiujian16: do you meant there will be multiple clustermangeraddon resource that specify the same crd? Would you clarify the scenario?

addon/v1alpha1/types_managedclusteraddon.go
TheRealHaoLiu: updated

addon/v1alpha1/types_clustermanagementaddon.go
```diff
@@ -0,0 +1,63 @@
...
+type ClusterManagementAddOnSpec struct {
+	// DisplayName represents the name that will be displayed.
+	// +required
+	DisplayName string `json:"displayName"`
+
+	// Description represents the detailed description of the add-on.
+	// +optional
+	Description string `json:"description"`
+
+	// AddOnConfigCRD is a reference to the name of the CRD that configures the add-on
```
swopebe: punctuation

addon/v1alpha1/types_clustermanagementaddon.go
```diff
@@ -0,0 +1,63 @@
...
+	// Note: There may be a case where a single add-on config CRD controls multiple related add-ons,
+	// in this case multiple ClusterManagementAddOn resource should be created.
+	// +required
```
swopebe: Split these into 2 full sentences with punctuation. (line 42 and line 43)

addon/v1alpha1/types_managedclusteraddon.go
```diff
@@ -0,0 +1,131 @@
...
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// ManagedClusterAddOn is the Custom Resource object which holds the current state
+// of an add-on. This object is used by add-on operators to convey their state.
+// The resource should be created in the ManagedCluster's cluster namespace.
+type ManagedClusterAddOn struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata"`
+
+	// spec holds configuration that could apply to any operator.
```
swopebe, line 21: capital S

addon/v1alpha1/types_managedclusteraddon.go
```diff
@@ -0,0 +1,131 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// +genclient
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// ManagedClusterAddOn is the Custom Resource object which holds the current state
+// of an add-on. This object is used by add-on operators to convey their state.
+// The resource should be created in the ManagedCluster's cluster namespace.
+type ManagedClusterAddOn struct {
```
swopebe, line 17: Try to stay away from possessive and rewrite:

Create the resource in the namespace for the ManagedCluster.

addon/v1alpha1/types_managedclusteraddon.go
```diff
@@ -0,0 +1,131 @@
...
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata"`
+
+	// spec holds configuration that could apply to any operator.
+	// +kubebuilder:validation:Required
+	// +required
+	Spec ManagedClusterAddOnSpec `json:"spec"`
+
+	// status holds the information about the state of an operator.  It is consistent with status information across
```
swopebe, line 26: Begin sentence with cap.

addon/v1alpha1/types_managedclusteraddon.go
```diff
@@ -0,0 +1,131 @@
...
+	// the Kubernetes ecosystem.
+	// +optional
+	Status ManagedClusterAddOnStatus `json:"status"`
+}
+
+// ManagedClusterAddOnSpec is empty for now, but you could imagine holding information like "pause".
+type ManagedClusterAddOnSpec struct {
+}
```
swopebe, line 34: Let's not tell user to imagine...

.... is currently empty, but can contain information such as "pause" (Is pause a status the user is looking for, if so let's write something like "pause" status?

addon/v1alpha1/types_managedclusteraddon.go
```diff
@@ -0,0 +1,131 @@
...
+
+// ManagedClusterAddOnStatus provides information about the status of the operator.
+// +k8s:deepcopy-gen=true
+type ManagedClusterAddOnStatus struct {
+	// conditions describe the state of the operator's managed and monitored components.
+	// +patchMergeKey=type
```
swopebe, line 40: Throughout, be sure to start sentences with caps. Also, always avoid possessives.

.... managed and monitored components for the operator.

addon/v1alpha1/types_managedclusteraddon.go
```diff
@@ -0,0 +1,131 @@
...
+	// +patchStrategy=merge
+	// +optional
+	Conditions []AddOnStatusCondition `json:"conditions,omitempty"  patchStrategy:"merge" patchMergeKey:"type"`
+
+	// addOnResource is a reference to the detailed resource driving the add-on
+	// this resource must be located in the same namespace as the ManagedClusterAddOn
```
swopebe: Caps, try not to use "driving" as it can mean many things to people who natively speak other languages.

addon/v1alpha1/types_managedclusteraddon.go
```diff
@@ -0,0 +1,131 @@
...
+	Resource string `json:"resource"`
+	// name of the referent.
+	// +kubebuilder:validation:Required
+	// +required
+	Name string `json:"name"`
+}
+
+// AddOnStatusCondition represents the state of the add-on's
+// managed and monitored components.
+// +k8s:deepcopy-gen=true
```
swopebe: try to rewrite to avoid possessives

addon/v1alpha1/types_managedclusteraddon.go
```diff
@@ -0,0 +1,131 @@
...
+type AddOnStatusCondition struct {
+	// type specifies the aspect reported by this condition.
```
swopebe: Caps throughout

addon/v1alpha1/types_managedclusteraddon.go
```diff
@@ -0,0 +1,131 @@
...
+	// +kubebuilder:validation:Required
+	// +required
+	Type AddOnStatusConditionType `json:"type"`
+
+	// status of the condition, one of True, False, Unknown.
+	// +kubebuilder:validation:Required
```
swopebe: Status of the condition. Status can be True, False, Unknown.

addon/v1alpha1/types_managedclusteraddon.go
```diff
@@ -0,0 +1,131 @@
...
+	// may have a lower quality of service.  A component may be Progressing but
+	// not Degraded because the transition from one state to another does not
+	// persist over a long enough period to report Degraded.  A service should not
+	// report Degraded during the course of a normal upgrade.  A service may report
+	// Degraded in response to a persistent infrastructure failure that requires
+	// administrator intervention.  For example, if a control plane host is unhealthy
+	// and must be replaced.  An operator should report Degraded if unexpected
+	// errors occur over a period, but the expectation is that all unexpected errors
+	// are handled as operators mature.
+	Degraded AddOnStatusConditionType = "Degraded"
```
swopebe: Break this out a bit and highlight with code or quotes the values of that status. Also, please consider a definition list:

1. Available: Indicates......
(Make a list like this -- easier to see definitions.)
2. But a `Degraded` state...

(Within a paragraph, set it apart with code blocks.)

swopebe (commented):
I have made some suggestions and highlighted some changes that are necessary, please see comments.

addon/v1alpha1/types_clustermanagementaddon.go
```diff
@@ -0,0 +1,65 @@
...
+
+// +genclient
+// +genclient:nonNamespaced
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+// +kubebuilder:subresource:status
+// +kubebuilder:resource:scope="Cluster"
+// +kubebuilder:printcolumn:name="DISPLAY NAME",type=string,JSONPath=`.spec.displayName`
+// +kubebuilder:printcolumn:name="CRD NAME",type=string,JSONPath=`.spec.addOnConfigCRD`
+
+// ClusterManagementAddOn represents the registration of an add-on to the cluster management.
```
pmorie: What operator CRD does this API have behavior with respect to?

addon/v1alpha1/types_managedclusteraddon.go
TheRealHaoLiu: this is referring to the `json:"status"`

addon/v1alpha1/types_managedclusteraddon.go
TheRealHaoLiu: this is refering to the `json:"spec"`

addon/v1alpha1/types_managedclusteraddon.go
TheRealHaoLiu: will remove `, but you could imagine holding information like "pause".`

addon/v1alpha1/types_managedclusteraddon.go
TheRealHaoLiu: this section is copy and pasted from https://github.com/openshift/api/blob/master/config/v1/types_cluster_operator.go#L138-L175

addon/v1alpha1/types_clustermanagementaddon.go
TheRealHaoLiu: can u clarify this comment

addon/v1alpha1/types_clustermanagementaddon.go
TheRealHaoLiu: you are correct, multiple addon may share the same configuration resource and in this case the controller will create multiple clustermanageraddon

addon/v1alpha1/types_clustermanagementaddon.go
TheRealHaoLiu: added

addon/v1alpha1/types_clustermanagementaddon.go
TheRealHaoLiu: added

addon/v1alpha1/types_managedclusteraddon.go
TheRealHaoLiu: fixed

addon/v1alpha1/types_managedclusteraddon.go
TheRealHaoLiu: would it be ok if we only use the latest version supplied?

addon/v1alpha1/types_managedclusteraddon.go
TheRealHaoLiu: type is referring to `json:"type"`, observing the conventions in other repos in openshift

addon/v1alpha1/types_managedclusteraddon.go
TheRealHaoLiu: updated

addon/v1alpha1/types_clustermanagementaddon.go
TheRealHaoLiu: updated the comment to further clarify the situation

addon/v1alpha1/types_managedclusteraddon.go
TheRealHaoLiu: fixed

addon/v1alpha1/types_managedclusteraddon.go
TheRealHaoLiu: the operator will internally know which version of the resource it should use

addon/v1alpha1/types_clustermanagementaddon.go
```diff
@@ -0,0 +1,64 @@
...
+	// displayName represents the name that will be displayed.
+	// +required
+	DisplayName string `json:"displayName"`
+
+	// description represents the detailed description of the add-on.
+	// +optional
+	Description string `json:"description"`
+
+	// addOnConfigCRD is a reference to the name of the CRD that configures the add-on.
+	// In scenario where a multiple add-ons share the same add-on config CRD multiple ClusterManagementAddOn resource need to be created.
```
deads2k: be clear that they can all reference the same CRD

addon/v1alpha1/types_clustermanagementaddon.go
```diff
@@ -0,0 +1,64 @@
...
+	// +required
+	AddOnConfigCRD string `json:"addOnConfigCRD"`
```
deads2k: A clever spec solution. I like it.  Since this represents a single add on, should there be an option to specify a name for the CR instance?  I can understand it being optional so you could control many, but it seems useful.

addon/v1alpha1/types_clustermanagementaddon.go
deads2k: provide guidance.  Do you see this as cluster scoped or namespace scoped?

addon/v1alpha1/types_managedclusteraddon.go
```diff
@@ -0,0 +1,134 @@
...
+
+	// status holds the information about the state of an operator.  It is consistent with status information across
+	// the Kubernetes ecosystem.
+	// +optional
+	Status ManagedClusterAddOnStatus `json:"status"`
+}
+
+// ManagedClusterAddOnSpec is empty for now.
+type ManagedClusterAddOnSpec struct {
+}
```
deads2k, line 34: I expected to see the CRD reference information and the display name and description also listed here.  The idea being that they fields are managed at the cluster scope, but this resource provides one-stop-shopping for "what does this mean".

addon/v1alpha1/types_managedclusteraddon.go
```diff
@@ -0,0 +1,134 @@
...
+	// conditions describe the state of the managed and monitored components for the operator.
+	// +patchMergeKey=type
+	// +patchStrategy=merge
+	// +optional
+	Conditions []AddOnStatusCondition `json:"conditions,omitempty"  patchStrategy:"merge" patchMergeKey:"type"`
+
+	// addOnResource is a reference to the detailed resource that configures the add-on.
+	// This resource must be located in the same namespace as the ManagedClusterAddOn.
+	// +required
+	AddOnResource ObjectReference `json:"addOnResource"`
```
deads2k: using `relatedObjects` like https://github.com/openshift/api/blob/master/config/v1/types_cluster_operator.go#L54 will make you automatically collect some data using must-gather or `oc adm inspect` on this resource

addon/v1alpha1/types_managedclusteraddon.go
```diff
@@ -0,0 +1,134 @@
...
+	Status ManagedClusterAddOnStatus `json:"status"`
+}
+
+// ManagedClusterAddOnSpec is empty for now.
+type ManagedClusterAddOnSpec struct {
+}
+
+// ManagedClusterAddOnStatus provides information about the status of the operator.
+// +k8s:deepcopy-gen=true
+type ManagedClusterAddOnStatus struct {
```
deads2k, line 38: I thought you had version reporting in here. I recommend having a place to report versions.  Something like https://github.com/openshift/api/blob/master/config/v1/types_cluster_operator.go#L47 is pretty unopinionated.

addon/v1alpha1/types_managedclusteraddon.go
deads2k: > the operator will internally know which version of the resource it should use

Yeah, all versions represent the same data.  Consumer can choose.  This schema matches https://github.com/openshift/api/blob/master/config/v1/types_cluster_operator.go#L79 to interact with `must-gather` and `oc adm inspect`.  You may want to indicate that in the type doc

addon/v1alpha1/types_managedclusteraddon.go
```diff
@@ -0,0 +1,134 @@
...
+	// conditions describe the state of the managed and monitored components for the operator.
+	// +patchMergeKey=type
+	// +patchStrategy=merge
+	// +optional
+	Conditions []AddOnStatusCondition `json:"conditions,omitempty"  patchStrategy:"merge" patchMergeKey:"type"`
```
deads2k: can you use `metav1.Condition` from kube 1.19?

addon/v1alpha1/types_managedclusteraddon.go
TheRealHaoLiu: nice, will change to relatedObjects

addon/v1alpha1/types_managedclusteraddon.go
TheRealHaoLiu: the version information has been removed from the first go around

addon/v1alpha1/types_managedclusteraddon.go
TheRealHaoLiu: currently the api project is not in kube 1.19 yet, I think @qiujian16 is working on this in a separate PR. will update the spec once that's in

addon/v1alpha1/types_clustermanagementaddon.go
TheRealHaoLiu: the current plan is to not impose a naming convention for the name of the config CR here and having "relatedResources" in the ManagedCluserAddOn to link to the specific instance of the config CR

addon/v1alpha1/types_managedclusteraddon.go
TheRealHaoLiu: this information exists in the ClusterManagementAddon resource which have a tight relationship with the ManageClusterAddon resource, do we need to duplicate the information here? or can we link to the ClusterManagementAddon resource?

addon/v1alpha1/types_clustermanagementaddon.go
TheRealHaoLiu: ClusterManagementAddOn is cluster scoped and the CRD will also be cluster scoped

addon/v1alpha1/types_clustermanagementaddon.go
TheRealHaoLiu: in the current design the `relatedObjects` will be used to ManagedClusterAddOn CR to the specific addon config resource, would this be sufficient?

addon/v1alpha1/types_managedclusteraddon.go
TheRealHaoLiu: the information is captured in the ClusterManagementAddon would it be sufficient to link out to that resource or should we duplicate the information here

addon/v1alpha1/types_managedclusteraddon.go
TheRealHaoLiu: talking to @showeimer i think we may need the `Version` field, in order to expose to our user which exact resource to update from UI 

also I think if any controller want to use dynamic clientset with this reference we may also need the version as well

deads2k:
don't forget to squash

addon/v1alpha1/types_managedclusteraddon.go
mdelder: The `ManagedClusterAddOn` is a namespace-scoped object that will be created per cluster. I would expect that it has some awareness of its cluster-scoped ClusterManagementAddOn component (where the CRD) reference is defined, but I would expect here we would have a reference to the CR instance of the actual CRD. 

e.g. 

`ClusterManagementAddOn` --> Here's a new cluster management subsystem (cluster scoped) & I look for CRDs named "CRDNAME".
`ManagedClusterAddOn` --> I'm part of cluster management subsystem CLUSTERMGMTADDON_NAME and in particular, I'm here to let you know that you can configure my ClusterManagementAddOn for CLUSTER_NAME with this instance: CRDNAME_INSTANCE (namespace scoped).

addon/v1alpha1/types_managedclusteraddon.go
mdelder: We will need to have support for this API going back to Kube 1.17 at least.

TheRealHaoLiu: @mdelder this should not cause backward compatibility issue

addon/v1alpha1/types_managedclusteraddon.go
TheRealHaoLiu: the linkage between ManagedClusterAddOn and ClusterManagmentAddOn is by name the 2 resource will have the same name

TheRealHaoLiu: the CR instance of the actual CRD will be referenced in the relatedObjects field in the status as we dont expect this to be a user configured parameter

leena-jawale (approved):

addon/v1alpha1/types_clustermanagementaddon.go
```diff
@@ -0,0 +1,76 @@
...
+	// addOnConfiguration is a reference to the name of the CRD and CR that configures the add-on.
+	// In scenario where a multiple add-ons share the same add-on CRD,
+	// multiple ClusterManagementAddOn resources need to be created and reference the same AddOnConfiguration.
+	// +optional
+	AddOnConfiguration ConfigCoordinates `json:"addOnConfiguration"`
+}
+
+// ConfigCoordinates represents the CRD and CR that configures the add-on
+type ConfigCoordinates struct {
+	// crdName represnt name of add-on configuration CRD
```
deads2k: crdName is the name of the CRD used to configure instances of the managed add-on.

addon/v1alpha1/types_clustermanagementaddon.go
```diff
@@ -0,0 +1,76 @@
...
+	// +optional
+	CRDName string `json:"crdName"`
```
deads2k, line 63: You should see if your consumers would prefer to see a separate group and resource.  I'm ok with either choice since you can split and join whichever way you want.

addon/v1alpha1/types_managedclusteraddon.go
```diff
@@ -0,0 +1,84 @@
...
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// +genclient
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+// +kubebuilder:subresource:status
+// +kubebuilder:printcolumn:name="Progressing",type=string,JSONPath=`.status.conditions[?(@.type=="Progressing")].status`
+// +kubebuilder:printcolumn:name="Available",type=string,JSONPath=`.status.conditions[?(@.type=="Available")].status`
```
deads2k, line 11: no degraded?

addon/v1alpha1/types_managedclusteraddon.go
```diff
@@ -0,0 +1,84 @@
...
+	// 1. the detailed resource driving the operator
+	// 2. operator namespaces
+	// 3. operand namespaces
+	// 4. related ClusterManagementAddon resource
+	// +optional
+	RelatedObjects []ObjectReference `json:"relatedObjects,omitempty"`
+
+	// displayName represents the name of add-on that will be displayed.
+	// +optional
+	DisplayName string `json:"displayName"`
```
deads2k: I would group these in a metadata struct.

Also, I don't see an instance of the `AddOnConfiguration ConfigCoordinates` here.

addon/v1alpha1/types_managedclusteraddon.go
TheRealHaoLiu: fixed

addon/v1alpha1/types_managedclusteraddon.go
TheRealHaoLiu: fixed

addon/v1alpha1/types_clustermanagementaddon.go
TheRealHaoLiu: updated

addon/v1alpha1/types_clustermanagementaddon.go
```diff
@@ -0,0 +1,86 @@
...
+	// status represents the current status of cluster management add-on.
+	// +optional
+	Status ClusterManagementAddOnStatus `json:"status,omitempty"`
+}
+
+// ClusterManagementAddOnSpec provides information for the add-on.
+type ClusterManagementAddOnSpec struct {
+	// addOnMeta is a reference to the metadata information for the add-on.
+	// +optional
+	AddOnMeta AddOnMeta `json:"addOnMeta"`
```
mdelder, line 38: Do we think that this kind of metadata is going to grow in the future? Would it be easier to just list the title/description/etc at the root of the `spec`?

addon/v1alpha1/types_managedclusteraddon.go
mdelder: If a user creates a `ManagedClusterAddOn` and a `ClusterManagmentAddOn` of the same name isn't found, will a validating webhook reject the creation? Having them associated by name alone may be hard for users to predict. By convention, using the same name is fine -- but as an artifact to build a relationship in a model, it could be errorproned.

mdelder (approved):
Overall I think the concepts are good.

We should reconsider the association by instance name between the addon and mgmt addon.

PR #60, opened 2020-08-25
Adopt metav1 condition

qiujian16:
·


cluster/v1/types.go
```diff
@@ -84,7 +84,7 @@ type ClientConfig struct {
 // ManagedClusterStatus represents the current status of joined managed cluster.
 type ManagedClusterStatus struct {
 	// Conditions contains the different condition statuses for this managed cluster.
-	Conditions []StatusCondition `json:"conditions"`
+	Conditions []metav1.Condition `json:"conditions"`
```
deads2k, line 87: be sure your current data is compatible with this change

deads2k:
Be sure any existing data you write is compatible with this change, but I'm in favor and I think it will make your life easier in the future.

/approve

cluster/v1/types.go
qiujian16: yes, i am quite sure it is compatible.

PR #62, opened 2020-09-04
Make ManagedClusterSet exclusive

elgnay:
We have a proposal about how to make ManagedClusterSet exclusive: https://docs.google.com/document/d/1Dl-8g6pJS6wA-MXiaBijDjwXLPX7cJ-HQCoQKIRDze0/edit#, which was discussed in ACM Architecture Topics Open Forum on Sept. 3rd. And it's agreed that option 3 is a better solution. That's the reason why I made this change.


cluster/v1/types.go
```diff
@@ -66,6 +66,18 @@ type ManagedClusterSpec struct {
 	// If its value is zero, the Klusterlet agent will update its lease every 60s by default
 	// +optional
 	LeaseDurationSeconds int32 `json:"leaseDurationSeconds,omitempty"`
+
+	// ClusterSet is the name of the ManagedClusterSet to which the ManagedCluster belongs to.
+	// If the value is empty, the ManagedCluster does not belong to any ManagedClusterSet.
+	// The following conditions must be met in order to set the value to a certain ManagedClusterSet:
+	//   1. This ManagedClusterSet must exist;
+	//   2. User must have a RBAC rule to UPDATE on this ManagedClusterSet;
+	//   3. If it already refers to a ManagedClusterSet, user must have a RBAC rule to UPDATE on that
```
deads2k: update on both the old and new managedclusterset

cluster/v1/types.go
```diff
@@ -66,6 +66,18 @@ type ManagedClusterSpec struct {
...
+	//      ManagedClusterSet either.
+	// If the value is set, deletion of the ManagedClusterSet will also trigger deletion of the containing
```
deads2k: We normally allow dangling references so that we have eventual consistency and don't have to handle referential integrity problems that result in user resource deletion.

deads2k:
I think this creating some tradeoffs that should be explicitly described in the doc and I'd like to see that in an MD file in this repo because they aren't all obvious.

1. permissions to join a managedclusterset.  If it is a separate resource, then the permissions model of "who can add a managed cluster to the 'confidential clusters'" group is easy.  As a field on the managedcluster, it is hard
2. lifecycle of a managedcluster and managedclusterset.  The two resources are distinct and eventually consistent.  there is no direct reference.  Using a field on the managecluster ensures only one managedclusterset per managedcluster, but creates referential integrity problems.
3. grouping clusters by different criteria.  Some choices of grouping are orthogonal to each other.  Particularly in a shared hub model (which many choices are leading towards).  For example, if you have DMZ and non-DMZ clusters and you have production and stage clusters.  Then each of the axes are valuable to slice on.

I'll propose one alternative.  If a customer wishes to have exclusive sets of managedclusters, could we instead recommend the creation of a specific label `clusters.open-cluster-management.io/managed-cluster-set={setOne,setTwo,setThree}`.  The ManagedClusterSet could even have a structured field like "specificallyNamedCluster" to help prevent typos.  This keeps the mechanism as label selection, but does not preclude subselection.

deads2k:
Fleshing out the idea further, imagine adding this to the `ClusterSelector`

```go
// ClusterSelector has three fields, if more than one field is set, then they are logically AND'd.
type ClusterSelector struct{
  // stuff
  // explicitClusterName is equivalent to setting `clusters.open-cluster-management.io/managed-cluster-set=explicitClusterName`.  This field gives an explicit intent that  is useful in cases where disjoint sets of managedclusters are desired.
  // If you want disjoint sets of managedclusters, you should also enable the OnlyDisjointMangedClusters validating admission webhook.
  ExplicitClusterName string
}
```

The OnlyDisjointMangedClusters admission webhook ensures that 
1. A ManagedClusterSet can have only one ClusterSelector
2. The ClusterSelector may only specify the ExplicitClusterName
The combination of these two qualities means that every ManagedCluster can only be included in a single ManagedClusterSet (there's only one value for the `clusters.open-cluster-management.io/managed-cluster-set` label per ManagedCluster).  It is possible to create two ManagedClusterSets that specify exactly the same set of managedclusters, but that is less an understandability gap and more of an alias.  The OnlyDisjointMangedClusters admission webhook could also do a best-effort at eliminating duplicate `ExplicitClusterName` values, but it's not a security boundary problem.

To enable this feature, the operator could have a flag for `OnlyDisjointMangedClusters` that enables the admission webhook *if* the existing ManagedClusterSets are compatible with enabling it.

cluster/v1/types.go
pmorie: This is the one property I am still a bit iffy about. I guess we at least are starting with the option that is easier to relax. It seems like it would be easier to begin allowing dangling references than to begin forbidding them. Let's pay careful attention to any feedback we get from users about this property.

cluster/v1alpha1/types.go
```diff
@@ -12,47 +12,28 @@ import (
 
 // ManagedClusterSet defines a group of ManagedClusters that user's workload can run on.
 // A workload can be defined to deployed on a ManagedClusterSet, which mean:
```
pmorie, line 14: I'd prefer to have more specific language here that refers to other parts of the API. Not blocking feedback for this PR.

pmorie: I'm also asking myself how much of this content is really about subscription. Again, not blocking feedback.

pmorie:
I do not think I understand what point labeling the `ManagedCluster` serves - is it convenience only? Is it to provide a second coordinate for the admission plugin when it enforces disjoint sets?

deads2k:
> I do not think I understand what point labeling the `ManagedCluster` serves - is it convenience only? Is it to provide a second coordinate for the admission plugin when it enforces disjoint sets?

Label selection is the "normal" way to select resources matching a pattern.  pods, namespaces, endpoints, RBAC roles, HPA, deployments, replicasets: all labels.

By selecting a particular label we can use an equality test to ensure disjoint groups are guaranteed while simultaneously having a single API for selection, having clear permission boundaries, and avoiding dangling references.

pmorie:
> I do not think I understand what point labeling the ManagedCluster serves - is it convenience only? Is it to provide a second coordinate for the admission plugin when it enforces disjoint sets?

I'm sorry, I had meant with the special label `clusters.open-cluster-management.io/managed-cluster-set=explicitClusterName`

elgnay:
@deads2k Here is my understanding of the alternative solution you proposed. Please correct me if I'm wrong.
### API changes

1. Update  `ClusterSelector` struct and add a new field `ExplicitClusterName`.
    ```
    type ClusterSelector struct {
    	// ClusterNames represents a list of cluster name
    	// +optional
    	ClusterNames []string `json:"clusterNames,omitempty"`

    	// LabelSelector represents a label selector to select cluster by label
    	// +optional
    	LabelSelector *metav1.LabelSelector `json:"labelSelector,omitempty"`

    	// explicitClusterName is equivalent to setting a label selector with
    	// `clusters.open-cluster-management.io/managed-cluster-set=explicitClusterName`.
    	// This field gives an explicit intent that is useful in cases where disjoint sets
    	// of managedclusters are desired.
    	// If you want disjoint sets of managedclusters, you should also enable the
    	// OnlyDisjointMangedClusterSets validating admission webhook.
    	// +optional
    	ExplicitClusterName string `json:"explicitClusterName,omitempty"`
    }
    ```

2. Add a field `OnlyDisjointManagedClusterSets` in `ClusterManagerSpec` struct.
    ```
    type ClusterManagerSpec struct {
    	// RegistrationImagePullSpec represents the desired image of registration controller/webhook installed on hub.
    	// +required
    	RegistrationImagePullSpec string `json:"registrationImagePullSpec"`

    	// WorkImagePullSpec represents the desired image configuration of work controller/webhook installed on hub.
    	// +required
    	WorkImagePullSpec string `json:"workImagePullSpec,omitempty"`

    	// OnlyDisjointManagedClusterSets represents whether disjoint ManagedClusterSets is mandatory.
    	// +optional
    	OnlyDisjointManagedClusterSets bool `json:"onlyDisjointManagedClusterSets,omitempty"`
    }
    ```


    If `OnlyDisjointManagedClusterSets` is set to ture, the following validation will be done by admission webhook when creating/updating ManagedClusterSets.

    * A ManagedClusterSet can have only one ClusterSelector
    * The ClusterSelector may only specify the ExplicitClusterName

### Usage

User is able to enable disjoint ManagedClusterSets feature with a ClusterManager CR as below.
```
apiVersion: operator.open-cluster-management.io/v1
kind: ClusterManager
metadata:
  name: cluster-manager
spec:
  registrationImagePullSpec: quay.io/open-cluster-management/registration
  workImagePullSpec: quay.io/open-cluster-management/work
  onlyDisjointManagedClusterSets: true
```

And a ManagedClusterSet looks like,

```
apiVersion: cluster.open-cluster-management.io/v1alpha1
kind: ManagedClusterSet
metadata:
  name: cs-prod
spec:
  clusterSelectors:
  - ExplicitClusterName: prod
```

It includes all ManagedClusters which have a label `clusters.open-cluster-management.io/managed-cluster-set` with value `prod`. An example looks like,

```
apiVersion: cluster.open-cluster-management.io/v1
kind: ManagedCluster
metadata:
  name: cluster1
  labels:
    clusters.open-cluster-management.io/managed-cluster-set: prod
spec:
  hubAcceptsClient: true

```

Any two ManagedClusterSets are either disjoint or exactly the same.

qiujian16:
> ```go
> // ClusterSelector has three fields, if more than one field is set, then they are logically AND'd.
> type ClusterSelector struct{
>   // stuff
>   // explicitClusterName is equivalent to setting `clusters.open-cluster-management.io/managed-cluster-set=explicitClusterName`.  This field gives an explicit intent that  is useful in cases where disjoint sets of managedclusters are desired.
>   // If you want disjoint sets of managedclusters, you should also enable the OnlyDisjointMangedClusters validating admission webhook.
>   ExplicitClusterName string
> }
> ```

I think `ExplicitClusterName` does not clearly represent the meaning of this field. How about `explicitClusterSetLabelValue`?

cluster/v1alpha1/0000_00_clusters.open-cluster-management.io_managedclustersets.crd.yaml
```diff
@@ -1,6 +1,8 @@
 apiVersion: apiextensions.k8s.io/v1beta1
 kind: CustomResourceDefinition
 metadata:
+  annotations:
+    controller-gen.kubebuilder.io/version: (devel)
```
qiujian16: should we remove this annotatino?

docs/clusterset.md
```diff
@@ -0,0 +1,18 @@
...
+## Lifecycle of ManagedClusters and ManagedClusterSets
+
+The two resources are distinct and eventually consistent. There is no direct reference between them. Deletion of ManagedClusterSets will not result in deletion of ManagedClusters.
+
+## Grouping ManagedClusters by different criteria. 
+
+With ManagedClusterSets, ManagedClusters can be grouped by different criteria. Some choices of grouping are orthogonal to each other. That means a ManagedCluster may belong to more than one ManagedClusterSets. There also are some user scenarios which requires all ManagedClusterSets are disjoint. In order to support both of them, ManagedClusterSets can be used in two modes:
+
+* Normal. ManagedClusterSets can overlap with each others
+* Disjoint. All ManagedClusterSets are disjoint.
```
qiujian16: I think we need to call out that we would need a validation webhook to ensure disjointness and how.

cluster/v1alpha1/0000_00_clusters.open-cluster-management.io_managedclustersets.crd.yaml
elgnay: Removed

docs/clusterset.md
elgnay: Added.

mdelder:
The primary use case that is motivating the introduction of ManagedClusterSet is around access control. We have other mechanisms that use Cluster Label Selection to define a slice of clusters (apps.open-cluster-management.io/PlacementRules). The model today allows a `cluster-admin` or `open-cluster-management:cluster-manager-admin` visibility to all `ManagedClusters`. Separately, users need:

1. `ClusterRoleBinding` to the `ClusterRole` generated for the `ManagedCluster` 
2. `RoleBinding` to the `Namespace` associated with the `ManagedCluster` (primarily for list as well as delivering content to the `ManagedCluster`).

The `PlacementRule` evaluates matches based on the permissions of the user that creates the `PlacementRule`. One of the use cases for `ManagedClusterSet` is to establish an access control boundary, much like a `Namespace` does today. 

If the `ManagedClusterSet` uses `LabelSelector` for set-inclusion, then the model still presents the same problem -- how is the set of available `ManagedClusters` determined for evaluation by the `LabelSelector` conditions? Either `ManagedClusterSets` can only be created by super users or the `LabelSelector` is limited by the current RBAC for the user that creates the `ManagedClusterSet`. So, I don' think we adequately address the goal of establishing an access control bounudary.

If the model applies a reference from the `ManagedCluster` to the `ManagedClusterSet`, then a `ManagedCluster` may optionally be part of a set (maintaining API compatibility for existing ManagedClusters). If the `ManagedClusterSet` is deleted, then all member `ManagedClusters` will also be destroyed. The semantics are consistent with Namespace-scoped resources. Child resources reference the parent. The role that allows the creation of a new `ManagedClusterSet` does not require any broad visibility to all of the existing `ManagedClusters`. So a user could:

1. User `skyler` with `ClusterRole` `open-cluster-management:managedclusterset-create` creates a new `ManagedClusterSet` `lob1-dev-clusters` (without assuming any `admin` privilege to existing `ManagedClusterSets`; necessary bindings are automatically defined to bind the user to `ClusterRole` `open-cluster-management:managedclusterset:lob1-dev-clusterss:admin`.
2. Then `skyler` adds user group `lob1-users` to `ClusterRole` `open-cluster-management:managedclusterset:lob1-dev-clusterss:edit` so that users in `lob-users` group are allowed to create new `ManagedClusters` with `managedClusterSet: lob1-dev-clusters`. 
3. Finally a user, `ursula` creates a `PlacementRule` with
```
  clusterReplicas: 3
  clusterSelector:
    matchLabels:
      apps/my-app: deployed
```
4. The `PlacementRule` will be scoped based on the `ManagedClusterSet` `lob1-dev-clusters` which may have 10 possible clusters but only 5 clusters that match the `PlacementRule`.

qiujian16:
I think we have two use cases for the `ManagedClusterSet` API change:
1. a cluster admin can create a exclusive `ManagedClusterSet` that the cluster in the  set cannot be added to another `ManagedClusterr`
2. an entity with lower privilege can add the `ManagedCluster` and add the `ManagedCluster` into `ManagedClusterSet`.  

Meanwhile, i think we want to leave the possibility for api support on non-exclusive `ManagedClusterSet`  in the  future. 
Could I suggest that we have a tradeoff solution as below:

1. On the `ManagedCluster`, user need to add a specific label to indicate that the `ManagedCluster` belongs to a certain `ManagedClusterSet`. 

```yaml
apiVersion: cluster.open-cluster-management.io/v1
kind: ManagedCluster
metadata:
  name: cluster1
  labels:
    clusters.open-cluster-management.io/managed-cluster-set: devclusterset
spec:
  hubAcceptsClient: true

```

The update of label needs a certain rule as:

```yaml
resources: ["clusterset/join"]
resourceNames: ["devclusterset"]
verbs: ["create"]
```

So user without  this rule cannot update this certain label on `ManagedCluster`.

2. In `ManagedClusterSet` API, we only leave `ExplicitClusterSetName` which means the cluster set is to use the certain label to select  clusters. 

```go
// ClusterSelector has three fields, if more than one field is set, then they are logically AND'd.
type ClusterSelector struct{
  // stuff
  // explicitClusterName is equivalent to setting `clusters.open-cluster-management.io/managed-cluster-set=explicitClusterName`.  This field gives an explicit intent that  is useful in cases where disjoint sets of managedclusters are desired.
  // If you want disjoint sets of managedclusters, you should also enable the OnlyDisjointMangedClusters validating admission webhook.
  ExplicitClusterSetName string
}
```

We remove labelSelector in `ClusterSelector` for now,  and revisit if we want to add it back in the future. For placementrule, i think finding cluster would be even easier, by adding this specific labelSelector as part of the labelSelector in placementrule API.

I think it should be able to support the two use cases, and we still can grow the API to support other use cases in the future.

cluster/v1alpha1/types.go
```diff
@@ -12,45 +12,24 @@ import (
...
+// `cluster.open-cluster-management.io/clusterset` on the ManagedCluster to refers to the ManagedClusterSet.
+// User is not allow to add/remove this label on a ManagedCluster unless they have a RBAC rule to CREATE on
+// a virtual subresource of managedclustersets/join. In order to update this label, user must have the permission
+// on both the old and new ManagedClusterSet.
 type ManagedClusterSet struct {
 	metav1.TypeMeta   `json:",inline"`
 	metav1.ObjectMeta `json:"metadata,omitempty"`
 
-	// Spec defines the attributes of the desired ManagedClusters
-	Spec ManagedClusterSetSpec `json:"spec"`
```
qiujian16, line 24: let's leave an empty spec

elgnay: Added the spec back.

PR #64, opened 2020-09-24
change printercolumn order in managedclusteraddon

leena-jawale:


hanqiuzh:
/approve 

Agree that available is most important and should be the first.
Looks good. Just want to see what will happen if I use this command 😂

hanqiuzh (approved):

PR #65, opened 2020-11-09
update contributing guide

skeeey:


CONTRIBUTING.md
```diff
@@ -17,6 +18,13 @@
 
 All contributions to the repository must be submitted under the terms of the [Apache Public License 2.0](https://www.apache.org/licenses/LICENSE-2.0).
 
+## Certificate of Origin
```
deads2k, line 21: How is  DCO different than a CLA? 

I remember this from a few years back. https://www.openshift.com/blog/keep-calm-and-merge-on-lowering-barriers-to-open-source-contributions-with-apache-v2

pmorie: I would prefer not to adopt any additional barriers to contribution

deads2k:
> How is DCO different than a CLA?

It appears the DCOs are intended to be lighter-weight than a CLA. If it becomes an inhibitor, it can be reassessed.

/approve

CONTRIBUTING.md
mdelder: The objective here is just to validate that the contribution adheres to a basic assertion that the contributor has rights to submit the change/code to the project. The intent is that the developer asserts this as part of their Pull Request. We do not expect that there will be a formal prerequisite or master list of contributors who "assert" they are within the terms of the DCO.

jberkus: The other reason is that we expect to eventually submit OCM to one of the various open source foundations, all of which like projects to have DCOs.

PR #72, opened 2021-03-30
Add placement APIs

elgnay:
The related task: https://github.com/open-cluster-management/backlog/issues/10715

The enhancement proposal: https://github.com/open-cluster-management/enhancements/tree/master/enhancements/sig-architecture/6-placements


cluster/v1alpha1/types.go
```diff
@@ -157,3 +160,230 @@ var ReservedClusterClaimNames = [...]string{
...
+	// 2) If a selected ManagedCluster ceases to match the selector (e.g. due to an update) after the
+	//    placement decisions are made, it will be eventually removed from the placement decisions;
+	// 3) If a ManagedCluster starts to match the selector after the placement decisions are made, it
+	//    will either be selected or at least has chance to be selected (when NumberOfClusters is
+	//    specified);
+	//
+	// Other scheduling strategies not yet implemented includes
+	// RequiredDuringSchedulingIgnoredDuringExecution and PreferredDuringSchedulingIgnoredDuringExecution
+	// +optional
+	RequiredDuringSchedulingRequiredDuringExecution *ClusterSelector `json:"requiredDuringSchedulingRequiredDuringExecution,omitempty"`
```
deads2k: What is the use-case for wanting this for some clustersets and not others?  Would it serve the function more effectively if it was directly on `PlacementSpec`?

deads2k: These are logically ANDed with the `clusterSets` if specified, right?

deads2k: The doc made me expect this an enumeration, with possible values of

1. `RequiredDuringSchedulingRequiredDuringExecution`
2. `RequiredDuringSchedulingIgnoredDuringExecution`

The idea of execution is a little weird.  Why would you want static placements?

cluster/v1alpha1/types.go
```diff
@@ -157,3 +160,230 @@ var ReservedClusterClaimNames = [...]string{
...
+
+// Affinity is a group of affinity scheduling rules.
+type Affinity struct {
+	// ClusterAntiAffinity describes cluster affinity scheduling rules for the placement.
+	// +optional
+	ClusterAntiAffinity *ClusterAntiAffinity `json:"clusterAntiAffinity,omitempty"`
+}
+
+// ClusterAntiAffinity is a group of cluster affinity scheduling rules.
+type ClusterAntiAffinity struct {
```
deads2k: Recommend getting someone associated with kube-scheduler or node to consider this API carefully.  Perhaps @damemi (scheduler) or @sjenning (node)

cluster/v1alpha1/types.go
```diff
@@ -157,3 +160,230 @@ var ReservedClusterClaimNames = [...]string{
...
+
+// PlacementDecision indicates a decision from a placement
+// PlacementDecision should has a label placement.open-cluster-management.io={placement name}
+// to reference a certain placement.
+type PlacementDecision struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	// Decisions is a slice of decisions according to a placement
+	// The number of decisions should not be larger than 100
```
deads2k, line 325: detail during implementation: make sure your sorting here is stable.

cluster/v1alpha1/types.go
damemi: this seems like a really similar goal to [descheduler](https://github.com/kubernetes-sigs/descheduler), which was also considered as a place to implement `Required/Required`. I don't think `RequiredDuringSchedulingRequiredDuringExecution` is implemented anywhere else in the upstream scheduler, and I'm not familiar with this project, so is your goal to essentially add that functionality here?

cluster/v1alpha1/types.go
elgnay: Yes, it is.

cluster/v1alpha1/types.go
elgnay: @damemi Yes, we are going to implement Required/Required in this API. I know it may introduced some potential issues, like
- Poor performance. Re-calculation is required once managed clusters change even after placement is scheduled;
- Unstable placement decisions. It may change frequently;

Do you have any other concerns on implementing such a feature?

damemi: No, I just didn't know if you were aware of the descheduler and curious if you had any other similar goals. 

Because this isn't implemented as a pod spec option upstream (https://github.com/kubernetes/api/blob/3a2d6b5bb7f84daa2125131865eb81b151448d5b/core/v1/types.go#L2746) this will be entirely on your end. That could expose you to some problems if this is later implemented upstream, and it conflicts with your implementation. But I think that risk is low

cluster/v1alpha1/types.go
elgnay: > What is the use-case for wanting this for some clustersets and not others? Would it serve the function more effectively if it was directly on `PlacementSpec`?

Moved `ClusterPredicate.ClusterSets` to `PlacementSpec.ClusterSets`

elgnay: > The doc made me expect this an enumeration, with possible values of
> 
>     1. `RequiredDuringSchedulingRequiredDuringExecution`
> 
>     2. `RequiredDuringSchedulingIgnoredDuringExecution`
> 
> 
> The idea of execution is a little weird. Why would you want static placements?

Renamed `RequiredDuringSchedulingRequiredDuringExecution` to `RequiredClusterSelector`.

deads2k:
So I've spent some time thinking about the interaction between `.spec.numberOfClusters`, `.spec.predicates[*]` and `.spec.affinity.clusterAntiAffinity`.  The case was described where a user matches on AWS cluster and OCP clusters and wants to specify a limit for each individual predicate, but the overlap between the two predicates (AWS and OCP) causes fewer clusters selected than the user intended.

I think the API presented here with `.spec.numberOfClusters` and `.spec.affinity.clusterAntiAffinity` instead of having numberOfClusters on each predicate is the best compromise.  Selecting a pool of clusters using the `.spec.predicates[*]`, then selecting from this pool of equal clusters based on the weighted (you probably want to add weights like the kube pod antiAffinity) anti-affinity rules will allow for balancing between the different predicates if it is desired.

If a person specifically desires X number of OCP clusters, the API will be clearer both to them and to us if they use multiple placements to describe it.  This parallels choices in kube like node selection.  If you want pods running on X nodes from region A and Y nodes from region B, you don't describe it in a single deployment.  You create two deployments, each with pod anti-affinity, and each targeting a different set of nodes.

I could not come up with an API that I thought worked well selecting numbersOfClusters by predicate.  I tried to build an example with an `ExclusiveOr` option, but in my attempts the mechanics of it rarely aligned with my desired behavior and didn't provide me more power than an predicate that was "OCP and not AWS" or "AWS and not OCP".

deads2k:
To elaborate slightly on the importance of weights.  In my local examples, I wanted to have anti-affinity based on vendor, platform, and geography, but the different anti-affinities had different levels of importance to me.  I most wanted to spread on geo to ensure good availability regardless of location.  Next was platform (AWS, GCE, Azure) to avoid a single vendor cloud outage bringing down my app. Finally was vendor, in case a single vendor had a security exposure and I faced a denial of service on my app running on that vendor.  (Just an example "if david was a sysadmin with an important app", not actual feedback.)

I think that weighting of anti-affinity would allow me to score those independently.

elgnay:
> To elaborate slightly on the importance of weights. In my local examples, I wanted to have anti-affinity based on vendor, platform, and geography, but the different anti-affinities had different levels of importance to me. I most wanted to spread on geo to ensure good availability regardless of location. Next was platform (AWS, GCE, Azure) to avoid a single vendor cloud outage bringing down my app. Finally was vendor, in case a single vendor had a security exposure and I faced a denial of service on my app running on that vendor. (Just an example "if david was a sysadmin with an important app", not actual feedback.)
> 
> I think that weighting of anti-affinity would allow me to score those independently.

Code updated on suggestion. Instead of weighting of anti-affinity, we try to address the balance problem with weighted cluster affinity.

deads2k:
Thinking through an example like

```yaml
apiVersion: cluster.open-cluster-management.io/v1alpha1
kind: Placement
metadata:
  name: example1
  namespace: ns1
spec:
  numberOfClusters: 5
  affinity:
    clusterAffinity:
      RequiredClusterAffinityTerms:
        - weight: 2
          clusterSelector:
            labelSelector:
              matchLabels:
                vendor: OpenShift
        - weight: 3
          clusterSelector:
            claimSelector:
              matchExpressions:
                - key: platform.open-cluster-management.io
                  operator: In
                  values:
                    - AWS
```

Based on weights being evaluated and additive, this is will result in a cluster running OCP on AWS getting a weight of five and being preferred.  Sometimes this is what I want: prefer the clusters that meet all these criteria.

However, sometimes I want the reverse.  I want to prefer clusters that are only OCP or AWS, but I'll settle for one that is both if that's all I have available.  This sort of negative weighting (what I *think* anti-affinity does) allows me to specify different attributes as having different negative weights. The weights are then subtracted from the score of a cluster 

```yaml
apiVersion: cluster.open-cluster-management.io/v1alpha1
kind: Placement
metadata:
  name: example1
  namespace: ns1
spec:
  numberOfClusters: 5
  affinity:
    clusterAntiAffinity:
      RequiredClusterAntiAffinityTerms:
        - weight: -2
          clusterSelector:
            labelSelector:
              matchLabels:
                vendor: OpenShift
        - weight: -3
          clusterSelector:
            claimSelector:
              matchExpressions:
                - key: platform.open-cluster-management.io
                  operator: In
                  values:
                    - AWS
```

In this example, I think that this is will result in a cluster running OCP on AWS getting a weight of -5 and being the last option considered compared to just AWS (-3) and just OCP (-2).

In working through this, I think I see the third thing now.  I think in some cases, you want the clusters already selected to influence the next cluster selected.  Affinity lets us prioritize intersecting criteria.  AntiAffinity lets us prioritize non-intersecting criteria.  The third category would be prioritizing clusters that have different selection criteria.  This is what pod anti-affinity does for pods by allowing currently scheduled pods to influence the next pod being scheduled (see https://github.com/kubernetes/api/blob/master/core/v1/types.go#L2707 being used here https://github.com/deads2k/cluster-openshift-apiserver-operator/blob/0556ad99ddb8ecf4a5fe5abf2e595d9acb1258f7/bindata/v3.11.0/openshift-apiserver/deploy.yaml#L206-L214).

The third category would allow something like, "select 5 OCP clusters on GCP, AWS, and Azure with a reasonable spread between them".

So, in conclusion, I think we want affiniity, anti-affinity, and this third type of affinity-against-working-set

elgnay:
In our original thought, `spec.affinity.clusterAffinity.requiredClusterAffinityTerms[*].Weight` is not used to calculate the score of each managed cluster, instead we try to use the ratio, `requiredClusterAffinityTerms[0].weight : requiredClusterAffinityTerms[1].weight2`, to balance the number of selected clusters matched different terms. In the first example above, let `N(i)` be the number of selected clusters which match the `requiredClusterAffinityTerms[i]`, then `N(0):N(1)` will be equal to `3:5` or as close as possible to `3:5`.

After discussion with @qiujian16 , we both think it's not a good idea because it might make user confused. But we'd like to keep the `ClusterAffinity` in `spec.affinity`, and the `Weight` of `RequiredClusterAffinityTerms` will be used to score the cluster as you suggested. It can only be set when `spec.NumberOfClusters` is set.

Another change we want to make is to rename `spec.affinity.ClusterAntiAffinity` to `WorkloadAntiAffinity` since it represent the anti-affinity relationships between workloads instead of clusters. An example will look like,
```yaml
apiVersion: cluster.open-cluster-management.io/v1alpha1
kind: Placement
metadata:
  name: clusters-across-regions
  namespace: ns1
spec:
  affinity:
    workloadAntiAffinity:
      requiredWorkloadAntiAffinityTerms:
        - topologyKey: region.open-cluster-management.io
          topologyKeyType: Claim

```

I know we have not solved the balance problem completely yet. It is worth a further discussion and maybe another enhancement proposal. How about we merge the basic part of the placement API first? and address the advanced feature later.

deads2k:
> How about we merge the basic part of the placement API first? and address the advanced feature later.

Yes.  I think it's worth removing affinity and merging with equal weighting of clusters.



> and maybe another enhancement proposal.

I think so. The three cases I outlined above are the ones I considered most interesting for an enhancement.

elgnay:
Removed affinity from spec of placement for now.

deads2k:
/approve
/lgtm
/hold

holding in case you have other interested reviewers. Feel free to release the hold when it's appropriate.

PR #73, opened 2021-04-06
Add addon registration configuration

qiujian16:


addon/v1alpha1/types_managedclusteraddon.go
```diff
@@ -29,8 +29,37 @@ type ManagedClusterAddOn struct {
 	Status ManagedClusterAddOnStatus `json:"status"`
 }
 
-// ManagedClusterAddOnSpec is empty for now.
+// ManagedClusterAddOnSpec defines the install configuration of
+// addon agent on managed cluster.
 type ManagedClusterAddOnSpec struct {
```
skeeey, line 34: would you add the `leaseDurationSeconds` on the spec to control the addon lease update frequency, by default, it is 60s, and would you add addon available condition type (`ManagedClusterAddOnConditionAvailable`)

addon/v1alpha1/types_managedclusteraddon.go
```diff
@@ -29,8 +29,68 @@ type ManagedClusterAddOn struct {
...
+}
+
+// Subject is the user subject of the addon agent to be registered to the hub.
+type Subject struct {
+	// user is the user name of the addon agent.
+	User string `json:"user"`
+
+	// uid is the user id of the addon agent.
+	// +optional
+	UID string `json:"uid,omitempty"`
```
deads2k: UID looks odd to me.  I don't think I'd let a client set this value.

addon/v1alpha1/types_managedclusteraddon.go
```diff
@@ -29,8 +29,68 @@ type ManagedClusterAddOn struct {
...
+
+	// groups is the user group of the addon agent.
+	// +optional
+	Groups []string `json:"groups,omitempty"`
+
+	// extras is the additional user info of the addon agent.
+	// These info includes organization unit, locality etc of the addon agent.
+	// +optional
+	Extras map[string][]string `json:"extras,omitempty"`
```
deads2k: I would avoid setting these as well.  The meaning isn't well defined in the kube API sense.

addon/v1alpha1/types_managedclusteraddon.go
```diff
@@ -29,8 +29,68 @@ type ManagedClusterAddOn struct {
...
+type RegistrationConfig struct {
+	// signerName is the name of signer that addon agent will use to create csr.
+	// +required
+	SignerName string `json:"signerName"`
+
+	// subject is the user subject of the addon agent to be registered to the hub.
+	// If it is not set, the addon agent will have the default subject
+	// "subject": {
+	//	"user": "system:open-cluster-management:addon:{addonName}:{clusterName}:{agentName}",
+	//  "groups: ["system:open-cluster-management:addon", "system:open-cluster-management:addon:{addonName}"]
```
deads2k: also system:authenticated should be added.

addon/v1alpha1/types_managedclusteraddon.go
```diff
@@ -29,8 +29,68 @@ type ManagedClusterAddOn struct {
...
+	// If it is not set, open-cluster-management-addon namespace is used to install the addon agent.
+	// +optional
+	InstallNamespace string `json:"installNamespace,omitempty"`
+
+	// leaseDurationSeconds is used to coordinate the lease update time of addon agents on the managed cluster.
+	// The addon agent should maintain a lease with the name of addon name in the namespace of installNamespace.
+	// The Klusterlet agent will monitor the lease and update the status of the related managedclusteraddon
+	// resource accordingly.  If its value is zero, the addon agent should update its lease every 60 seconds by default
+	// +optional
+	LeaseDurationSeconds int32 `json:"leaseDurationSeconds,omitempty"`
```
deads2k: This is slightly less clear to me.  do you mind making the case the in a separate PR.  Is your goal here to allow longer times or shorter times?  Who would choose to change this and why?

deads2k:
Great docs on this API. You need to add validation tags to the APIs to enforce things like max length and possibly patterns on some strings.

addon/v1alpha1/types_managedclusteraddon.go
qiujian16: There are some cases that a client may set OrganizationUnit, would it be fine to only add ou in the spec?

addon/v1alpha1/types_managedclusteraddon.go
```diff
@@ -29,8 +29,68 @@ type ManagedClusterAddOn struct {
 	Status ManagedClusterAddOnStatus `json:"status"`
 }
 
-// ManagedClusterAddOnSpec is empty for now.
+// ManagedClusterAddOnSpec defines the install configuration of
+// an addon agent on managed cluster.
 type ManagedClusterAddOnSpec struct {
+	// installNamespace is the namespace on the managed cluster to install the addon agent.
+	// If it is not set, open-cluster-management-addon namespace is used to install the addon agent.
```
skeeey: The default namespace may be `open-cluster-management-agent-addon`, ACM use `open-cluster-management-agent-addon` namespace as the addons installation namespace now

addon/v1alpha1/types_managedclusteraddon.go
qiujian16: removed

addon/v1alpha1/types_managedclusteraddon.go
qiujian16: updated

addon/v1alpha1/types_managedclusteraddon.go
qiujian16: removed for now. You are right, I have to rethink on this.

qiujian16:
add some validation. I do not find an easy regex to match signerName though.
And I will add more validation for other API in a separate PR

deads2k:
the updates look good. Thanks.

/approve
/lgtm
/hold

Holding for another reviewer if you have one.  If not, feel free to release the hold.

deads2k:
/lgtm

same hold rules.  if you want a look from someone else, feel free.  If not, feel free to release the hold.

PR #76, opened 2021-05-06
Bugfix: Adding `omitempty` jsontag for ManifestWork.status.conditions

yue9944882:
>$ kubectl apply --server-side -f manifestwork-example.yaml
....
....
I0506 15:03:19.363009   25916 request.go:1097] Request Body: {"apiVersion":"work.open-cluster-management.io/v1","kind":"ManifestWork","metadata":{"creationTimestamp":"2021-04-30T12:18:44Z","labels":{"foo":"bar"},"name":"demo-test","namespace":"zuoxiu-local"},"status":{"conditions":[]}} 
....
....
I0506 15:03:19.386561   25916 request.go:1097] Response Body: {"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"Apply failed with 1 conflict: conflict with \"work\" using work.open-cluster-management.io/v1: .status.conditions","reason":"Conflict","details":{"causes":[{"reason":"FieldManagerConflict","message":"conflict with \"work\" using work.open-cluster-management.io/v1","field":".status.conditions"}]},"code":409}
F0506 15:03:19.387107   25916 helpers.go:115] error: Apply failed with 1 conflict: conflict with "work" using work.open-cluster-management.io/v1: .status.conditions
Please review the fields above--they currently have other managers. Here...

b/c the status.conditions field doesnt have the omitempty tag, the server-side applying for manifestwork resources will be constantly failing due to apply conflict. as we can tell from the patch request body, a correct SSA body omitting status subresource should be `{"status":{}}`,  this pull will make SSA work on the manifestwork resources.

### DIFF
--- a/work/v1/types.go
+++ b/work/v1/types.go
@@ -117,7 +117,7 @@ type ManifestWorkStatus struct {
 	// 3. Available represents workload in ManifestWork exists on the managed cluster.
 	// 4. Degraded represents the current state of workload does not match the desired
 	// state for a certain period.
-	Conditions []metav1.Condition `json:"conditions"`
+	Conditions []metav1.Condition `json:"conditions,omitempty"`
 
 	// ResourceStatus represents the status of each resource in manifestwork deployed on a
 	// managed cluster. The Klusterlet agent on managed cluster syncs the condition from the managed cluster to the hub.


qiujian16:
I think it needs `make update`

/ok-to-test

qiujian16:
/assign @deads2k 

~It is a bit weird to me since it means any status field has to be omitempty given the status itself is omitempty already?~

Discussed with @yue9944882 and  it seems server-side-apply will generate patch considering status field even though it is ignored when calling rest. We can also consider using another  merge strategy here, but omitempty is the most straightforward approach. 

/lgtm

deads2k:
How can an update to spec cause a conflict with status?  That seems like a more pressing concern for server-side apply itself.  If updates to spec adjust a field manager for status, then it seems that our controllers updating status would risk conflicts with your client setting spec.

Can you followup with @apelisse in kube-slack #sig-api-machinery is a decent place to start.

apelisse:
I think that's a duplicate of https://github.com/kubernetes/kubernetes/issues/75564, which we fixed in v1.21.

yue9944882:
@deads2k @apelisse thanks, i will verify it's fixed after upgrading our clusters to 1.21. btw is there any plan to pick it to 1.19 or 1.20?

deads2k:
> @deads2k @apelisse thanks, i will verify it's fixed after upgrading our clusters to 1.21. btw is there any plan to pick it to 1.19 or 1.20?

My memory is that the change was significant API change because of the adjustment to managed fields.  In kube, I don't think we will backport.

Regarding this PR here.  Please confirm the problem is fixed with 1.21 so we won't face it in the GA apply feature.  After that, I'm willing to consider a change here for conditions.  Get me on slack.

yue9944882:
> Regarding this PR here. Please confirm the problem is fixed with 1.21 so we won't face it in the GA apply feature.

just confirmed it's fixed in 1.21 release. SSA works like a charm now.

deads2k:
The original SSA bug has been fixed, so in 1.21 this is no longer necessary. However, this change makes apply function on older releases of kube.  It is technically breaking in that .status.conditions was always present in serializations before, but it seems unlikely that someone was distinguishing the cases in a meaningful way.

/approve

I leave lgtm with @qiujian16 
/assign @qiujian16

PR #77, opened 2021-05-11
add placementImagePullSpec into ClusterManager spec

elgnay:
The related task: https://github.com/open-cluster-management/backlog/issues/12423


operator/v1/types.go
```diff
@@ -27,12 +27,19 @@ type ClusterManager struct {
 // ClusterManagerSpec represents a desired deployment configuration of controllers that govern registration and work distribution for attached Klusterlets.
 type ClusterManagerSpec struct {
 	// RegistrationImagePullSpec represents the desired image of registration controller/webhook installed on hub.
-	// +required
-	RegistrationImagePullSpec string `json:"registrationImagePullSpec"`
+	// +optional
```
deads2k, line 30: need to tell me in the godoc what each of these fields do if they are not set

PR #79, opened 2021-05-27
update placement doc

elgnay:

### DIFF
--- a/cluster/v1alpha1/types.go
+++ b/cluster/v1alpha1/types.go
@@ -276,8 +276,15 @@ type PlacementStatus struct {
 
 const (
 	// PlacementConditionSatisfied means Placement requirements are satisfied.
-	// A placement is not satisfied only if there is empty ClusterDecision in the status.decisions
-	// of PlacementDecisions.
+	// A placement is not satisfied if
+	// 1) No ManagedClusterSetBinding is found in the placement namespace;
+	// 2) ClusterSets in spec is specified and none of them is bound to the placement namespace;
+	// 3) No memeber ManagedCluster is found for all eligible ManagedClusterSets; A ManagedClusterSet
+	//    is eligible for a placement if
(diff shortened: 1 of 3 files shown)


cluster/v1alpha1/types.go
```diff
@@ -276,8 +276,15 @@ type PlacementStatus struct {
 
 const (
 	// PlacementConditionSatisfied means Placement requirements are satisfied.
-	// A placement is not satisfied only if there is empty ClusterDecision in the status.decisions
-	// of PlacementDecisions.
+	// A placement is not satisfied if
+	// 1) No ManagedClusterSetBinding is found in the placement namespace;
+	// 2) ClusterSets in spec is specified and none of them is bound to the placement namespace;
+	// 3) No memeber ManagedCluster is found for all eligible ManagedClusterSets; A ManagedClusterSet
```
qiujian16, line 282: There is no ManagedCluster in the  ManagedClusterSets specified in the spec. And I do not think you need a and b

elgnay: > There is no ManagedCluster in the ManagedClusterSets specified in the spec. And I do not think you need a and b

The above description does not cover the case: placement without Clustersets specified in the spec.

qiujian16: Would you like to update a bit on this description?

I think it will be

```
1) No ManagedClusterSetBinding in the Placement namespace;
2) ClusterSets in the Placement spec is specified and none of them is bound to the namespace of the Placement;
3) No ManagedCluster in the ClusterSets which are specified in the Placement spec or bound to this namespace if ClusterSets is not specified in the Placement spec.
4) No ManagedCluster matches any of the cluster predicates in the Placement spec;
5) NumberOfSelectedClusters in status is less than NumberOfClusters in the Placement spec if it is specified;
```

PR #83, opened 2021-06-28
Add delete option in work api

qiujian16:
implement https://github.com/open-cluster-management-io/enhancements/tree/main/enhancements/sig-architecture/10-deletepropagationstrategy


work/v1/types.go
```diff
@@ -46,6 +51,56 @@ type ManifestsTemplate struct {
...
+type DeleteOption struct {
+	// propagationPolicy can be Orphan or SelectivelyOrphan
+	// SelectivelyOrphan should be rarely used.  It is provided for cases where particular resources is transfering
+	// ownership from one ManifestWork to another or another management unit.
+	// Setting this value will allow a flow like
+	// 1. create manifestwork/2 to manage foo
+	// 2. update manifestwork/1 to selectively orphan foo
+	// 3. remove foo from manifestwork/1 without impacting continuity because manifestwork/2 adopts it.
+	// +optional
+	PropagationPolicy DeletePropagationPolicyType `json:"propagationPolicy,omitempty"`
```
deads2k: There are some discriminated union kubebuilder tags you can use.

work/v1/types.go
```diff
@@ -46,6 +51,56 @@ type ManifestsTemplate struct {
 	Manifests []Manifest `json:"manifests,omitempty"`
 }
 
+type DeleteOption struct {
+	// propagationPolicy can be Orphan or SelectivelyOrphan
```
deads2k: I think Foreground needs to be an option too, right?  That way I can have this stanza and explicitly indicate my desire for the current default.

work/v1/types.go
```diff
@@ -30,6 +30,11 @@ type ManifestWork struct {
 type ManifestWorkSpec struct {
 	// Workload represents the manifest workload to be deployed on a managed cluster.
 	Workload ManifestsTemplate `json:"workload,omitempty"`
+
+	// DeleteOption represents deletion strategy when the manifestwork is deleted.
+	// Foreground deletion strategy is applied to all the resource in this manifestwork if it is not set.
+	// +optional
+	DeleteOption DeleteOption `json:"deleteOption,omitempty"`
```
deads2k: on non-configuration/operator APIs, adding new structs as pointers allows empty serialization.

work/v1/types.go
deads2k: now that you've switch to CRD v1, you can  use defaulting to set this field to Foreground by default.

deads2k:
Some minor comments.  The API looks good overall.

work/v1/types.go
```diff
@@ -46,6 +51,59 @@ type ManifestsTemplate struct {
...
 
+type DeleteOption struct {
+	// propagationPolicy can be ForeGround, Orphan or SelectivelyOrphan
+	// SelectivelyOrphan should be rarely used.  It is provided for cases where particular resources is transfering
+	// ownership from one ManifestWork to another or another management unit.
+	// Setting this value will allow a flow like
+	// 1. create manifestwork/2 to manage foo
+	// 2. update manifestwork/1 to selectively orphan foo
+	// 3. remove foo from manifestwork/1 without impacting continuity because manifestwork/2 adopts it.
+	// +kubebuilder:default=ForeGround
```
deads2k, line 62: nit: Foreground

work/v1/types.go
```diff
@@ -46,6 +51,59 @@ type ManifestsTemplate struct {
...
+	// selectivelyOrphan represents a list of resources following orphan deletion stratecy
+	SelectivelyOrphan *SelectivelyOrphan `json:"selectivelyOrphans,omitempty"`
+}
+
+type DeletePropagationPolicyType string
+
+const (
+	// DeletePropagationPolicyTypeForeGround represents that all the resources in the manifestwork is should
+	// be fourground deleted.
+	DeletePropagationPolicyTypeForeGround DeletePropagationPolicyType = "ForeGround"
```
deads2k: I think it's normally Foreground, not ForeGround

deads2k:
I think it's normally Foreground.  lgtm after the spelling update.

PR #90, opened 2021-08-17
Addon-Api: Optional structs should be pointer

yue9944882:
the tags `omitempty` on the the plain struct will not work, and conventionally in k8s api semantics we're supposed to make the struct field with `// +optional` marker a pointer to clarify its optionality. also, this will help trimming serialized json/yaml outputs.


qiujian16:
We would need to doc for any pointer the difference between a nil and empty struct. Commonly if there is not big difference, we would avoid using nil.

qiujian16:
We would need to doc for any pointer the difference between a nil and empty struct. Commonly if there is not big difference, we would avoid using nil.

PR #93, opened 2021-08-26
update placement API with PrioritizerConfigs

haoqing0110:
Ref: https://github.com/open-cluster-management/backlog/issues/15599


cluster/v1alpha1/types.go
```diff
@@ -267,6 +271,20 @@ type ClusterClaimSelector struct {
 	MatchExpressions []metav1.LabelSelectorRequirement `json:"matchExpressions,omitempty"`
 }
 
+// PrioritizerConfig represents the configuration of prioritizer
+type PrioritizerConfig struct {
+	// Name is the name of prioritizer
```
qiujian16: Name is the name of a prioritizer, and it also need to doc what is the valid name today.

cluster/v1alpha1/types.go
```diff
@@ -267,6 +271,20 @@ type ClusterClaimSelector struct {
...
+	// +required
+	Name string `json:"name"`
+
+	// Weight defines the weight of prioritizer. The value must be ranged in [0,10].
```
qiujian16, line 313: it worth explaining here the implication of the weight. What does it mean if it has a higher/lower weight. You might say like

each prioritizer will calculate an integer score of a cluster in the range of [-100, 100]. The final score of a cluster will be sum(weight * prioritizer). A higher weight indicates that the prioritizer weighs more in the cluster selection, while 0 weight indicate thats the prioritizer is disabled.

cluster/v1alpha1/types.go
haoqing0110: Fixed.

cluster/v1alpha1/types.go
haoqing0110: Fixed.

cluster/v1alpha1/types.go
```diff
@@ -267,6 +271,24 @@ type ClusterClaimSelector struct {
 	MatchExpressions []metav1.LabelSelectorRequirement `json:"matchExpressions,omitempty"`
 }
 
+// PrioritizerConfig represents the configuration of prioritizer
+type PrioritizerConfig struct {
+	// Name is the name of a prioritizer.
+	// The valid names are: balance, steady, ResourceRatioCPU, ResourceAllocatableCPU, ResourceRatioMemory, ResourceAllocatableMemory.
```
qiujian16: s/balance/Balance/
s/steady/Steady

I would think we also add some explanations on each.

haoqing0110: @qiujian16 thanks for the comment, update in the latest code.

cluster/v1alpha1/types.go
```diff
@@ -267,6 +271,52 @@ type ClusterClaimSelector struct {
...
+// PrioritizerConfig represents the configuration of prioritizer
+type PrioritizerConfig struct {
+	// Name is the name of a prioritizer. Below are the valid names:
+	// 1) Balance
+	//    Balance prioritizer balance the number of decisions among the clusters.
+	//    The cluster with the highest number of decison is given the lowest score,
+	//    while the empty cluster is given the highest score. The score ranges from -100 to 100.
+	// 2) Steady
+	//    Steady prioritizer ensure the existing decision is stabilized.
+	//    The clusters that existing decisions choose are given the highest score while the clusters
```
qiujian16: choose --> chosen

cluster/v1alpha1/types.go
```diff
@@ -267,6 +271,52 @@ type ClusterClaimSelector struct {
 	MatchExpressions []metav1.LabelSelectorRequirement `json:"matchExpressions,omitempty"`
 }
 
+// PrioritizerConfig represents the configuration of prioritizer
+type PrioritizerConfig struct {
+	// Name is the name of a prioritizer. Below are the valid names:
+	// 1) Balance
```
qiujian16: maybe we can do like

Name is the name of a prioritizer. Below are the valid names:
1) Balance: balance the decisions among the cluster.
2) Steady: ensure the existing decision is stabilized.
3) ResourceRatioCPU & ResourceRatioMemory: sort cluster based on the allocatable to capacity ratio.
4) ResourceAllocatableCPU & ResourceAllocatableMemory: sort cluster based on the allocatable.

qiujian16:
/lgtm

/assign @deads2k 
would you like to take a a look at this.

cluster/v1alpha1/0000_03_clusters.open-cluster-management.io_placements.crd.yaml
```diff
@@ -12,7 +12,17 @@ spec:
...
+  - additionalPrinterColumns:
+    - jsonPath: .status.conditions[?(@.type=="PlacementSatisfied")].status
+      name: Succeeded
+      type: string
+    - jsonPath: .status.conditions[?(@.type=="PlacementSatisfied")].reason
+      name: Reason
+      type: string
+    - jsonPath: .status.numberOfSelectedClusters
+      name: SelectedClusters
+      type: integer
```
elgnay, line 24: I think you can add print columns by adding kubebuilder markers on go struct instead of updating the generated CRD directly. Here is an example: https://github.com/open-cluster-management-io/api/blob/9667ae902e4b505bbcecd8eedfebcc2281718f3b/addon/v1alpha1/types_clustermanagementaddon.go#L12-L13

haoqing0110: @elgnay Not sure if it's my kubebuilder version issue, the sequence changes. This part is the same as https://github.com/open-cluster-management-io/api/pull/93/files#diff-8584af8df06b8cc38c1d152b48e5c590f870be9b792b6abb178c66f89201c41fL265.

cluster/v1alpha1/types.go
haoqing0110: @qiujian16 Thanks. The original idea is, as we expect users to set the weight to some Prioritizer, give more details like score range may help. If it's not necessary, it's good to me to simplify the words.

cluster/v1alpha1/types.go
```diff
@@ -233,6 +233,10 @@ type PlacementSpec struct {
 	// Predicates represent a slice of predicates to select ManagedClusters. The predicates are ORed.
 	// +optional
 	Predicates []ClusterPredicate `json:"predicates,omitempty"`
+
+	// PrioritizerConfigs override the default configuration of the prioritizers
```
deads2k: describe what empty means.  Is there a difference between empty and nil?

If I have no preference, is there some kind of  default that I can get?

deads2k: If I express some preferences for say balance and steady, and a new prioritizer like resourceratioCPU comes along, what happens?

haoqing0110: 1. For the empty and nil question. 
If no PrioritizerConfigs defined(empty and nil) , the placement will use default Prioritizer setting. That is "Steady" and "Balance" with weight 1, other prioritizers with weight 0. Will update the doc.
2. Is the second question an upgrade case? 
I suppose for any new prioritizer added in the future, if their weight is 0 in the default setting, user needs to update the placement with new prioritizer explicitly to enable it.
/cc @elgnay @qiujian16

cluster/v1alpha1/types.go
```diff
@@ -233,6 +233,13 @@ type PlacementSpec struct {
 	// Predicates represent a slice of predicates to select ManagedClusters. The predicates are ORed.
 	// +optional
 	Predicates []ClusterPredicate `json:"predicates,omitempty"`
+
+	// PrioritizerConfigs defines the configurations of the prioritizers.
+	// Default configuration is used if it is not specified, in which Steady and Balance prioritizers have the weight of 1
+	// while other prioritizers have the weight of 0.
+	// Referring to PrioritizerConfig to see more description about prioritizer valid names and weight.
+	// +optional
+	PrioritizerConfigs []PrioritizerConfig `json:"prioritizerConfigs,omitempty"`
```
deads2k: @sttts  Can we distinguish between nil and empty on this?

cluster/v1alpha1/types.go
```diff
@@ -233,6 +233,13 @@ type PlacementSpec struct {
 	// Predicates represent a slice of predicates to select ManagedClusters. The predicates are ORed.
 	// +optional
 	Predicates []ClusterPredicate `json:"predicates,omitempty"`
+
+	// PrioritizerConfigs defines the configurations of the prioritizers.
+	// Default configuration is used if it is not specified, in which Steady and Balance prioritizers have the weight of 1
```
deads2k: If I specific `{ResourceRatioCPU}`, what is the resulting weight of `Steady`?

cluster/v1alpha1/types.go
sttts: `nullable` is a think in openapi. But it is an anti-pattern to distinguish between unset and null. Why is it needed? I would prefer seeing an API that doesn't depend on that. Note that things like JSON merge patch breaks with null.

deads2k: > But it is an anti-pattern to distinguish between unset and null

Between `nil` and `{}`.  Where one could be "no opinion" and the other could be "I have an opinion and want no prioritizers" as opposed to enumerating an ever increasing list..

sttts: so you actually want `undefined`, `null`, `{}`, all meaning something different?

deads2k: I don't think we need to have undefined different than null.  But I think I see benefit to `null != {}`.  The CRD schema generation allows the distinction?

How does the json merge failure manifest?

cluster/v1alpha1/types.go
qiujian16: it (Steady) will still use the default weight of 1 if it is not specified here.

cluster/v1alpha1/types.go
haoqing0110: As discussed with @qiujian16 and @elgnay today, we would like to distinguish between nil and empty. The behavior is:
1. If PrioritizerConfigs is nil (unset), use the default config "steady weight 1, balance weight 1, others weight 0"
2. If PrioritizerConfigs is empty (PrioritizerConfigs: []), disable all the Prioritizers with weight 0.
3. If PrioritizerConfigs is defined, placement will use what user defined (replace default config). That means,
  * If the user only wants to use resourceratioCPU, should config with 
```
  prioritizerConfigs:
    - name: ResourceRatioCPU
      weight: 2
```
  * If the user wants to use resourceratioCPU, as well as the default "steady and balance", should config with 
```
  prioritizerConfigs:
    - name: ResourceRatioCPU
      weight: 1
    - name: Steady
    - name: Balance
```

cluster/v1alpha1/types.go
qiujian16: after discussed with @elgnay and @haoqing0110. We think If we specific `{ResourceRatioCPU}` but not specify `Steady`, it means the `Steady` is disabled. 

If the whole PrioritizerConfigs is not set, it will use the default values which is `Steady` & `Balance` =1, and others = 0

cluster/v1alpha1/types.go
```diff
@@ -233,6 +233,15 @@ type PlacementSpec struct {
 	// Predicates represent a slice of predicates to select ManagedClusters. The predicates are ORed.
 	// +optional
 	Predicates []ClusterPredicate `json:"predicates,omitempty"`
+
+	// PrioritizerConfigs defines the configurations of the prioritizers.
+	// If this field is unset, i.e. nil, then default configuration is used, in which
+	// Steady and Balance prioritizers have the weight of 1 while other prioritizers have the weight of 0.
```
deads2k: minor: indicate that this default may change over time.

cluster/v1alpha1/types.go
deads2k: > * If the user wants to use resourceratioCPU, as well as the default "steady and balance", should config with
> 
> ```
>   prioritizerConfigs:
>     - name: ResourceRatioCPU
>       weight: 1
>     - name: Steady
>     - name: Balance
> ```

In this example, why isn't the steady/balance weight set to 1?

deads2k: Otherwise I think this behavior makes sense.

cluster/v1alpha1/types.go
```diff
@@ -233,6 +233,15 @@ type PlacementSpec struct {
...
+	// If you don't want to use any prioritizer, set this field to an empty slice, i.e. `prioritizerConfigs: []`
+	// in the YAML file, then all the prioritizers will be disabled.
```
deads2k: You should explicitly indicate that once a single prioritizer has been configured, then every desired prioritizer *must* be enumerated.

I suggest you make a note in your enhancement that this behavior should be reassessed for whether it has worked well before beta.

It is also possible that you'll want a config like

```
PrioritizerConfig PrioritizerConfig

PrioritizerConfig struct{
  // mode is either Exact, Additive, "" where "" is Additive by default.
  // in Additive mode, any prioritizer not explicitly enumerated is enabled in its default mode
  // in Exact mode, any prioritizer not explicitly enumerated is weighted as zero
  // Exact requires knowing the full set of prioritizers you want, but avoids behavior changes between releases
  // Additive doesn't require configuring all prioritizers.  As new prioritizers are added, additional
  // prioritization will happen
  Mode string

  PrioritizerConfigs []PrioritizerConfig
}

```

cluster/v1alpha1/types.go
haoqing0110: As the default weight is 1. The above example has a typo that may make you confused. Below 2 should be the same:
```
  prioritizerConfigs:
    - name: ResourceRatioCPU
      weight: 2
    - name: Steady
    - name: Balance
 ```
```
  prioritizerConfigs:
    - name: ResourceRatioCPU
      weight: 2
    - name: Steady
      weight: 1
    - name: Balance
      weight: 1
 ```

cluster/v1alpha1/types.go
haoqing0110: @deads2k thanks, this makes sense to us.

deads2k:
I think this fell out pretty well, but keep track of how well it works on the way to beta.

/lgtm
/approve
/hold

holding for additional reviewers if desired.

PR #94, opened 2021-08-30
Add nvidia gpu support

suigh:
Signed-off-by: suigh <suigh@cn.ibm.com>

The registration component will be enhanced to report the GPU information to hub cluster. As types.go in API will be used by this enhancement, update types.go here, I will update the corresponding one in registration later.

### DIFF
--- a/cluster/v1/types.go
+++ b/cluster/v1/types.go
@@ -150,7 +150,7 @@ const (
 	ResourceCPU ResourceName = "cpu"
 	// ResourceMemory defines the amount of memory in bytes. (500Gi = 500GiB = 500 * 1024 * 1024 * 1024)
 	ResourceMemory ResourceName = "memory"
-	// ResourceGPU defines the number of GPUs on a node
+	// ResourceNvidiaGPU defines the number of GPUs on a node
 	ResourceNvidiaGPU ResourceName = "nvidia.com/gpu"
 )
 
--- a/cluster/v1/types.go
+++ b/cluster/v1/types.go
@@ -150,6 +150,8 @@ const (
 	ResourceCPU ResourceName = "cpu"
 	// ResourceMemory defines the amount of memory in bytes. (500Gi = 500GiB = 500 * 1024 * 1024 * 1024)
 	ResourceMemory ResourceName = "memory"
+	// ResourceGPU defines the number of GPUs on a node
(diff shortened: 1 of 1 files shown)


cluster/v1/types.go
```diff
@@ -150,6 +150,8 @@ const (
 	ResourceCPU ResourceName = "cpu"
 	// ResourceMemory defines the amount of memory in bytes. (500Gi = 500GiB = 500 * 1024 * 1024 * 1024)
 	ResourceMemory ResourceName = "memory"
+	// ResourceNvidiaGPU defines the number of GPUs on a node
+	ResourceNvidiaGPU ResourceName = "nvidia.com/gpu"
```
qiujian16, line 154: do we need to define here? I think it can be set even it is not defined as const here. Also is it necessary to add `nvidia` here?

suigh: @qiujian16 
Thanks for your comments.

Defining this variable here is to make the codes clean as registration uses "ResourceCPU", "ResourceMemory" that defined here in the codes. If you prefer to use the "nvidia.com/gpu" directly in the related codes, I will do that.

For the "nvidia.com/gpu", it is the resource name that reported by [nvidia gpu plugin](https://github.com/NVIDIA/k8s-device-plugin) , and it is also used by the k8s, this is why I didn't use a simple name like `gpu` here. This name can help us to handle gpus from different companies in the future.

Here is the related codes:
```
		return clusterv1.ResourceList{
				clusterv1.ResourceCPU:    cpuCapacity,
				clusterv1.ResourceMemory: formatQuantityToMi(memoryCapacity),
			},
			clusterv1.ResourceList{
				clusterv1.ResourceCPU:    cpuAllocatable,
				clusterv1.ResourceMemory: formatQuantityToMi(memoryAllocatable),
			}, nil
```

qiujian16: I think it is not normal to add a vendor name in API, we commonly do not do that.  If we want to sort by GPU it is possible that user can set `ResourceAllocatableNVIDIAGPU` on placement without having an explicit definition here. 

For registration, we could merge all the resources but not just CPU/memory. WDYT?

suigh: Agree. I will close this request.

PR #96, opened 2021-09-03
add nodeSelector and tolerations for the ClusterManager operator

zhiweiyin318:
this PR is to add nodeSelector and Tolerations for pods to run on the specified nodes like master, ocp infra nodes.

Signed-off-by: Zhiwei <zyin@redhat.com>


operator/v1/types.go
```diff
@@ -40,6 +41,14 @@ type ClusterManagerSpec struct {
 	// +optional
 	// +kubebuilder:default=quay.io/open-cluster-management/placement
 	PlacementImagePullSpec string `json:"placementImagePullSpec,omitempty"`
+
+	// NodeSelector defines which Nodes the Pods are scheduled on. The default is an empty list.
+	// +optional
```
qiujian16, line 46: put NodeSelector and Tolerations in a struct.

operator/v1/types.go
```diff
@@ -40,6 +41,14 @@ type ClusterManagerSpec struct {
...
+	NodeSelector map[string]string `json:"nodeSelector,omitempty"`
+
+	// Tolerations defines the tolerations for the Pods. The default is an empty list.
```
qiujian16: I think you would need to ref this to the definition of toleration. Current doc is not clear.

zhiweiyin318: done. put them in NodePlacement

operator/v1/types.go
zhiweiyin318: done. put them in NodePlacement

operator/v1/types.go
zhiweiyin318: done

PR #98, opened 2021-09-15
add clusterset v1beta1

elgnay:
The related task: https://github.com/open-cluster-management/backlog/issues/16155


cluster/v1beta1/doc.go
```diff
@@ -0,0 +1,9 @@
+// Package v1beta1 contains API Schema definitions for the cluster v1beta1 API group
+// +k8s:deepcopy-gen=package,register
+// +k8s:conversion-gen=open-cluster-management.io/api/cluster
+// +k8s:defaulter-gen=TypeMeta
+// +k8s:openapi-gen=true
+
+// +kubebuilder:validation:Optional
```
qiujian16, line 7: why need this here?

elgnay: It sets the default validation rule for fields of structs in this package. Other API versions,  including cluster v1/v1alpha1, work and addon, have this setting as well.

PR #99, opened 2021-09-27
update placement API

haoqing0110:
Ref: https://github.com/open-cluster-management/backlog/issues/15599
Refer to kubernetes node-allocatable definition https://kubernetes.io/docs/tasks/administer-cluster/reserve-compute-resources/#node-allocatable, it's a relatively stable value. 
As OCM allocatable is a sum of all the node's allocatable, so using the allocatable to capacity ratio doesn't help so much in scheduling. 
Remove it until we find a better solution. 


qiujian16 (commented):
/approve
/lgtm

PR #101, opened 2021-10-31
update docs for cluster join process

### LINKED ISSUE #87: request for docs updating for cluster join process
Currently, I followed [quick start of registration](https://github.com/open-cluster-management-io/registration#quickstart) to register a cluster, and the cluster registered succeeded, but I found the docs of [cluster join process](https://github.com/open-cluster-management-io/api/blob/main/docs/clusterjoinprocess.md) is outdated. For example:
- [x] clusterrolebinding's name is not correct, I did not find any clusterrolebing named `system:open-cluster-management:clusterName:agentName` in my cluster, I think it should be `open-cluster-management:managedcluster:clusterName`(https://github.com/open-cluster-management-io/registration/pull/55 gave me some clue) 
> 5. hub-controller creates a clusterrolebinding on the hub with the identity of `system:open-cluster-management:clusterName:agentName`

The [cluster join process](https://github.com/open-cluster-management-io/api/blob/main/docs/clusterjoinprocess.md) doc is very important for newcomers to understand the whole cluster registration process, very nice if it could be updated(very pleased to try to do it by myself when I figure out the whole process). Thanks a lot.

zhujian7:
Fix: https://github.com/open-cluster-management-io/api/issues/87
Signed-off-by: zhujian <jiazhu@redhat.com>

### DIFF
--- a/docs/clusterjoinprocess.md
+++ b/docs/clusterjoinprocess.md
@@ -17,28 +17,26 @@ and deploy the agent on managed cluster.
   - it has the identity to create `ManagedCluster` and create/watch csr.
 2. agent on managed cluster creates `ManagedCluster` if it does not exist.
   - The name of `ManagedCluster` is read from Cluster UID in openshift.
-  - Otherwise agent generates a UID and saves it in `Configmap` in managed cluster, so restarting agent or redeploying
-  agent will not lose the UID.
+  - Otherwise agent generates an agent UID and use it as part of `CommonName` to request a certificate(step 3), and once the CSR is approved on the hub, the agent on the managed cluster will store it as `agent-name` in `Secret` `hub-kubeconfig-secret`(step 10), so restarting agent or redeploying agent will not lose the UID after the cluster is managed successfully.
 3. agent on managed cluster creates CSR on hub cluster using bootstrap kubeconfig.
-  - The subject in csr is
+  - The subject in CSR is
 `{"Organization": ["system:open-cluster-management:clusterName"], "CommonName":"system:open-cluster-management:clusterName:agentName"}`.
   - The name of the csr is the digest of subject and private key, with a common prefix.
   CSR will specify the signer name as the kube-client one.
 4. cluster-admin on hub-cluster approve the CSR.
 5. hub-controller creates a clusterrolebinding on the hub with the identity of
-`system:open-cluster-management:clusterName:agentName`
+`open-cluster-management:managedcluster:clusterName`
    - Allows status update of `ManagedCluster`
 6. cluster-admin on hub update `spec.hubAcceptsClient` to `true`.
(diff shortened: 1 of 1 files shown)


docs/clusterjoinprocess.md
```diff
@@ -17,28 +17,26 @@ and deploy the agent on managed cluster.
...
    - Allows status update of `ManagedCluster`
 6. cluster-admin on hub update `spec.hubAcceptsClient` to `true`.
-  - Only user on hub who has the RBAC permision to update subresource of `managedclusters/accept`
+  - Only user on hub who has the RBAC permission to update subresource of `managedclusters/accept`
   can update this field.
-7. hub-controller updates condition of `ManagedCluster` to `HubApprovedJoin`.
+7. hub-controller updates condition of `ManagedCluster` to `HubAcceptedManagedCluster`.
 8. hub-controller creates a namespace as the name of cluster on hub cluster if it does not exist.
   - managed cluster can only join a hub once, and it can join to multiple hubs.
   - The UID of the managed cluster is identical on each of the hub the Klusterlet agent joins.
```
zhujian7, line 36: Here I have a question: as far as I know, cluster join starting with a bootstrap-secret(specified by a flag of the spoken agent, for example, `bootstrap-secret`),  there only can be **one** `bootstrap-secret`(representing one hub) in the agent running namespace, so I do not know how to register one managed cluster to multiple hubs by one agent? Of course, we can run multiple agents in separate namespaces to register the cluster to multiple hubs, but I am curious how to keep the UID identical.

qiujian16: UID here is cluster name, and it is identical in hub 1, it should also be identical in hub2.

zhujian7: Sorry I got it wrong, thanks for the explanation

docs/clusterjoinprocess.md
```diff
@@ -56,6 +54,6 @@ based on the following steps:
 - check if organization field and commonName field is valid.
 - check if user name in csr is the same as commonName in certificate to ensure the request
 is originated from the same identity.
-- check if the corresponding `ManagedCluster` is in the conndition of `HubApprovedJoin`.
+- check if the corresponding `ManagedCluster` is in the conndition of `HubAcceptedManagedCluster`.
```
zhujian7: I did not find this check logic in [hub-csr-controller](https://github.com/open-cluster-management-io/registration/blob/main/pkg/hub/csr/controller.go), should we remove this, or add the check logic into hub-csr-controller?
cc @qiujian16 @skeeey

qiujian16: oh let's remove this, accepted is only about authz. So we do not need to check this for csr.

zhujian7:
@qiujian16 PTAL

PR #103, opened 2021-11-05
Add spec fields in clusterManager

xuezhaojun:
Signed-off-by: xuezhaojun <zxue@redhat.com>


xuezhaojun:
This one is related with https://github.com/open-cluster-management-io/registration-operator/issues/157#event-5531255134.

And a draft of design is https://app.excalidraw.com/l/6YsJahWEKqP/1NHDlFr6CED

Haven't run "make update" yet, because it's still in process.

cc @zhujian7 
assign @qiujian16

operator/v1/types.go
```diff
@@ -45,6 +45,21 @@ type ClusterManagerSpec struct {
 	// NodePlacement enables explicit control over the scheduling of the deployed pods.
 	// +optional
 	NodePlacement NodePlacement `json:"nodePlacement,omitempty"`
+
+	// HubDeployMode represents the mode used to deploy hub components. There are two mode: mode `Unite` as default, and mode `Separate`.
+	// In Separate mode, a "shell-hub" holds all CRDs, and "worker-hub" has all deployment running in it.
+	// +kubebuilder:default=Unite
+	HubDeployMode string `json:"hubDeployMode"`
```
qiujian16: What is shell-hub? I would name it as `DeployMode`

operator/v1/types.go
```diff
@@ -45,6 +45,21 @@ type ClusterManagerSpec struct {
...
+
+	// HubInstalledNamespace represents the desired namespace used to install all hub components.
+	// +optional
+	// +kubebuilder:default=open-cluster-management-hub
+	HubInstalledNamespace string `json:"hubInstalledNamespace,omitempty"`
```
qiujian16: put this and kubeconfig in a struct. Do not add hub in the field name.

I would think you have things like

```go

type DeployOption {
    DeployMode string // it can be Default or ExternalController. External controller indicates the controller is deployed in another cluster but not the hub cluster.

    ExternalController ExternalControllerConfig
}

type ExternalControllerConfig struct {
    InstallNamespace
}
```

Do we need to specify hubkubeconfig in API, I think it can be a fixe name with fixed key in it.

operator/v1/types.go
xuezhaojun: It represents the cluster installed with CRD. I named them "shell" and "worker" currently but there could be other suitable names.

xuezhaojun: For the DeployMode, are we going to use this one for both cluster-manager deploy and klusterlet deploy?

operator/v1/types.go
xuezhaojun: Sure, it's a better way with a fixed name and fixed key.

xuezhaojun:
/hold

It's weird that run `make update` would affect a lot of files and failed the test. Working on this currently.

operator/v1/types.go
```diff
@@ -45,6 +45,27 @@ type ClusterManagerSpec struct {
...
+	// +optional
+	// +kubebuilder:default=Default
+	DeployMode string `json:"deployMode,omitempty"`
+
+	// ExternalController represents the config when using the ExternalController mode.
+	// +optional
+	ExternalController ExternalControllerConfig `json:"externalController,omitempty"`
+}
+
+type ExternalControllerConfig struct {
```
zhujian7: how can we get the hub kubeconfig with mode `ExternalController`?

operator/v1/types.go
```diff
@@ -45,6 +45,27 @@ type ClusterManagerSpec struct {
...
+	// DeployOptions contains the options of deploying a cluster-manager
+	// +optional
+	DeployOption DeployOption `json:"deployOptions,omitempty"`
+}
+
+type DeployOption struct {
+	// DeployMode can be Default or ExternalController, ExternalController indicates the controller is deployed in another cluster but not the hub cluster.
+	// +optional
+	// +kubebuilder:default=Default
+	DeployMode string `json:"deployMode,omitempty"`
```
qiujian16: I think it should be required

operator/v1/types.go
```diff
@@ -45,6 +45,27 @@ type ClusterManagerSpec struct {
...
+
+	// ExternalController represents the config when using the ExternalController mode.
+	// +optional
```
qiujian16: Specify more details on what is the meaning of ExternalController mode

operator/v1/types.go
qiujian16: it can be hardcoded. But we need to doc here. I do not think this will need user to mutate.

operator/v1/types.go
```diff
@@ -45,6 +45,27 @@ type ClusterManagerSpec struct {
 	// NodePlacement enables explicit control over the scheduling of the deployed pods.
 	// +optional
 	NodePlacement NodePlacement `json:"nodePlacement,omitempty"`
+
+	// DeployOptions contains the options of deploying a cluster-manager
+	// +optional
+	DeployOption DeployOption `json:"deployOptions,omitempty"`
```
qiujian16: deployOption

operator/v1/types.go
qiujian16: deployMode--> mode

operator/v1/types.go
xuezhaojun: From the above discussion https://github.com/open-cluster-management-io/api/pull/103#discussion_r743455095 , we may use a hardcode fixed namespace and secret name to provide hubkubeconfig.

qiujian16:
/assign @zhiweiyin318 

I still have some concern on the document, but need @zhiweiyin318 to take another look.

operator/v1/types.go
zhujian7: Does this mean one registration operator only works for one hub cluster?  If we want to deploy multiple `registration-hub` for multiple hub clusters in one cluster(separate namespaces), we should run multiple registration operators, right?

operator/v1/types.go
```diff
@@ -45,6 +45,23 @@ type ClusterManagerSpec struct {
...
+	// Mode can be Default or Hosted.
+	// In Hosted mode, hub hosting all deployments on another cluster.
+	// The purpose of Hosted mode is to give it more flexibility, for example we can install a hub on a cluster with no worker nodes, meanwhile running all deployments on another more powerful cluster.
+	// The cluster installed with CRDs and services, we still call it hub-cluster.
+	// The cluster running all deployments, we call it hosted-cluster.
+	// In Hosted mode, a ClusterManager is applied to a hosted-cluster, and it leverage a secret named "external-hub-kubeconfig" in the namespace with the same name of itself.
+	// "external-hub-kubeconfig" should contain the kubeconfig with admin permission of hub-cluster.
+	// +required
+	// +kubebuilder:default=Default
+	Mode string `json:"mode,omitempty"`
```
qiujian16: addn kubebuilder required marker. rm omitempty.

operator/v1/types.go
```diff
@@ -45,6 +45,23 @@ type ClusterManagerSpec struct {
...
 	NodePlacement NodePlacement `json:"nodePlacement,omitempty"`
+
+	// DeployOption contains the options of deploying a cluster-manager
+	// +optional
+	DeployOption DeployOption `json:"deployOption,omitempty"`
+}
+
+type DeployOption struct {
+	// Mode can be Default or Hosted.
+	// In Hosted mode, hub hosting all deployments on another cluster.
```
qiujian16: Which ns the deployments are running on with hosted mode?

The following sentence should also be changed
```
// ClusterManager will only be deployed in open-cluster-management-hub namespace.
```

qiujian16: In hosted mode, only crd and configurations are installed on the hub. Controllers are run in another cluster (defined as hosted cluster) and connect to the hub with the kubeconfig in secret of  "external-hub-kubeconfig".

operator/v1/types.go
qiujian16: i would think you want to define the type mode, and also set enum marker for this field

operator/v1/types.go
```diff
@@ -45,6 +46,22 @@ type ClusterManagerSpec struct {
 	// NodePlacement enables explicit control over the scheduling of the deployed pods.
 	// +optional
 	NodePlacement NodePlacement `json:"nodePlacement,omitempty"`
+
+	// DeployOption contains the options of deploying a cluster-manager
+	// +optional
```
qiujian16, line 52: Default mode is used if DeployOption is not set.

operator/v1/types.go
```diff
@@ -45,6 +46,22 @@ type ClusterManagerSpec struct {
...
+type DeployOption struct {
+	// Mode can be Default or Detached.
+	// In Detached mode, only crd and configurations are installed on the hub. Controllers are run in another cluster (defined as management cluster) and connect to the hub with the kubeconfig in secret of "external-hub-kubeconfig".
+	// The purpose of Detached mode is to give it more flexibility, for example we can install a hub on a cluster with no worker nodes, meanwhile running all deployments on another more powerful cluster.
+	// Also "external-hub-kubeconfig" should contain the kubeconfig with admin permission of hub-cluster.
+	// +kubebuilder:validation:Required
+	// +required
+	// +kubebuilder:default=Default
+	// +kubebuilder:validation:Enum=Default;Detached
+	Mode string `json:"mode"`
```
qiujian16:
```go
Mode InstallMode `json:"mode"`

type InstallMode string
```

operator/v1/types.go
xuezhaojun: Add the comments

xuezhaojun: Added.

qiujian16 (commented):
/approve
/hold

@zhujian7 @zhiweiyin318 would need to review this also

operator/v1/types.go
```diff
@@ -45,6 +46,30 @@ type ClusterManagerSpec struct {
...
+const (
+	InstallModeDefault  InstallMode = "Default"
+	InstallModeDetached InstallMode = "Detached"
+)
+
+type DeployOption struct {
+	// Mode can be Default or Detached.
+	// In Detached mode, only crd and configurations are installed on the hub. Controllers are run in another cluster (defined as management cluster) and connect to the hub with the kubeconfig in secret of "external-hub-kubeconfig".
+	// The purpose of Detached mode is to give it more flexibility, for example we can install a hub on a cluster with no worker nodes, meanwhile running all deployments on another more powerful cluster.
+	// Also "external-hub-kubeconfig" should contain the kubeconfig with admin permission of hub-cluster.
```
zhujian7: Does the `external-hub-kubeconfig` still need admin permission?

operator/v1/0000_01_operator.open-cluster-management.io_clustermanagers.crd.yaml
```diff
@@ -30,6 +30,19 @@ spec:
...
               type: object
               properties:
+                deployOption:
+                  description: DeployOption contains the options of deploying a cluster-manager Default mode is used if DeployOption is not set.
+                  type: object
+                  required:
+                    - mode
+                  properties:
+                    mode:
+                      description: Mode can be Default or Detached. In Detached mode, only crd and configurations are installed on the hub. Controllers are run in another cluster (defined as management cluster) and connect to the hub with the kubeconfig in secret of "external-hub-kubeconfig". The purpose of Detached mode is to give it more flexibility, for example we can install a hub on a cluster with no worker nodes, meanwhile running all deployments on another more powerful cluster. Also "external-hub-kubeconfig" should contain the kubeconfig with admin permission of hub-cluster.
```
zhiweiyin318: Controller ~are~ runs ... and connects

zhiweiyin318: do we need to doc Default mode here too?

operator/v1/types.go
xuezhaojun: I suppose yes. With admin permission do no harm since we will use kubeconfig generated by SA syncer.

operator/v1/0000_01_operator.open-cluster-management.io_clustermanagers.crd.yaml
xuezhaojun: Yes, a brief description of the Default mode would be helpful. Let me make a update.

operator/v1/types.go
zhujian7: I think a detailed permission description is more friendly for users. We describe the minimum permission requirement, users can use admin as well.

xuezhaojun: Yes, that makes sense. I added a description to emphasize the kubeconfig should have appropriate permission here. In the registration-operator, I will add a makefile command to generate an external-hub-config with a correct role and rolebinding as same as the default mode: https://github.com/open-cluster-management-io/registration-operator/blob/main/deploy/cluster-manager/config/rbac/cluster_role.yaml

zhujian7: I think neither "appropriate authority" nor "management authority" is a proper description. but the detailed permission needs a lot of description, so I don't know how to describe it here either. @qiujian16 WDYT?

PR #108, opened 2021-12-01
Add interested field

qiujian16:


work/v1/types.go
```diff
@@ -66,6 +70,64 @@ type DeleteOption struct {
...
+	// +kubebuilder:validation:Required
+	// +required
+	Type FeedBackType `json:"type"`
+
+	// JsonPaths defines the json path under status field to be synced.
+	// +optional
+	JsonPaths []JsonPath `json:"jsonPaths,omitempty"`
+}
+
+// +kubebuilder:validation:Enum=wellKnownStatus;jsonPaths
```
yue9944882: conventionally, the enums should all be capitalized in k8s api world, is it?

work/v1/types.go
```diff
@@ -218,11 +280,68 @@ type ManifestCondition struct {
...
+	// +optional
+	String string `json:"string,omitempty"`
+
+	// Boolean is bool value when type is boolean.
+	// +optional
+	Boolean bool `json:"boolean,omitempty"`
+}
+
+// +kubebuilder:validation:Enum=integer;string;boolean
+type ValueType string
```
yue9944882, line 342: ditto ^^^^

work/v1/types.go
```diff
@@ -66,6 +70,64 @@ type DeleteOption struct {
...
+	// returned
+	JsonPathsType FeedBackType = "jsonPaths"
+)
+
+type JsonPath struct {
+	// name represents the aliase name for this field
+	// +kubebuilder:validation:Required
+	// +required
+	Name string `json:"name"`
+	// JsonPaths represents the json path of the field under status.
```
yue9944882: nit: s/JsonPaths/Path/

work/v1/types.go
```diff
@@ -66,6 +70,64 @@ type DeleteOption struct {
...
+	// include replica, readyReplica and availableReplica.
+	WellKnownStatusType FeedBackType = "wellKnownStatus"
+
+	// JsonPathsType represents that values of status fields with certain json paths specified will be
+	// returned
+	JsonPathsType FeedBackType = "jsonPaths"
+)
+
+type JsonPath struct {
+	// name represents the aliase name for this field
```
yue9944882: nit: s/name/Name/

work/v1/types.go
```diff
@@ -66,6 +70,64 @@ type DeleteOption struct {
...
+	// +kubebuilder:validation:Required
+	// +required
+	Name string `json:"name"`
+	// JsonPaths represents the json path of the field under status.
+	// The path must point to a field with single value in the type of integer, bool or string.
+	// If the path points points to a non-existing field, no value will be returned.
```
yue9944882: s/points points/points/

work/v1/types.go
```diff
@@ -218,11 +280,68 @@ type ManifestCondition struct {
...
+// The value of the status field can only be integer, string or boolean.
+type FieldValue struct {
+	// Type represents the type of the value, it can be integer, string or boolean.
+	// +kubebuilder:validation:Required
+	// +required
+	Type ValueType `json:"type"`
+
+	// Integer is the integer value when type is integer.
+	// +optional
+	Integer int32 `json:"integer,omitempty"`
```
yue9944882, line 330: how you think if make these values pointer? so we can leave the value to be null when no values are present in the referencing objects. in kubebuilder marker, it can be done by adding `// +nullable` to the godoc, and the generators will be declaring the field as `nullable: true` in the openapi spec which allows the field to be `null` in yaml. by this way, we can discrinimate the case where the interest field of type string is an empty string `""` or the field doesnt exists (e.g. b/c no parent fields present).

work/v1/types.go
```diff
@@ -218,11 +280,68 @@ type ManifestCondition struct {
...
+type StatusFeedbackResult struct {
+	// Values represents the synced value of the interested field.
+	// +listType:=map
+	// +listMapKey:=name
+	// +optional
+	Values []FeedbackValue `json:"values,omitempty"`
+}
+
+type FeedbackValue struct {
+	// name represents the aliase name for this field. It is the same as what is specified
```
yue9944882: nit: s/name/Name/

work/v1/types.go
yue9944882: when the path points to a non-existing path, a `null` value will be returned instead

work/v1/types.go
```diff
@@ -66,6 +70,64 @@ type DeleteOption struct {
...
+type StatusFeedbackOption struct {
+	// ResourceIdentifier represents the group, resource, name and namespace of a resoure.
+	// +kubebuilder:validation:Required
+	// +required
+	ResourceIdentifier ResourceIdentifier `json:"resourceIdentifier"`
+
+	// FeedbackRules defines what resource status field should be returned.
+	// +kubebuilder:validation:Required
+	// +required
+	FeedbackRules []FeedbackRule `json:"feedbackRule"`
```
yue9944882: s/feedbackRule/feedbackRules/

yue9944882 (commented):
thanks for starting this. overall i have a minor concern about renaming the go struct `OrphaningRule` to `ResourceIdentifier` which will break the compatibility for those who's importing the go api.

work/v1/types.go
qiujian16: updated, I actually change this name to be manifestConfigs. I am thinking we might have more config option based on resource identifier later.

work/v1/types.go
qiujian16: We will have 3 scenarios
1. the fieldValue is not present
2. the value is empty
3. the value is nil

My original thought is scenario 1 indicate there is a mismatch on field, or the value type is not supported. scenario 2 will indicate just empty value.

qiujian16:
> thanks for starting this. overall i have a minor concern about renaming the go struct `OrphaningRule` to `ResourceIdentifier` which will break the compatibility for those who's importing the go api.

good point, I've changed it back and add a new resourceIdentifier struct.

work/v1/types.go
```diff
@@ -66,6 +70,65 @@ type DeleteOption struct {
...
+}
+
+type FeedbackRule struct {
+	// Type defines the option of how status can be returned.
+	// It can be jsonPaths only or wellKnownStatus.
+	// If the type is jsonPaths, user should specify the jsonPaths field
+	// If the type is wellKnownStatus, certain commont fields of status will be reported,
```
deads2k: typo commont

deads2k: WellKnownStatus

work/v1/types.go
```diff
@@ -66,6 +70,65 @@ type DeleteOption struct {
...
+	// FeedbackRules defines what resource status field should be returned.
+	// +kubebuilder:validation:Required
+	// +required
+	FeedbackRules []FeedbackRule `json:"feedbackRule"`
+}
+
+type FeedbackRule struct {
+	// Type defines the option of how status can be returned.
+	// It can be jsonPaths only or wellKnownStatus.
+	// If the type is jsonPaths, user should specify the jsonPaths field
```
deads2k: JsonPaths

work/v1/types.go
```diff
@@ -66,6 +70,65 @@ type DeleteOption struct {
...
+	// include replica, readyReplica and availableReplica.
+	WellKnownStatusType FeedBackType = "WellKnownStatus"
+
+	// JsonPathsType represents that values of status fields with certain json paths specified will be
+	// returned
+	JsonPathsType FeedBackType = "JsonPaths"
+)
+
+type JsonPath struct {
+	// Name represents the aliase name for this field
```
deads2k: alias

work/v1/types.go
```diff
@@ -66,6 +70,65 @@ type DeleteOption struct {
...
+	// If the type is wellKnownStatus, certain commont fields of status will be reported,
+	// including replicas, readyReplicas, and availableReplicas. And if these status fields
+	// do not exist, no values will be reported.
+	// +kubebuilder:validation:Required
+	// +required
+	Type FeedBackType `json:"type"`
+
+	// JsonPaths defines the json path under status field to be synced.
+	// +optional
+	JsonPaths []JsonPath `json:"jsonPaths,omitempty"`
```
deads2k, line 100: Something to consider in API evolution: do you want a way to define WellKnownStatus for particular resources so that someone's custom type can have well known status fields to avoid subtely different names?

work/v1/types.go
```diff
@@ -88,18 +151,46 @@ type SelectivelyOrphan struct {
...
+type ResourceIdentifier struct {
+	// Group is the API Group of the Kubernetes resource.
+	// +kubebuilder:validation:Required
+	// +required
+	Group string `json:"group"`
+
+	// Version is the version of the Kubernetes resource.
+	// +kubebuilder:validation:Required
+	// +required
+	Version string `json:"version"`
```
deads2k: The Version is a property the JSONPath has, right?  The resource itself is identified without a version, but the jsonpath requires the version, but something like WellKnownSTatus would not.

work/v1/types.go
```diff
@@ -138,26 +229,7 @@ type ManifestResourceMeta struct {
 // AppliedManifestResourceMeta represents the group, version, resource, name and namespace of a resource.
 // Since these resources have been created, they must have valid group, version, resource, namespace, and name.
 type AppliedManifestResourceMeta struct {
-	// Group is the API Group of the Kubernetes resource.
-	// +required
-	Group string `json:"group"`
-
-	// Version is the version of the Kubernetes resource.
-	// +required
-	Version string `json:"version"`
```
deads2k, line 147: This has a version because it is used for status reporting.  When a user identifies a resource in spec, there isn't a logical version requirement.

work/v1/types.go
qiujian16: Yes, I think we could have another cluster scoped API to define the list of status field for custom type.

work/v1/types.go
qiujian16: I think it is possible that user put the same resource with different versions in the manifests. E.g. crd with v1&v1beta1. Having version here can help to identify which single manifest this option is used to get status. An example

```yaml
apiVersion: work.open-cluster-management.io/v1
Kind: Manifestwork
metadata:
  name: work-1
  namespace: cluster1
spec:
  workload:
    manifests:
    - {resource Foo with version v1}
    - {resource Foo with version v2}
  manifestConfigs:
  - resourceIdentifier:
      resource: Foo
      version: v1
      group: FooGroup
      name: foo
    feedbackRules:
    - type: WellKnownStatus
```

So we only get status from resource Foo with version v1

yue9944882: i suppose the version field can be just removed from the struct. we can identify a resource completely w/o the version and the reporting api version should be implicitly aligning w/ the original api version from the manifest. i dont think there's will be cases where we need to e.g. distribute the resource in v1 api but reporting in v1beta1.

work/v1/types.go
yue9944882: by "misamatch", do you mean the parent field of the interested fields is not present on the resource? remember that a non-exist field can also be logically a nil field (e.g. if the field is tagged w/ `omitempty` in the go code), i presume practically it's not feasible to discriminate case (1) and (3) from the perspective of the work agent.

work/v1/types.go
```diff
@@ -66,6 +70,65 @@ type DeleteOption struct {
...
+	JsonPaths []JsonPath `json:"jsonPaths,omitempty"`
+}
+
+// +kubebuilder:validation:Enum=WellKnownStatus;JsonPaths
+type FeedBackType string
+
+const (
+	// WellKnownStatusType represents that values of some common status fields will be returned, which
+	// include replica, readyReplica and availableReplica.
+	WellKnownStatusType FeedBackType = "WellKnownStatus"
```
yue9944882, line 109: IIUC, the `WellKnownStatus` is actually a pre-built template of interested fields for reporting. if so, will the reporting fields differ depends on the resource type? e.g. if we apply `WellKnownStatus` to a job resource, then the reporting fields should be `.status.succeeded` etc. while if it's a deployment then `.spec.replicas` will be reported?

work/v1/types.go
yue9944882: > I think it is possible that user put the same resource with different versions in the manifests. E.g. crd with v1&v1beta1

oh i didnt know about such use cases before. i though every element in the manifest list should be a logically unique resource

work/v1/types.go
qiujian16: I think this is similar as the comment here https://github.com/open-cluster-management-io/api/pull/108#discussion_r763066481 ^^^.  I think we should consider a WellKnownStatus per type or make it configurable in the long term. Maybe we start with a small list and if an item in the list cannot be found in status, we just skip it?

work/v1/types.go
qiujian16: Do you mean when we want feedback of `.status.foo.bar` we should discriminate between

```yaml
status:
  foo:
    bar: nil
```
and

```yaml
status:
  foor: {}
```

My original thought is we do not need to discriminate them by just skipping the related result, since none of them returned a meaningful result. Any reason we should discriminate them?

yue9944882: just skipping should be fine in the first stage

work/v1/types.go
```diff
@@ -66,6 +70,65 @@ type DeleteOption struct {
...
+	// +required
+	Type FeedBackType `json:"type"`
+
+	// JsonPaths defines the json path under status field to be synced.
+	// +optional
+	JsonPaths []JsonPath `json:"jsonPaths,omitempty"`
+}
+
+// +kubebuilder:validation:Enum=WellKnownStatus;JsonPaths
+type FeedBackType string
```
yue9944882, line 104: we can have another feedback type like `StatusCondition`, which will be working like:

```
manifestConfigs: 
 - resourceIdentifier: 
     resource: Foo 
     version: v1 
     group: FooGroup 
     name: foo 
   feedbackRules: 
   - type: StatusCondition
     statusCondition:
       conditionType: MyConditionType    
...
resourceStatus:
  manifests:
  - resourceMeta:  ...
    statusFeedback:
      values:
      - name: foo
         fieldValue: 
           type: Condition
           condition: ...    
```

qiujian16: with json path, you can have

```yaml
   - type: StatusJsonPaths
      jsonPaths:
      - name: availableCondition
        path: .conditions[?(@.type=="Available")].status 
```

will it be good enough?

yue9944882: IIRC we dont need a new feedback type `StatusJsonPaths` if it's describe'ble via a json path, is it?

cc @zychina for evaluating the usability.

work/v1/types.go
deads2k: > So we only get status from resource Foo with version v1

if the name, namespace, group, and resource are the same, then the distinction between getting v1 status and v1beta1 status is only on whether the JSONPath resolves right?  For the purposes of "WellKnownStatus", you would choose the semantically latest.

For the purposes of locating a resource in a cluster this doesn't provide a distinction.

work/v1/types.go
deads2k: > Maybe we start with a small list and if an item in the list cannot be found in status, we just skip it?

I'm ok with this. Perhaps with a rule that hardcoded items will only be for types in k8s.io/api and open-cluster-management/api, so we don't get vendor specific bits hardcoded?

qiujian16: Updated the comment for this type

work/v1/types.go
qiujian16: AH, I see. Thanks for the explanation. API updated accordingly.

work/v1/types.go
```diff
@@ -66,6 +70,72 @@ type DeleteOption struct {
 	SelectivelyOrphan *SelectivelyOrphan `json:"selectivelyOrphans,omitempty"`
 }
 
+// ManifestConfigOption represents the configurations of a manifest defined in workload field.
+type ManifestConfigOption struct {
+	// ResourceIdentifier represents the group, resource, name and namespace of a resoure.
+	// +kubebuilder:validation:Required
+	// +required
+	ResourceIdentifier ResourceIdentifier `json:"resourceIdentifier"`
```
deads2k, line 79: remember that in a v2 API, it would be a better choice to combine this with a `Manifest` itself to have spacial locality. We cannot add it now because `Manifest` is an inline rawExtension.  Sins of the past.

work/v1/types.go
```diff
@@ -66,6 +70,72 @@ type DeleteOption struct {
 	SelectivelyOrphan *SelectivelyOrphan `json:"selectivelyOrphans,omitempty"`
 }
 
+// ManifestConfigOption represents the configurations of a manifest defined in workload field.
+type ManifestConfigOption struct {
+	// ResourceIdentifier represents the group, resource, name and namespace of a resoure.
```
deads2k, line 75: Clarify in a comment that if this refers to a thing not created by this manifest work, the result will *not* be shown.

work/v1/types.go
```diff
@@ -66,6 +70,72 @@ type DeleteOption struct {
...
+type FeedBackType string
+
+const (
+	// WellKnownStatusType represents that values of some common status fields will be returned, which
+	// is reflected with a hardcoded rule only for types in k8s.io/api and open-cluster-management/api.
+	WellKnownStatusType FeedBackType = "WellKnownStatus"
+
+	// JsonPathsType represents that values of status fields with certain json paths specified will be
+	// returned
+	JsonPathsType FeedBackType = "JsonPaths"
```
deads2k: kube convention is JSONPaths

work/v1/types.go
```diff
@@ -88,22 +158,32 @@ type SelectivelyOrphan struct {
 	OrphaningRules []OrphaningRule `json:"orphaningRules,omitempty"`
 }
 
-// OrphaningRule identifies a single resource included in this manifestwork
-type OrphaningRule struct {
-	// Group is the api group of the resources in the workload that the strategy is applied
+// ResourceIdentifier identifies a single resource included in this manifestwork
+type ResourceIdentifier struct {
+	// Group is the API Group of the Kubernetes resource.
+	// +kubebuilder:validation:Required
```
deads2k: does this rule work with something like `group=""` for Pod or Service ?

deads2k:
Straight forward comments that should be addressed pre-merge, but the overall shape lgtm

/approve

work/v1/types.go
qiujian16: yes, I will consider this change to work-api in sigs

work/v1/types.go
yue9944882: +1

yue9944882 (commented):
LGTM

PR #111, opened 2021-12-09
set managedcluster leaseduration seconds value with 60

champly:
Signed-off-by: champly <champly@outlook.com>

[discussion this](https://github.com/open-cluster-management-io/registration/pull/184#discussion_r764964632)


champly:
@qiujian16  add managedcluster crd default value, PTAL

PR #113, opened 2021-12-09
add disableLeaseCheck flag to managedclusteraddon

skeeey:


addon/v1alpha1/types_managedclusteraddon.go
```diff
@@ -38,6 +38,16 @@ type ManagedClusterAddOnSpec struct {
...
+
+	// disableLeaseCheck indicates whether the registration agent should check the lease of the
+	// addon on the managed cluster.
+	// By default, the registration agent maintains the addon available status by checking the
+	// lease of the addon on the managed cluster, so the addon agent should maintain a lease in
+	// its installation namespace with its status.
+	// If this field sets to true, the registration agent will not check the lease of the addon on
+	// the managed cluster, the addon should maintain its available status by itself
+	// +optional
+	DisableLeaseCheck bool `json:"disableLeaseCheck,omitempty"`
```
yue9944882: https://github.com/open-cluster-management-io/addon-framework/pull/51

as an alternative of lease-based healthcheck, i think we can have another type of checker which is mapping the status of corresponding work resource to the addon, which can also be done in the registration controller. so it's probably better to generalize a classful healthcheck e.g. including `None`, `Lease`, `Work` b/c it's more extensible. btw i am also sort of worrying if it's a good idea to make the registration controller a centralized healthchecker in the long term, considering that it can be a potential scalability bottleneck. besides, imagine a case where a singleton healthcheck is temporarily down in maintainence mode, will all the addons be down after that?

skeeey: +1, a healthcheck type looks more extensible and more clear than a bool flag

qiujian16: I would want this as a struct with a mode in it.

skeeey: introduce a struct to represent this

addon/v1alpha1/types_managedclusteraddon.go
```diff
@@ -38,6 +38,26 @@ type ManagedClusterAddOnSpec struct {
 	// +kubebuilder:validation:MaxLength=63
 	// +kubebuilder:validation:Pattern=^[a-z0-9]([-a-z0-9]*[a-z0-9])?$
 	InstallNamespace string `json:"installNamespace,omitempty"`
+
+	// healthCheck defines how to check the healthiness status of the current addon.
+	// If not set, the lease mode will be used
+	// +optional
+	HealthCheck HealthCheck `json:"healthCheck,omitempty"`
```
qiujian16: I think this should be in the status field. User should not set this.

skeeey: so the process will be the addon-framework will provide different implementations, then addon-framework will update the health check mode for addon

addon/v1alpha1/types_managedclusteraddon.go
```diff
@@ -38,6 +38,26 @@ type ManagedClusterAddOnSpec struct {
...
+}
+
+type HealthCheckMode string
+
+const (
+	// HealthCheckModeLease, the addon should maintain a lease in its installation namespace with
+	// its status, the registration agent will check this lease to maintain the addon healthiness
+	// status
+	HealthCheckModeLease HealthCheckMode = "lease"
```
qiujian16: lease-->Lease

Maybe also another mode named as “Customized” or "None"

skeeey: add `Customized ` mode

addon/v1alpha1/types_managedclusteraddon.go
skeeey: move to status field

Makefile
```diff
@@ -46,8 +46,8 @@ update-scripts:
 .PHONY: update-scripts
 update: update-scripts update-codegen-crds
 
-update-with-container: update-scripts update-codegen-crds
-
-generate-with-container: Dockerfile.build
+build-runtime-image: Dockerfile.build
 	$(RUNTIME) build -t $(RUNTIME_IMAGE_NAME) -f Dockerfile.build .
-	$(RUNTIME) run -ti --rm -v $(PWD):/go/src/open-cluster-management.io/api:z -w /go/src/open-cluster-management.io/api $(RUNTIME_IMAGE_NAME) make update-with-container
+
```
qiujian16: why need to change this file?

skeeey: rebase probelm, it's ok now :)

addon/v1alpha1/types_managedclusteraddon.go
```diff
@@ -142,6 +147,24 @@ type ObjectReference struct {
...
+	HealthCheckModeLease HealthCheckMode = "Lease"
+
+	// HealthCheckModeCustomized, the addon maintains its healthiness status by itself.
+	HealthCheckModeCustomized HealthCheckMode = "Customized"
+)
+
+type HealthCheck struct {
+	// mode indicates which mode will be used to check the healthiness status of the addon.
+	// +kubebuilder:default=Lease
+	Mode HealthCheckMode `json:"mode"`
```
qiujian16, line 168: +optional?

skeeey: yeah, added

addon/v1alpha1/types_managedclusteraddon.go
```diff
@@ -142,6 +147,25 @@ type ObjectReference struct {
 	Name string `json:"name"`
 }
 
+type HealthCheckMode string
```
yue9944882, line 152: nit: enumerize

skeeey: yeah, add the enum comment

qiujian16: I think you need to add kubebuilder's enum marker

skeeey: use kubebuilder marker

PR #115, opened 2021-12-10
support taint & toleration

elgnay:

Feature request: https://github.com/open-cluster-management-io/community/issues/48
Enhancement proposal: https://github.com/open-cluster-management-io/enhancements/tree/main/enhancements/sig-architecture/20-taint-toleration


cluster/v1/types.go
```diff
@@ -82,6 +86,43 @@ type ClientConfig struct {
...
+// any placement that does not tolerate the Taint.
+type Taint struct {
+	// The taint key to be applied to a cluster. e.g. bar or foo.example.com/bar.
+	// The regex it matches is (dns1123SubdomainFmt/)?(qualifiedNameFmt)
+	// +required
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:Pattern=`^([a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*/)?(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])$`
+	// +kubebuilder:validation:MaxLength=316
+	Key string `json:"key"`
+	// The taint value corresponding to the taint key.
```
qiujian16: It commonly start with "Value defines ..."

cluster/v1alpha1/types.go
```diff
@@ -321,6 +327,46 @@ type PrioritizerConfig struct {
...
 }
 
+// Toleration represents the toleration object that can be attached to a placement.
+// The placement this Toleration is attached to tolerates any taint that matches
+// the triple <key,value,effect> using the matching operator <operator>.
+type Toleration struct {
+	// Key is the taint key that the toleration applies to. Empty means match all taint keys.
+	// If the key is empty, operator must be Exists; this combination means to match all values and all keys.
+	// +optional
+	Key string `json:"key,omitempty"`
```
qiujian16, line 386: the regex format here should be the same as taint? Why this is optional?

cluster/v1alpha1/types.go
```diff
@@ -321,6 +327,46 @@ type PrioritizerConfig struct {
...
+	// If the operator is Exists, the value should be empty, otherwise just a regular string.
+	// +optional
+	Value string `json:"value,omitempty"`
+	// Effect indicates the taint effect to match. Empty means match all taint effects.
+	// When specified, allowed values are NoSelect, PreferNoSelect and NoSelectIfNew.
+	// +optional
+	Effect v1.TaintEffect `json:"effect,omitempty"`
+	// TolerationSeconds represents the period of time the toleration (which must only be effective under the Evict Senario,
+	// otherwise this field should be ignored) tolerates the taint. By default,
+	// It is not set, which means tolerate the taint forever (do not evict).
```
qiujian16: If it is nil, it indicates to tolerate the taint forever.

cluster/v1alpha1/types.go
```diff
@@ -321,6 +327,46 @@ type PrioritizerConfig struct {
...
+type Toleration struct {
+	// Key is the taint key that the toleration applies to. Empty means match all taint keys.
+	// If the key is empty, operator must be Exists; this combination means to match all values and all keys.
+	// +optional
+	Key string `json:"key,omitempty"`
+	// Operator represents a key's relationship to the value.
+	// Valid operators are Exists and Equal. Defaults to Equal.
+	// Exists is equivalent to wildcard for value, so that a placement can
+	// tolerate all taints of a particular category.
+	// +optional
```
qiujian16, line 392: add default marker?

cluster/v1alpha1/types.go
```diff
@@ -321,6 +327,46 @@ type PrioritizerConfig struct {
...
+	Operator TolerationOperator `json:"operator,omitempty"`
+	// Value is the taint value the toleration matches to.
+	// If the operator is Exists, the value should be empty, otherwise just a regular string.
+	// +optional
```
qiujian16, line 397: maxlength marker

cluster/v1alpha1/types.go
```diff
@@ -321,6 +327,46 @@ type PrioritizerConfig struct {
...
+	Value string `json:"value,omitempty"`
+	// Effect indicates the taint effect to match. Empty means match all taint effects.
+	// When specified, allowed values are NoSelect, PreferNoSelect and NoSelectIfNew.
+	// +optional
```
qiujian16, line 402: enum marker

cluster/v1/types.go
```diff
@@ -82,6 +86,43 @@ type ClientConfig struct {
...
+	// +optional
+	// +kubebuilder:validation:MaxLength=1024
+	Value string `json:"value,omitempty"`
+	// Required. The effect of the taint on placements that do not tolerate the taint.
+	// Valid effects are NoSelect, PreferNoSelect and NoSelectIfNew.
+	Effect TaintEffect `json:"effect"`
```
qiujian16, line 110: enum marker

cluster/v1/types.go
```diff
@@ -82,6 +86,43 @@ type ClientConfig struct {
...
+	// TimeAdded represents the time at which the taint was added.
+	TimeAdded metav1.Time `json:"timeAdded"`
```
qiujian16, line 114: should this be pointer?

cluster/v1/types.go
elgnay: Updated.

cluster/v1alpha1/types.go
elgnay: > the regex format here should be the same as taint?

Yes.

> Why this is optional?

If key=empty and operator=Exists, it means to match all values and all keys. In another word, the placement tolerates all taints.

cluster/v1alpha1/types.go
elgnay: Updated.

cluster/v1alpha1/types.go
elgnay: Added.

cluster/v1alpha1/types.go
elgnay: Added.

cluster/v1alpha1/types.go
elgnay: Added

cluster/v1/types.go
elgnay: Added

cluster/v1/types.go
elgnay: Since we always know the time the taint was added, David suggested we don't use a point for this field. See more details here:
https://github.com/open-cluster-management-io/enhancements/pull/22#discussion_r745047189

qiujian16: add optional?

cluster/v1/types.go
qiujian16: add // +kubebuilder:validation:Required

cluster/v1/types.go
elgnay: If we make it optional, we may need a mutating webhook to fill this field automatically.

cluster/v1/types.go
elgnay: Done.

cluster/v1/types.go
qiujian16: I think you will need it anyway...

cluster/v1/types.go
```diff
@@ -82,6 +86,46 @@ type ClientConfig struct {
...
+	// +kubebuilder:validation:Pattern=`^([a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*/)?(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])$`
+	// +kubebuilder:validation:MaxLength=316
+	// +required
+	Key string `json:"key"`
+	// Value is the taint value corresponding to the taint key.
+	// +kubebuilder:validation:MaxLength=1024
+	// +optional
+	Value string `json:"value,omitempty"`
+	// Effect indicates the effect of the taint on placements that do not tolerate the taint.
+	// Valid effects are NoSelect, PreferNoSelect and NoSelectIfNew.
```
deads2k, line 106: Expand on the meaning of these three values please.

EDIT: found below

cluster/v1/types.go
deads2k: > should this be pointer?

Why should this be a pointer or allow it to be optional if we always know the value?

cluster/v1alpha1/types.go
```diff
@@ -321,6 +327,51 @@ type PrioritizerConfig struct {
...
+	// If the operator is Exists, the value should be empty, otherwise just a regular string.
+	// +kubebuilder:validation:MaxLength=1024
+	// +optional
+	Value string `json:"value,omitempty"`
+	// Effect indicates the taint effect to match. Empty means match all taint effects.
+	// When specified, allowed values are NoSelect, PreferNoSelect and NoSelectIfNew.
+	// +kubebuilder:validation:Enum:=NoSelect;PreferNoSelect;NoSelectIfNew
+	// +optional
+	Effect v1.TaintEffect `json:"effect,omitempty"`
+	// TolerationSeconds represents the period of time the toleration (which must only be effective under the Evict Senario,
```
deads2k: Evict Scenario is a pod thing, right?

elgnay: Similar to pod scheduling, a placement might be evicted from a cluster (in another word, remove the cluster from the placement decision) if the time the placement tolerates the taint on this cluster is longer than `TolerationSeconds` in the placement.

If specified, `TolerationSeconds` will be ignored when the toleration effect is `NoSelectIfNew`. In that case, all selected clusters will be kept in the placement decision.

cluster/v1/types.go
elgnay: Users don't have to input `TimeAdded`, and the mutating webhook will set its value. So it's still required in the spec, but optional for user input.

cluster/v1/types.go
```diff
@@ -67,6 +67,10 @@ type ManagedClusterSpec struct {
 	// +optional
 	// +kubebuilder:default=60
 	LeaseDurationSeconds int32 `json:"leaseDurationSeconds,omitempty"`
+
+	// Taints is a property of managed cluster that allow the cluster to be repelled when scheduling.
```
deads2k, line 71: Describe which actors you are expecting to have rights to remove a taint.  Does an agent running in the spoke cluster have the power to remove a taint?  Any taint?

Nodes are forbidden to modify taints to avoid steering workloads based on compromised node credentials (see https://github.com/kubernetes/kubernetes/blob/master/plugin/pkg/admission/noderestriction/admission.go#L428-L432)

I'm open to reconsidering this while starting from scratch by asserting that Taints are only for anti-selection and therefore could be set by any agent, but we should explicitly state that here.

deads2k:
/approve
/hold

holding for the doc clarification I requested

/assign @qiujian16 

for final review

cluster/v1/types.go
elgnay: Doc updated for the clarification.

Issue #112, opened 2021-12-09
Separate API to *_types.go files


qiujian16:
multiple apis are put in one single types.go file which make it hard to read. It is better that we can separate them to multiple files.

PR #122, opened 2021-12-20
fixed issue #112 - split types.go file

ilan-pinto:


qiujian16 (commented):
i think it should be managedclusterset_types.go and managedclustersetbinding_types.go

ilan-pinto:
it always should start with types_XXXX.go ill rename it

PR #125, opened 2021-12-23
update validation rule of taint.timeAdded

elgnay:

### DIFF
--- a/cluster/v1/types.go
+++ b/cluster/v1/types.go
@@ -109,7 +109,6 @@ type Taint struct {
 	// +required
 	Effect TaintEffect `json:"effect"`
 	// TimeAdded represents the time at which the taint was added.
-	// +kubebuilder:validation:Required
 	// +required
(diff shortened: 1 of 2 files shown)


cluster/v1/types.go
```diff
@@ -109,7 +109,6 @@ type Taint struct {
 	// +required
 	Effect TaintEffect `json:"effect"`
 	// TimeAdded represents the time at which the taint was added.
-	// +kubebuilder:validation:Required
 	// +required
```
qiujian16, line 112: also change this +required?

elgnay: This field is still required. It should be maintained by webhook. If we add `// +kubebuilder:validation:Required `, the validation will be done before the create/update request of managedcluster reaches the webhook. That means user can not add a taint without specifying timeAdded.

Anyway, there is a risk that this field is not set properly because the webhook is not running for some reason.

qiujian16 (commented):
/approve
/lgtm

PR #128, opened 2022-01-06
Change integer to int64

qiujian16:


work/v1/types.go
```diff
@@ -327,7 +327,7 @@ type FieldValue struct {
 
 	// Integer is the integer value when type is integer.
 	// +optional
-	Integer int32 `json:"integer,omitempty"`
```
yue9944882, line 330: @qiujian16 i suppose the `omitempty` tag for the integer, string, boolean type is questionable. e.g. what if a `0` or `false` is expected in the feedback value? they will be omitted in the yaml content due to the tag, which can be sometimes confusing to the readers, is it? how do you think about refactoring them to pointer types?

yue9944882 (commented):
/lgtm

PR #135, opened 2022-01-21
Add registraiton & work webhook configration for detached mode

xuezhaojun:
Signed-off-by: xuezhaojun <zxue@redhat.com>

Relevant issue: open-cluster-management-io/registration-operator#157


operator/v1/types.go
```diff
@@ -52,6 +52,18 @@ type ClusterManagerSpec struct {
 	// +optional
 	// +kubebuilder:default={mode: Default}
 	DeployOption DeployOption `json:"deployOption,omitempty"`
+
+	// RegistrationAPIServiceTargetPort is used in rendering the file "manifests/cluster-manager/hub/cluster-manager-registration-webhook-apiservice.yaml".
```
qiujian16: Put this in a struct

```
type Detached struct {
    RegistrationWebhookAPIService {
        Port int32 `json:"registrationAPIServiceTargetPort"`
        Address string
        CABundle []byte
    }
    ....
}
```

operator/v1/types.go
```diff
@@ -52,6 +52,18 @@ type ClusterManagerSpec struct {
...
+	//
+	// +optional
+	// +kubebuilder:default=443
+	RegistrationAPIServiceTargetPort int `json:"registrationAPIServiceTargetPort"`
+
+	// WorkAPISerivceTargetPort is used in redering the file "manifests/cluster-manager/hub/cluster-manager-work-webhook-apiservice.yaml".
+	//
+	// +optional
+	// +kubebuilder:default=443
+	WorkAPISerivceTargetPort int `json:"workAPIServiceTargetPort"`
```
qiujian16: should be int32

xuezhaojun: Fixed

operator/v1/types.go
xuezhaojun: Added new struct, fixed.

operator/v1/types.go
```diff
@@ -52,6 +52,33 @@ type ClusterManagerSpec struct {
 	// +optional
 	// +kubebuilder:default={mode: Default}
 	DeployOption DeployOption `json:"deployOption,omitempty"`
+
+	// Detached include configurations we needs in the detached mode.
+	// +optional
+	Detached Detached `json:"detached"`
```
qiujian16: add omitempty

operator/v1/types.go
```diff
@@ -52,6 +52,33 @@ type ClusterManagerSpec struct {
...
+}
+
+// Detached represents customized configurations we need to set in the detached mode.
+type Detached struct {
```
qiujian16: Maybe DetachedClusterManagerConfiguration for the struct name, you cannot ensure we do not have klusterlet configuration later.

operator/v1/types.go
```diff
@@ -52,6 +52,33 @@ type ClusterManagerSpec struct {
...
+	// RegistrationAPIServiceTargetPort is used in rendering the file "manifests/cluster-manager/hub/cluster-manager-registration-webhook-apiservice.yaml".
```
qiujian16: this comment is not correct

operator/v1/types.go
```diff
@@ -52,6 +52,33 @@ type ClusterManagerSpec struct {
...
+
+	// WorkAPIWebhookAPISerivce is used in redering the file "manifests/cluster-manager/hub/cluster-manager-work-webhook-apiservice.yaml".
+	//
+	// +optional
+	WorkWebhookAPIService APIService `json:"workWebhookAPIService"`
+}
+
+// APIService represents customized configurations we need to set for APIService.
+type APIService struct {
+	// ServicePort represents the `Spec.Service.Port` in APIService.
```
qiujian16: ServicePort is the port of the service that serving the apiservice

operator/v1/types.go
```diff
@@ -52,6 +52,33 @@ type ClusterManagerSpec struct {
...
+	// If this field is not set or equals to 0, we will use the default value: 443
+	//
+	// +optional
+	// +kubebuilder:default=443
+	ServicePort int32 `json:"servicePort"`
```
qiujian16: why only port, I think you would still need address and cabundle

operator/v1/types.go
xuezhaojun: Fixed.

operator/v1/types.go
xuezhaojun: Sure. it's better to have this clearification.

operator/v1/types.go
xuezhaojun: Yes, it should be added.

operator/v1/types.go
xuezhaojun: Updated.

operator/v1/types.go
xuezhaojun: For the ca-bundle, based on the current implementation, this filed is using the `ca-bundle-configmap` generated by certrotation-controller.

For the address, the strategy is as the following:

This is the manifest file we need to render:
```
apiVersion: apiregistration.k8s.io/v1
kind: APIService
metadata:
  name: v1.admission.cluster.open-cluster-management.io
spec:
  group: admission.cluster.open-cluster-management.io
  version: v1
  service:
    name: cluster-manager-registration-webhook
    namespace: {{ .ClusterManagerNamespace }}
    port: {{.RegistrationServicePort}} # This is what we change
  caBundle: {{ .RegistrationAPIServiceCABundle }}
  groupPriorityMinimum: 10000
  versionPriority: 20
```
We only change the `port` field and apiservice always point to the service -- `cluster-manager. cluster-manager-registration-webhook`.
However, in the detached mode, this service needs users to create by themselves. They can create whatever they need (ExternalName, InCluster, LoadBalancer, and so on) just with the fixed service name and namespace.

qiujian16: If a user need to set such service, why do we still need to set the port. The user can configure the port to what it needs, right?

xuezhaojun: It's true, and it's my first thinking too, but as my experience of tests, it's not handy. 

I tested last night as the following.

image

And my configuration of Ingress would be:

```yaml
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  name: cluster-manager-webhook-ingress
  namespace: cluster-manager
spec:
  rules:
  - http:
      paths:
      - pathType: Prefix
        path: "/apis/admission.cluster.open-cluster-management.io/v1"
        backend:
          service:
            name: cluster-manager-registration-webhook
            port:
              number: 443
      - pathType: Prefix
        path: "/apis/admission.work.open-cluster-management.io/v1"
        backend:
          service:
            name: cluster-manager-work-webhook
            port:
              number: 443
```

I see the logs of the ingress-controller show the traffic is directed to the right pod, but the response is 400.

Then I guess the ingress-controller is not transparent enough, and then I tried many different nginx parameters but all failed. 😭

After hours of testing, finally, I decided to make service ports of APISerivce configurable to leverage NodePort.  In another word, sacrificing simplicity and consistency to gain flexibility.

operator/v1/types.go
```diff
@@ -52,6 +52,45 @@ type ClusterManagerSpec struct {
...
 	// +kubebuilder:default={mode: Default}
 	DeployOption DeployOption `json:"deployOption,omitempty"`
+
+	// Detached includes configurations we needs for clustermanager in the detached mode.
+	// +optional
+	Detached DetachedClusterManagerConfiguration `json:"detached,omitempty"`
+}
+
+// DetachedClusterManagerConfiguration represents customized configurations we need to set for clustermanager in the detached mode.
+type DetachedClusterManagerConfiguration struct {
```
xuezhaojun, line 62:
```go
type DetachedClusterManagerConfiguration struct {
     RegistrationWebhook WebhookConfiguration
     WorkWebhook WebhookConfiguration
}

type WebhookConfiguration struct {
  Address string 
  Port int32
}
```

operator/v1/types.go
```diff
@@ -52,6 +52,43 @@ type ClusterManagerSpec struct {
...
+	// +optional
+	RegistrationWebhookConfiguration WebhookConfiguration `json:"registrationWebhookConfiguration,omitempty"`
+
+	// WorkWebhookConfiguration represents the customized webhook-server configuration of work.
+	// +optional
+	WorkWebhookConfiguration WebhookConfiguration `json:"workWebhookConfiguration,omitempty"`
+}
+
+// WebhookConfiguration has two properties: Address and Port.
+// The registration and work provide webhook-server by using [generic-admission-server](https://github.com/openshift/generic-admission-server).
```
qiujian16: I do not think this doc needs to be added. These are internal implementation

operator/v1/types.go
```diff
@@ -52,6 +52,43 @@ type ClusterManagerSpec struct {
...
+// And this lib has simplified the authentication part of a webhook-server, and the only thing we need to do is set up a Service for the respective APIService automatically.
+type WebhookConfiguration struct {
+	// Address represents the address of a webhook-server. It could be in IP format or fqdn format.
+	// If Address is in IP format, operator will generate a headless Service pointing to the IP.
```
qiujian16: It does not have to be doced whether headless or externalName is used here. The only thing you may need to callout is the the Address must be reachable by apiserver of the hub cluster.

operator/v1/types.go
```diff
@@ -52,6 +52,43 @@ type ClusterManagerSpec struct {
...
 	DeployOption DeployOption `json:"deployOption,omitempty"`
+
+	// Detached includes configurations we needs for clustermanager in the detached mode.
+	// +optional
+	Detached DetachedClusterManagerConfiguration `json:"detached,omitempty"`
+}
+
+// DetachedClusterManagerConfiguration represents customized configurations we need to set for clustermanager in the detached mode.
+// After a detached mode deployment, the webhook-servers will not be feasible until the user exposes the webhook-servers in the management cluster manually. And the way to do that is depending on the platform and user's preference.
+// For example, if we are using a OCP platform, we could use Route to expose the webhook-servers.
```
qiujian16: We do not need to mention ocp here.

operator/v1/types.go
xuezhaojun: Yes, removed this part.

operator/v1/types.go
xuezhaojun: Fixed.

operator/v1/types.go
xuezhaojun: Fixed.

qiujian16 (commented):
/approve
/lgtm

PR #139, opened 2022-02-09
Add version in operator status

qiujian16:


yue9944882:
@qiujian16 is this for the print column? i guess the new `clusteradm get hub-info` command is sufficient for printing the version?

qiujian16:
also for the upgrade case, that we can know the previous version.

operator/v1/types.go
```diff
@@ -127,6 +127,9 @@ type ClusterManagerStatus struct {
 	// +optional
 	ObservedGeneration int64 `json:"observedGeneration,omitempty"`
 
+	// Version represents the version of the compoenents in cluster manager
+	Version ClusterManagerVersion `json:"version,omitempty"`
```
yue9944882, line 131: either make this a pointer or remove `omitempty`?

operator/v1/types.go
```diff
@@ -297,12 +315,27 @@ type NodePlacement struct {
...
+}
+
 // KlusterletStatus represents the current status of Klusterlet agent.
 type KlusterletStatus struct {
 	// ObservedGeneration is the last generation change you've dealt with
 	// +optional
 	ObservedGeneration int64 `json:"observedGeneration,omitempty"`
 
+	// Version represents the version of the compoenents in klusterlet
+	Version KlusterletVersion `json:"version,omitempty"`
```
yue9944882, line 337: ditto

yue9944882:
> also for the upgrade case, that we can know the previous version.

@qiujian16 will this work for vendor customized version? e.g. conventionally we extend a platform version suffix after the semver `vX.Y.Z.vendor.1`

qiujian16:
I think we would read version in version.go. So yes if the customized version is added in ldflag.

PR #140, opened 2022-02-16
Addon certificate tracking and more addon status condition types

yue9944882:
1. syncing a few existing condition types from the addon-framework, including `RegistrationApplied` and `ManifestApplied`
2. extending new fields in the addon status to track the validtiy of the certificate


addon/v1alpha1/types_managedclusteraddon.go
```diff
@@ -58,6 +58,18 @@ type RegistrationConfig struct {
 	//
 	// +optional
 	Subject Subject `json:"subject,omitempty"`
+
+	// certificateStatus actively tracks the status of the certificate used by the addon.
+	CertificateStatus *RegistrationConfigCertificateStatus `json:"certificateStatus,omitempty"`
```
qiujian16, line 63: this field is only used for configuration of registration. I do not think we want to expose this field to addon when setting the configuration. Maybe we should move this out with the name of `RegistrationStatuses`?

yue9944882: if we're moving this out, we will need an identifier field for each certificate status struct to correlate the status and registration config. i cant find such an identifier to use, otherwise we will need to duplicate the whole config. wdyt?

yue9944882:
/hold

backoff api-changes until the next release

PR #141, opened 2022-02-22
Clusterset api change

ldpliu:
https://github.com/stolostron/backlog/issues/20181


cluster/v1beta1/types_managedclusterset.go
```diff
@@ -36,8 +36,37 @@ type ManagedClusterSet struct {
 	Status ManagedClusterSetStatus `json:"status,omitempty"`
 }
 
-// ManagedClusterSetSpec describes the attributes of the ManagedClusterSet
 type ManagedClusterSetSpec struct {
+	// ClusterSelector represents a selector of ManagedClusters
+	ClusterSelector ManagedClusterSelector `json:"clusterSelector"`
```
qiujian16: optional and omitempty

cluster/v1beta1/types_managedclusterset.go
```diff
@@ -36,8 +36,37 @@ type ManagedClusterSet struct {
...
+}
+
+// ManagedClusterSelector represents a selector of ManagedClusters
+type ManagedClusterSelector struct {
+	// SelectorType could be "" or ClusterLabel, will support more SelectorType later
+	// "" means to use the current mechanism of matching label <cluster.open-cluster-management.io/clusterset:<ManagedClusterSet Name>.
+	// "ClusterLabel" means to use a particular cluster label. It is guaranteed that clustersets with same label key are exclusive with each others
+	// +optional
+	SelectorType SelectorType `json:"selectorType"`
```
qiujian16, line 53: add enum marker and default

cluster/v1beta1/types_managedclusterset.go
```diff
@@ -36,8 +36,37 @@ type ManagedClusterSet struct {
...
+}
+
+type SelectorType string
+
+const (
+	ClusterLabel SelectorType = "ClusterLabel"
+)
+
+//Label defines one label
+type Label struct {
```
qiujian16: I would call it ClusterLabel

cluster/v1beta1/types_managedclusterset.go
```diff
@@ -36,8 +36,37 @@ type ManagedClusterSet struct {
...
-// ManagedClusterSetSpec describes the attributes of the ManagedClusterSet
 type ManagedClusterSetSpec struct {
+	// ClusterSelector represents a selector of ManagedClusters
+	ClusterSelector ManagedClusterSelector `json:"clusterSelector"`
+}
+
+// ManagedClusterSelector represents a selector of ManagedClusters
+type ManagedClusterSelector struct {
+	// SelectorType could be "" or ClusterLabel, will support more SelectorType later
+	// "" means to use the current mechanism of matching label <cluster.open-cluster-management.io/clusterset:<ManagedClusterSet Name>.
```
qiujian16: I would not suggest you have "" as an option

cluster/v1beta1/types_managedclusterset.go
```diff
@@ -36,8 +36,37 @@ type ManagedClusterSet struct {
...
+	// "ClusterLabel" means to use a particular cluster label. It is guaranteed that clustersets with same label key are exclusive with each others
```
qiujian16: It is guaranteed that clustersets with same label key are exclusive with each others -->
It guarantees that clustersets with same label key are exclusive

cluster/v1beta1/types_managedclusterset.go
```diff
@@ -36,8 +36,37 @@ type ManagedClusterSet struct {
...
+	// +optional
+	SelectorType SelectorType `json:"selectorType"`
+
+	// ClusterLabel defines one label which clusterset could use to select target managedClusters. In this way, we will:
+	// 1. Guarantee clustersets with same label key are exclusive
+	// 2. Enable additional permission check when cluster joining/leaving a clusterset
```
qiujian16: this is not clear, what permission check?

cluster/v1beta1/types_managedclusterset.go
```diff
@@ -36,8 +36,37 @@ type ManagedClusterSet struct {
...
+	ClusterLabel Label `json:"clusterLabel"`
```
qiujian16: this should be optional and omitEmpty. I think it should be a pointer also.

cluster/v1beta1/types_managedclusterset.go
ldpliu: done

cluster/v1beta1/types_managedclusterset.go
ldpliu: done

cluster/v1beta1/types_managedclusterset.go
ldpliu: As the `Clusterlabel` is conflict with `Clusterlabel` in slectorType. So I changed to `ManagedClusterLabel`

cluster/v1beta1/types_managedclusterset.go
ldpliu: done

cluster/v1beta1/types_managedclusterset.go
ldpliu: done

cluster/v1beta1/types_managedclusterset.go
ldpliu: done

cluster/v1beta1/types_managedclusterset.go
ldpliu: done

cluster/v1beta1/types_managedclusterset.go
```diff
@@ -36,8 +36,42 @@ type ManagedClusterSet struct {
...
+
+const (
+	ClusterLabel SelectorType = "ClusterLabel"
+)
+
+//ManagedClusterLabel defines one label of managedCluster
+type ManagedClusterLabel struct {
+	//The key must be cluster.open-cluster-management.io/clusterset, it will be set by default if unspecified
+	Key string `json:"key"`
+	// The value must be <clusterSet name>, it will be set by default if unspecified)
```
qiujian16: remove the )

ldpliu: done

cluster/v1beta1/types_managedclusterset.go
```diff
@@ -36,8 +36,42 @@ type ManagedClusterSet struct {
 	Status ManagedClusterSetStatus `json:"status,omitempty"`
 }
 
-// ManagedClusterSetSpec describes the attributes of the ManagedClusterSet
```
qiujian16, line 39: why remove this line?

ldpliu: done

cluster/v1beta1/types_managedclusterset.go
```diff
@@ -38,6 +38,41 @@ type ManagedClusterSet struct {
...
+	// he/she should have "create" permission for subresource "managedclusters/label" with name "cluster.open-cluster-management.io/clusterset:<ManagedClusterSet Name>".
+	// +optional
+	ExclusiveLabel *ManagedClusterLabel `json:"exclusiveLabel,omitempty"`
+}
+
+type SelectorType string
+
+const (
+	ExclusiveLabel SelectorType = "ExclusiveLabel"
+)
```
ldpliu, line 60: Maybe, we could use another `word` to replace the `ExclusiveLabel`.
There are three options: `ClusterLabel`, `SingleLabel`, `ExclusiveLabel`.

`ClusterLabel` or `SingleLabel` will be easily understand for user, but could not contain the Exclusive meaning.
`ExclusiveLabel` can not be easily understood by user. and it's only exclusive when the Label Key is same for different sets.

I am not sure which one is better, or anyone has other options?

cluster/v1beta1/types_managedclusterset.go
```diff
@@ -38,6 +38,41 @@ type ManagedClusterSet struct {
...
+
+//ManagedClusterLabel defines one label of managedCluster
+type ManagedClusterLabel struct {
```
deads2k: Thinking about the API, I like this way of describing it, **but** this would require a new API version and a conversionwebhook set it to the .metadata.name value.  That's a lot of overhead.

If we were willing to limit the configuration available, we could instead have a `LegacyClusterSetLabel` value in the discriminated union with no additional configuration.  The `LegacyClusterSetLabel` would have the matching behave the way it functions today.  This would allow us to use the built-in defaulting inside the same API version without any additional webhook logic.

qiujian16 (commented):
/approve

cluster/v1beta1/types_managedclusterset.go
```diff
@@ -38,8 +38,27 @@ type ManagedClusterSet struct {
...
+	ClusterSelector ManagedClusterSelector `json:"clusterSelector,omitempty"`
+}
+
+// ManagedClusterSelector represents a selector of ManagedClusters
+type ManagedClusterSelector struct {
+	// SelectorType could only be "LegacyClusterSetLabel" now, will support more SelectorType later
+	// "LegacyClusterSetLabel" means to use label "cluster.open-cluster-management.io/clusterset:<ManagedClusterSet Name>"" to select target clusters.
+	// +kubebuilder:validation:Enum=LegacyClusterSetLabel
+	// +kubebuilder:default:="LegacyClusterSetLabel"
+	// +optional
```
elgnay: Remove `+optional`?

PR #142, opened 2022-02-23
Fix wrong deploy option for klusterlet

qiujian16:


operator/v1/types.go
```diff
@@ -243,13 +249,11 @@ type KlusterletSpec struct {
 	Namespace string `json:"namespace,omitempty"`
 
 	// RegistrationImagePullSpec represents the desired image configuration of registration agent.
-	// +required
-	// +kubebuilder:default=quay.io/open-cluster-management/registration
-	RegistrationImagePullSpec string `json:"registrationImagePullSpec"`
+	// +optional
```
zhujian7, line 253: Even if we removed the `default` of the `RegistrationImagePullSpec` and `WorkImagePullSpec`, should they remain `required`?

qiujian16: it is for backward compatible, I do not think we require this in the previous version

qiujian16: will add a comment on what image is used if unspecified.

zhujian7:
LGTM

operator/v1/0001_00_operator.open-cluster-management.io_klusterlets.crd.yaml
```diff
@@ -38,52 +38,10 @@ spec:
               type: object
               default:
```
zhujian7: Still has a `default` value here

operator/v1/types.go
```diff
@@ -269,7 +275,7 @@ type KlusterletSpec struct {
 	// DeployOption contains the options of deploying a klusterlet
 	// +optional
 	// +kubebuilder:default={mode: Default}
```
zhujian7: Should delete this line too.

zhujian7 (changes requested):

PR #144, opened 2022-03-02
set default for setapi and clustermanager and klusterlet

ldpliu:
Signed-off-by: ldpliu <daliu@redhat.com>


qiujian16 (commented):
/approve
/lgtm

PR #149, opened 2022-04-01
support Registration FeatureGates to pass feature gates(e.g. V1beta1CSRAPICompatibility) to registration

ivan-cai:
add HubRegistrationFeatureGates to ClusterManager Spec, and SpokeRegistrationFeatureGates to Klusterlet Spec.

input is like：
```
  hubRegistrationFeatureGates:
  - V1beta1CSRAPICompatibility=true
  ```

```
  spokeRegistrationFeatureGates:
  - AddonManagement=true
  - V1beta1CSRAPICompatibility=true
  ``` 

And then, registration-operator will  convert them to `--feature-gates=AddonManagement=true`,`--feature-gates=V1beta1CSRAPICompatibility=true`, add pass to registration deployment

We need to set `V1beta1CSRAPICompatibility` dynamically to registration controller.

The PR of `registration-operator` [pr 230](https://github.com/open-cluster-management-io/registration-operator/pull/230) on project api, I will modify go.mod and vendor after merging this PR

an example of klustelet CR
image

Signed-off-by: ivan-cai <caijing.cai@alibaba-inc.com>


ivan-cai:
@mikeshng @qiujian16  pls review this pr, thanks!

mikeshng:
I think instead of me @deads2k should be the one reviewing this change. David could you please take a look? Thanks.

My comment is that it seems like the word `hubRegistration` and `spokeRegistration` might be redundant. Is `featureGates` good enough?

operator/v1/types.go
```diff
@@ -53,6 +53,10 @@ type ClusterManagerSpec struct {
 	// +optional
 	// +kubebuilder:default={mode: Default}
 	DeployOption ClusterManagerDeployOption `json:"deployOption,omitempty"`
+
+	// HubRegistrationFeatureGates represents the list of feature gates for hub registration
+	// +optional
+	HubRegistrationFeatureGates []string `json:"hubRegistrationFeatureGates,omitempty"`
```
qiujian16: maybe `FeatureGates` instead.

operator/v1/types.go
```diff
@@ -284,6 +288,10 @@ type KlusterletSpec struct {
 	// DeployOption contains the options of deploying a klusterlet
 	// +optional
 	DeployOption KlusterletDeployOption `json:"deployOption,omitempty"`
+
+	// SpokeRegistrationFeatureGates represents the list of feature gates for spoke registration
+	// +optional
+	SpokeRegistrationFeatureGates []string `json:"spokeRegistrationFeatureGates,omitempty"`
```
qiujian16: this will need more consideration and document.

1. if this is not set, default feature gates will be used.
2. If this is set, will it replace all default feature gates or it overrides the default feature gates?
3. What should user inputs? is it as `["featuregate1=true", "featuregate2=false"]`?

operator/v1/types.go
ivan-cai: > maybe `FeatureGates` instead.

ok

operator/v1/types.go
ivan-cai: > this will need more consideration and document.
> 
> 1. if this is not set, default feature gates will be used.
> 2. If this is set, will it replace all default feature gates or it overrides the default feature gates?
> 3. What should user inputs? is it as `["featuregate1=true", "featuregate2=false"]`?

I will add more consideration and document.

1. yes
2. it will not replace all default feature gates. if you set featuregate1, it will only overrides featuregate1
3. yes

ivan-cai:
@qiujian16 pls review again.  I have change `HubRegistrationFeatureGates` and `SpokeRegistrationFeatureGates` to `FeatureGates`, and I add more description. thx!

ivan-cai:
> I think instead of me @deads2k should be the one reviewing this change. David could you please take a look? Thanks.
> 
> My comment is that it seems like the word `hubRegistration` and `spokeRegistration` might be redundant. Is `featureGates` good enough?

@mikeshng Yes, `featureGates` is good enough, I have changed.

operator/v1/types.go
```diff
@@ -53,6 +53,13 @@ type ClusterManagerSpec struct {
 	// +optional
 	// +kubebuilder:default={mode: Default}
 	DeployOption ClusterManagerDeployOption `json:"deployOption,omitempty"`
+
+	// FeatureGates represents the list of feature gates for hub registration
+	// If it is set empty, default feature gates will be used.
+	// If it is set not empty, hub registration controller will use all items of it as args.
```
qiujian16: I do not think we should restrict it to registration right? Also we should explicitly describe this is to override the default features.

ivan-cai: @qiujian16 If we are not restrict it to registration, we do not know paas items of it to which deployment(registration? work? placement? or their webhook?). 
So, we may be should separate registration feature gates, work feature gates,... ?
And I found that currently only registration has feature gates, other components can be extended. should I use `regitrationFeatureGates` instead of `FeatureGates`, and can add `workFeatureGates`/`placementFeatureGates` when them are needed?

like this:
image

qiujian16: I still think we should have a FeatureGates, and we probably should make it as a struct instead of a []string. Alternatively we should have something like:

```
registrationConfig:
  featureGates: [...]
```

@deads2k thought?

ivan-cai: I think you are right. We need an entry that contains all registration configuration.

ivan-cai:
@qiujian16 @deads2k  I hava add `RegistrationConfiguration` struct, featureGates in it. 
pls review again

qiujian16:
I think it looks fine.

operator/v1/types.go
```diff
@@ -53,6 +53,19 @@ type ClusterManagerSpec struct {
...
+	RegistrationConfiguration RegistrationConfiguration `json:"registrationConfiguration,omitempty"`
+}
+
+type RegistrationConfiguration struct {
+	// FeatureGates represents the list of feature gates for registration
+	// If it is set empty, default feature gates will be used.
+	// If it is set not empty, registration controller will use all items of it as args.
+	// And it should be set like ["featuregate1=true", "featuregate2=false"]
+	// +optional
+	FeatureGates []string `json:"featureGates,omitempty"`
```
yue9944882: fully structurelize the expression `a=true` into a string-bool map, in case of injection

operator/v1/types.go
```diff
@@ -284,6 +297,10 @@ type KlusterletSpec struct {
 	// DeployOption contains the options of deploying a klusterlet
 	// +optional
 	DeployOption KlusterletDeployOption `json:"deployOption,omitempty"`
+
+	// RegistrationConfiguration contains the configuration of registration
+	// +optional
+	RegistrationConfiguration RegistrationConfiguration `json:"registrationConfiguration,omitempty"`
```
yue9944882: omitempty doesn take effect on struct, consider make this a pointer

operator/v1/types.go
```diff
@@ -53,6 +53,19 @@ type ClusterManagerSpec struct {
...
 	DeployOption ClusterManagerDeployOption `json:"deployOption,omitempty"`
+
+	// RegistrationConfiguration contains the configuration of registration
+	// +optional
+	RegistrationConfiguration RegistrationConfiguration `json:"registrationConfiguration,omitempty"`
+}
+
+type RegistrationConfiguration struct {
+	// FeatureGates represents the list of feature gates for registration
+	// If it is set empty, default feature gates will be used.
```
yue9944882, line 64: plz clarify how non-empty feature-gate merges w/ those default set.

operator/v1/types.go
ivan-cai: 👌🏻

ivan-cai:
@deads2k @qiujian16 plz review, thank you!

operator/v1/types.go
```diff
@@ -53,6 +53,18 @@ type ClusterManagerSpec struct {
...
+	// +optional
+	RegistrationConfiguration *RegistrationConfiguration `json:"registrationConfiguration,omitempty"`
+}
+
+type RegistrationConfiguration struct {
+	// FeatureGates represents the list of feature gates for registration
+	// If it is set empty, default feature gates will be used.
+	// If it is set not empty, registration controller will use all items of it as args. And the default feature gates will be overrode if set.
+	// +optional
+	FeatureGates map[string]bool `json:"featureGates,omitempty"`
```
deads2k: What is the behavior if the featuregate keys specified (the strings) are invalid or not recognized?

This will effectively turn your feature gates into an API. Can you walk me through how an example like this from kube would work on this struct.
1. release 5, featuregate/Foo does not exist.  Kube does not allow specifying featuregate/Foo=false
2. release 6, featuregate/Foo exists and is false by default.  It is now possible to set featuregate/Foo=[false|true]
3. release 7, featuregate/Foo exists and is true by default.  If a cluster-admin upgrading from 6 to seven wants to continue having featuregate/Foo=false, he can set featuregate/Foo=false before upgrading.  Let's say the cluster-admin wants featuregate/Foo=false.
4. release 8, featuregate/Foo exists, is set to true, and can no longer be set to false. Setting featuregate/Foo=false causes the kube-apiserver to exit.  In an HA upgrade, this allows determination during testing in test or stage clusters and rollback. This is preferable to having the cluster-admin intent not honored.

Using this API, how would similar feedback be provided back to the person setting this value.

ivan-cai:
@deads2k  If the featuregate keys specified (the strings) are invalid or not recognized, we can discard it in registration-operator. How about this?

operator/v1/types.go
yue9944882: a working example i can think of is:

1. in release5 where featuregate/`Foo` doesnt exist, the registration-operator will skip configuring `Foo` for the components and set a degrade condition in the status of `ClusterManager`/`Klusterlet`.
2. in release6 where featuregate/`Foo` is under alpha phase (exists but defaulted to disabled), setting `Foo=[true|false]` in the operator api will override the defaults.
3. in release7 where featuregate/`Foo` is beta or GA (exists and defaulted to enabled), setting `Foo=[true|false]` in the operator api will override the defaults. e.g. explicitly disabling a GA featuregate means that the admin wants `Foo`=false.
4. in release8 where featuregate/`Foo` is locked to true, the configured setting in the operator api will be skipped w/ a degrade condition reporting the conflict.

yue9944882: also, i have another concern about configuring the `--feature-gate` args per-component. say the work has a feature gate `Foo` while the registration doesnt have it, what's the expected behavior of the operator? or, how does the operator manage the mapping between feature-gate and components? a feasible way is:

```golang
var (
  // listing global feature gates here
  Foo featuregate = {phase: alpha, default: false}
  Bar featuregate = {phase: beta, default: true}
)

// manage the mapping here
var componentApplicableFeatureGates = map[<component name>][]featuregate
```

qiujian16: this is in registratonConfig, so this feature gates are only for registration. I think instead of map, maybe consider an array like

```yaml
featureGates:
- feature: Foo
   mode: Enable
- feature: Bar
   mode: Disable
```

ivan-cai:
```go
Author
@[ivan-cai](https://github.com/ivan-cai) ivan-cai Pending
@qiujian16 Plz review again. I have change FeatureGates from map[string]bool to FeatureGate array.

@ivan-cai	Reply...
}

type FeatureGate struct {
	// Feature is the key of feature gate. e.g. featuregate/Foo.
	Feature string `json:"feature"`

	// Mode is either Enable, Disable, "" where "" is Disable by default.
	// In Enable mode, a valid feature gate `featuregate/Foo` will be set to "--featuregate/Foo=true".
	// In Disable mode, a valid feature gate `featuregate/Foo` will be set to "--featuregate/Foo=false".
	// +kubebuilder:default:=Disable
	// +optional
	Mode FeatureGateModeType `json:"mode,omitempty"`
}

type FeatureGateModeType string

const (
	// Valid FeatureGateModeType value is Enable, Disable.
	FeatureGateModeTypeEnable  FeatureGateModeType = "Enable"
	FeatureGateModeTypeDisable FeatureGateModeType = "Disable"
)
```

ivan-cai:
@qiujian16 Plz review again. I have change FeatureGates from map[string]bool to FeatureGate array.

operator/v1/types.go
```diff
@@ -53,8 +53,44 @@ type ClusterManagerSpec struct {
...
 }
 
+type FeatureGate struct {
+	// Feature is the key of feature gate. e.g. featuregate/Foo.
+	Feature string `json:"feature"`
+
+	// Mode is either Enable, Disable, "" where "" is Disable by default.
+	// In Enable mode, a valid feature gate `featuregate/Foo` will be set to "--featuregate/Foo=true".
+	// In Disable mode, a valid feature gate `featuregate/Foo` will be set to "--featuregate/Foo=false".
+	// +kubebuilder:default:=Disable
```
qiujian16, line 83: add enum marker

operator/v1/types.go
```diff
@@ -53,8 +53,44 @@ type ClusterManagerSpec struct {
...
+	//   2. If featuregate/Foo exists and is false by default. It is now possible to set featuregate/Foo=[false|true]
+	//   3. If featuregate/Foo exists and is true by default. If a cluster-admin upgrading from 1 to 2 wants to continue having featuregate/Foo=false,
+	//  	he can set featuregate/Foo=false before upgrading. Let's say the cluster-admin wants featuregate/Foo=false.
+	// +optional
+	FeatureGates []FeatureGate `json:"featureGates,omitempty"`
 }
 
+type FeatureGate struct {
+	// Feature is the key of feature gate. e.g. featuregate/Foo.
+	Feature string `json:"feature"`
```
qiujian16, line 78: I think this is required?

qiujian16 (commented):
thanks this looks good to me. just some nit comment. Also would you add some tests in tests directory just to ensure the API works as desired?

ivan-cai:
> thanks this looks good to me. just some nit comment. Also would you add some tests in tests directory just to ensure the API works as desired?

@qiujian16 Plz review again. I have add enum marker and required. And I also add some tests for FeatureGates.

operator/v1/types.go
qiujian16: // +kubebuilder:validation:Required

qiujian16:
/lgtm

would need @deads2k to take another look.

ivan-cai:
> /lgtm
> 
> would need @deads2k to take another look.

@deads2k Plz take a look. Thanks！

deads2k:
how will a cluster-admin know if he has typo'd a name?  Discarding so the failure isn't fatal makes sense, but reporting-wise I like @yue9944882's suggestion to set the degraded condition.  that seems appropriate.

/approve
/hold

holding to be sure we agree about the condition after discarding. If we do, then feel free to release the hold.

ivan-cai:
> how will a cluster-admin know if he has typo'd a name? Discarding so the failure isn't fatal makes sense, but reporting-wise I like @yue9944882's suggestion to set the degraded condition. that seems appropriate.
> 
> /approve /hold
> 
> holding to be sure we agree about the condition after discarding. If we do, then feel free to release the hold.

We can update a condition like this to ClusterManager/Klusterlet:
```go
  - lastTransitionTime: "2022-04-27T21:00:10Z"
    message: featuregate/Foo is not exists, discarding it
    reason: FeatureKeyNotExists
    status: "False"
    type: RegistrationFeatureGatesValid
```
How about this？ we can determine the key and reason. @deads2k @yue9944882 @qiujian16

qiujian16:
@ivan-cai I think it is good. Let's merge this. Thanks!

/unhold

PR #151, opened 2022-05-06
Add a helper function to get all valid ManagedClusterSetBindings

mikeshng:
Signed-off-by: Mike Ng <ming@redhat.com>

Based on the community meeting discussion, this repo was recommended to be a place where some OCM cluster API helper functions can live. I am looking for some feedback before I proceed with enhancing the tests and adding more test cases.

### DIFF
--- a/cluster/v1beta1/helpers.go
+++ b/cluster/v1beta1/helpers.go
@@ -3,21 +3,27 @@ package v1beta1
 import (
 	"fmt"
 
+	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	v1 "open-cluster-management.io/api/cluster/v1"
 )
 
-type ManagedClusterGetter interface {
+type ManagedClustersGetter interface {
 	List(selector labels.Selector) (ret []*v1.ManagedCluster, err error)
 }
 
-type ManagedClusterSetGetter interface {
+type ManagedClusterSetsGetter interface {
 	List(selector labels.Selector) (ret []*ManagedClusterSet, err error)
 }
 
+type ManagedClusterSetBindingsGetter interface {
+	List(namespace string, selector labels.Selector) (ret []*ManagedClusterSetBinding, err error)
+}
+
 // GetClustersFromClusterSet return the ManagedClusterSet's managedClusters
-func GetClustersFromClusterSet(clusterSet *ManagedClusterSet, clusterGetter ManagedClusterGetter) ([]*v1.ManagedCluster, error) {
+func GetClustersFromClusterSet(clusterSet *ManagedClusterSet,
+	clustersGetter ManagedClustersGetter) ([]*v1.ManagedCluster, error) {
 	var clusters []*v1.ManagedCluster
 
 	if clusterSet == nil {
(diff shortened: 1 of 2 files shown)


pkg/helpers/cluster/helpers.go
```diff
@@ -0,0 +1,39 @@
...
+
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	clusterclientset "open-cluster-management.io/api/client/cluster/clientset/versioned"
+	clusterapiv1beta1 "open-cluster-management.io/api/cluster/v1beta1"
+)
+
+// GetValidManagedClusterSetBindings returns all bindings found in the namespace.
+func GetValidManagedClusterSetBindings(
```
qiujian16: I think this helper should be in the pkg/apis/cluster/v1beta1.

It may have a `getter` interface so you would not need it to rely on client code. In this case, user can implement getter using either informer or controller-runtime

qiujian16 (commented):
thanks I think we would need several different helpers. 

1. get decisions from a placement
2. get cluster from clusterset
3. get clusterset of a cluster

FYI @ldpliu @haoqing0110

ldpliu:
Submitted [pr](https://github.com/open-cluster-management-io/api/pull/153) for 
1. get cluster from clusterset
2. get clusterset of a cluster

mikeshng:
Also changed `ManagedClusterGetter` => `ManagedClustersGetter` and `ManagedClusterSetGetter` => `ManagedClusterSetsGetter` to match convention. 

Please take a look @qiujian16 Thanks.

cluster/v1beta1/helpers.go
```diff
@@ -82,3 +90,29 @@ func BuildClusterSelector(clusterSet *ManagedClusterSet) (labels.Selector, error
 		return nil, fmt.Errorf("selectorType is not right: %s", clusterSet.Spec.ClusterSelector.SelectorType)
 	}
 }
+
+// GetValidManagedClusterSetBindings returns all bindings found in the namespace.
+func GetValidManagedClusterSetBindings(namespace string,
+	clusterSetBindingsGetter ManagedClusterSetBindingsGetter,
+	clusterSetsGetter ManagedClusterSetsGetter) ([]*ManagedClusterSetBinding, error) {
+	// get all clusterset bindings under the namespace
+	bindings, err := clusterSetBindingsGetter.List(namespace, labels.Everything())
```
qiujian16, line 97: hrm, should we have a  binding status to reflect whether there is valid clusterset? This could be useful since an operator would not need permission to list clusterset.

FYI @elgnay @ldpliu

qiujian16, line 97: hrm, should we have a  binding status to reflect whether there is valid clusterset? This could be useful since an operator would not need permission to list clusterset.

FYI @elgnay @ldpliu

ldpliu: There is an issue to trace this: https://github.com/stolostron/backlog/issues/21580#issuecomment-1094625671

mikeshng:
@qiujian16 with the merge of https://github.com/open-cluster-management-io/api/pull/158 does that mean this PR is no longer valid? Thanks.

qiujian16:
maybe still useful but avoid list all clustersets.

mikeshng:
@qiujian16 could you please take another look? I am not sure if the bound status is implemented yet but I am guessing the condition check is against status == true. Thanks.

qiujian16:
/approve
/lgtm

need this https://github.com/open-cluster-management-io/registration/pull/248

PR #152, opened 2022-05-12
Feat: Adding work subject executor feature related api

yue9944882:


work/v1/types.go
```diff
@@ -39,6 +39,11 @@ type ManifestWorkSpec struct {
 	// ManifestConfigs represents the configurations of manifests defined in workload field.
 	// +optional
 	ManifestConfigs []ManifestConfigOption `json:"manifestConfigs,omitempty"`
+
+	// Executor is the configuration of how work agent executes/applies the workloads to the local managed
+	// cluster.
+	// +optional
+	Executor *ManifestWorkExecutor `json:"executor,omitempty"`
```
qiujian16, line 49: is it because we want backward compatible so we set it as a pointer? I think we should doc the meaning of setting this as nil.

work/v1/types.go
```diff
@@ -84,6 +89,52 @@ type ManifestConfigOption struct {
 	FeedbackRules []FeedbackRule `json:"feedbackRules"`
 }
 
+// ManifestWorkExecutor is the executor that applies the resources to the managed cluster. i.e. the
+// work agent.
+type ManifestWorkExecutor struct {
+	// Subject is the subject identity which the work agent uses to talk to the
+	// local cluster when applying the resources.
+	// +optional
+	Subject *ManifestWorkExecutorSubject `json:"subject,omitempty"`
```
qiujian16: maybe not pointer here

work/v1/types.go
```diff
@@ -84,6 +89,52 @@ type ManifestConfigOption struct {
...
+	// in the managed cluster.
+	ExecutorSubjectTypeServiceAccount ManifestWorkExecutorSubjectType = "ServiceAccount"
+)
+
+// ManifestWorkSubjectServiceAccount is
+type ManifestWorkSubjectServiceAccount struct {
+	// Namespace is the namespace of the service account.
+	// +kubebuilder:validation:Required
+	// +required
+	Namespace string `json:"namespace"`
```
qiujian16, line 136: should we add some format validation using regex here?

qiujian16 (commented):
I think we should add some tests in test directory. e.g.

1. executor is not set
2. set an invalid executor type.

work/v1/types.go
yue9944882: added doc about the nil executor

work/v1/types.go
yue9944882: updated

work/v1/types.go
yue9944882: added validation regex upon namespace/name

test/integration/manifestwork_test.go
```diff
@@ -0,0 +1,75 @@
...
+
+	workv1 "open-cluster-management.io/api/work/v1"
+
+	"github.com/onsi/ginkgo"
+	"github.com/onsi/gomega"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/rand"
+)
+
+var _ = ginkgo.Describe("ManifestWork API test", func() {
```
deads2k, line 15: these tests don't seem to cover the case of a valid manifestwork and ensuring proper permission bounding.  Where are those tests?

EDIT:  I see, API repo, so this is appropriate and we add the behavior tests with the controller.

work/v1/types.go
```diff
@@ -39,6 +39,14 @@ type ManifestWorkSpec struct {
 	// ManifestConfigs represents the configurations of manifests defined in workload field.
 	// +optional
 	ManifestConfigs []ManifestConfigOption `json:"manifestConfigs,omitempty"`
+
+	// Executor is the configuration for customizing how work agent executes/applies the workloads to the
```
deads2k: Let's get more specific than customizing.  Setting the executor tells the work agent to check whether the executor has permissions to perform mutations on the individual resources in manifestwork.

deads2k:
I'd like to see a minor tweak to the comment, the API looks good.

/approve
/hold

holding for the godoc update, feel free to re-tag after and release the hold.

work/v1/types.go
yue9944882: clarified the godoc

test/integration/manifestwork_test.go
yue9944882: yeah the e2e tests covering the actual functionalities will be placed to the work repo

yue9944882:
@deads2k please take another review, thanks!

qiujian16 (commented):
/lgtm
/unhold

PR #153, opened 2022-05-12
add help func for cluster and clusterset

ldpliu:
Signed-off-by: ldpliu <daliu@redhat.com>

### DIFF
--- a/cluster/v1beta1/helpers.go
+++ b/cluster/v1beta1/helpers.go
@@ -1,0 +1,84 @@
+package v1beta1
+
+import (
+	"fmt"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	v1 "open-cluster-management.io/api/cluster/v1"
+)
+
+type ManagedClusterGetter interface {
+	List(selector labels.Selector) (ret []*v1.ManagedCluster, err error)
+}
+
+type ManagedClusterSetGetter interface {
+	List(selector labels.Selector) (ret []*ManagedClusterSet, err error)
+}
+
+// GetClustersFromClusterSet return the ManagedClusterSet's managedClusters
+func GetClustersFromClusterSet(clusterSet *ManagedClusterSet, clusterGetter ManagedClusterGetter) ([]*v1.ManagedCluster, error) {
+	var clusters []*v1.ManagedCluster
+
+	if clusterSet == nil {
+		return nil, nil
+	}
+
+	clusterSelector, err := BuildClusterSelector(clusterSet)
+	if err != nil {
+		return nil, err
+	}
+	if clusterSelector == nil {
+		return nil, fmt.Errorf("failed to build ClusterSelector with clusterSet: %v", clusterSet)
+	}
+	clusters, err = clusterGetter.List(clusterSelector)
+	if err != nil {
+		return nil, fmt.Errorf("failed to list ManagedClusters: %w", err)
+	}
+	return clusters, nil
+}
+
+// GetClusterSetsOfClusterByCluster return the managedClusterSets of a managedCluster
+func GetClusterSetsOfCluster(cluster *v1.ManagedCluster, clusterSetGetter ManagedClusterSetGetter) ([]*ManagedClusterSet, error) {
+	var returnClusterSets []*ManagedClusterSet
+
+	if cluster == nil {
+		return nil, nil
+	}
+
+	allClusterSets, err := clusterSetGetter.List(labels.Everything())
+	if err != nil {
+		return nil, err
+	}
+	for _, clusterSet := range allClusterSets {
+		clusterSelector, err := BuildClusterSelector(clusterSet)
+		if err != nil {
+			return nil, err
+		}
+		if clusterSelector == nil {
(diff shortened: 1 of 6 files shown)


cluster/v1beta1/helpers.go
```diff
@@ -0,0 +1,119 @@
...
+}
+
+type ManagedClusterSetLister interface {
+	List(selector labels.Selector) (ret []*ManagedClusterSet, err error)
+}
+
+// GetClustersFromClusterSet return the ManagedClusterSet's managedClusters
+func GetClustersFromClusterSet(clusterSet *ManagedClusterSet, clusterLister ManagedClusterLister) ([]*v1.ManagedCluster, error) {
+	var clusterSelector labels.Selector
+	var err error
```
qiujian16: why declare err at the beginning?

cluster/v1beta1/helpers.go
```diff
@@ -0,0 +1,119 @@
...
+	var clusters []*v1.ManagedCluster
+
+	if clusterSet == nil {
+		return clusters, nil
```
qiujian16: return nil, nil here?

cluster/v1beta1/helpers.go
```diff
@@ -0,0 +1,119 @@
...
+	clusters, err = clusterLister.List(clusterSelector)
+	if err != nil {
+		return clusters, fmt.Errorf("failed to list ManagedClusters: %w", err)
+	}
+	return clusters, nil
+}
+
+// GetClusterSetsOfClusterByClusterName return the managedClusterSets of a managedCluster using clusterName
+func GetClusterSetsOfClusterByClusterName(clusterName string, clusterSetLister ManagedClusterSetLister, clusterLister ManagedClusterLister) ([]*ManagedClusterSet, error) {
+	var err error
```
qiujian16: this is not needed

cluster/v1beta1/helpers.go
```diff
@@ -0,0 +1,119 @@
...
+				returnClusterSets = append(returnClusterSets, set)
+				break
+			}
+		}
+	}
+	return returnClusterSets, err
+}
+
+// GetClusterSetsOfClusterByClusterLabels return the managedClusterSets of a managedCluster using cluster labels
+func GetClusterSetsOfClusterByClusterLabels(cluster *v1.ManagedCluster, clusterSetLister ManagedClusterSetLister) ([]*ManagedClusterSet, error) {
```
qiujian16: I think we need GetClusterSetsOfClusterByCluster which understand clusterselector type. GetClusterSetsOfClusterByClusterName seems not necessary.

ldpliu: done

cluster/v1beta1/helpers.go
ldpliu: done

cluster/v1beta1/helpers.go
ldpliu: done

cluster/v1beta1/helpers.go
ldpliu: changed

cluster/v1beta1/helpers.go
```diff
@@ -0,0 +1,86 @@
...
+
+import (
+	"fmt"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	v1 "open-cluster-management.io/api/cluster/v1"
+)
+
+const clusterSetLabel = "cluster.open-cluster-management.io/clusterset"
```
qiujian16: is this in the type.go? if not we should put it in it and also make it public

cluster/v1beta1/helpers.go
```diff
@@ -0,0 +1,86 @@
...
+
+type ManagedClusterLister interface {
+	List(selector labels.Selector) (ret []*v1.ManagedCluster, err error)
```
qiujian16, line 12: do we know if this can be implemented in controller runtime?

qiujian16: I would call it ManagedClusterGetter and ManagedClusterSetGetter

cluster/v1beta1/helpers.go
ldpliu: move it to type.go

cluster/v1beta1/helpers.go
ldpliu: 1. Yes, I implement a controller runtime func in ut files:  https://github.com/open-cluster-management-io/api/pull/153/files#diff-efeb535d351c5e85c1e72ba7905e9fb47e76336faa6fea8f9eb58e25c9e72449R116
2. I think the ManagedClusterLister/ManagedClusterSetLister is better as it's a standard list interface. If you still think ManagedClusterGetter/ManagedClusterSetGetter is better. I will change it. And @elgnay what do you think about this?

qiujian16: I think we should use different name to avoid confusion.

ldpliu: > I think we should use different name to avoid confusion.

Got it, and changed to ManagedClusterGetter and ManagedClusterSetGetter

qiujian16 (commented):
/approve

PR #154, opened 2022-05-16
Work Update strategy

qiujian16:


qiujian16:
still need to add tests.

work/v1/0000_00_work.open-cluster-management.io_manifestworks.crd.yaml
```diff
@@ -75,7 +75,6 @@ spec:
                     description: ManifestConfigOption represents the configurations of a manifest defined in workload field.
                     type: object
                     required:
-                      - feedbackRules
```
yue9944882, line 112: is this intended change?

yue9944882 (commented):
looks like the change about `feedbackRules` is not related to this PR, i think we should separate it to another PR or do it in a follow-up. otherwise LGTM

work/v1/0000_00_work.open-cluster-management.io_manifestworks.crd.yaml
qiujian16: This allows that user only set updateStrategy but not feedbackRules.

work/v1/types.go
```diff
@@ -79,9 +79,55 @@ type ManifestConfigOption struct {
 	ResourceIdentifier ResourceIdentifier `json:"resourceIdentifier"`
 
 	// FeedbackRules defines what resource status field should be returned.
+	// +optional
```
deads2k, line 91: This is technically breaking, but newer clients needed to handle the missing case anyway.  Please clarify if there is a difference between missing and empty (there shouldn't be a difference).

work/v1/types.go
```diff
@@ -79,9 +79,55 @@ type ManifestConfigOption struct {
...
+	FeedbackRules []FeedbackRule `json:"feedbackRules,omitempty"`
+
+	// UpdateStrategy defines the strategy to update this manifest
+	// optional
```
deads2k, line 96: Please clarify if there is a difference between missing and empty (there shouldn't be a difference without a really good reason).

work/v1/types.go
```diff
@@ -79,9 +79,55 @@ type ManifestConfigOption struct {
...
+	UpdateStrategy UpdateStrategy `json:"updateStrategy"`
```
deads2k: to make this work how you want as optional, I think you it to be a pointer and be omitempty.

work/v1/types.go
```diff
@@ -79,9 +79,55 @@ type ManifestConfigOption struct {
...
+	// on the spoke to own the control of the resource.
+	UpdateStrategyTypeCreateOnly UpdateStrategyType = "CreateOnly"
+
+	// ServerSideApply type means to update resource using server side apply with work-controller as the field manager.
+	// If there is conflict, the related Applied condition of manifest will be in the status of False with the
+	// reason of ApplyConflict. This type allows another controller on the spoke to control certain field of the resource.
+	UpdateStrategyTypeServerSideApply UpdateStrategyType = "ServerSideApply"
+)
+
+type ServerSideApplyConfig struct {
```
deads2k, line 188: Do you want to provide a way to control the agent string as well?  Some kind of suffix perhaps rather than allowing complete control.

Once this capability exists, it becomes possible to have two manifestworks touching different parts of the same resource.

work/v1/types.go
qiujian16: comment added

work/v1/types.go
qiujian16: yes, add an fieldManager field with prefix validation

yue9944882 (commented):
/lgtm

deads2k:
/approve

/hold

not sure why it's a WIP.  holding until you think its ready, feel free to release when it's no  longer WIP.

PR #156, opened 2022-05-24
add help func for placement

haoqing0110:

### DIFF
--- a/cluster/v1beta1/helpers.go
+++ b/cluster/v1beta1/helpers.go
@@ -2,10 +2,12 @@ package v1beta1
 
 import (
 	"fmt"
+	"sync"
 
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/util/sets"
 	v1 "open-cluster-management.io/api/cluster/v1"
 )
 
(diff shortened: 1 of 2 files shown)


cluster/v1beta1/helpers.go
```diff
@@ -82,3 +86,23 @@ func BuildClusterSelector(clusterSet *ManagedClusterSet) (labels.Selector, error
 		return nil, fmt.Errorf("selectorType is not right: %s", clusterSet.Spec.ClusterSelector.SelectorType)
 	}
 }
+
+// GetDecisionsOfPlacement return the PlacementDecisions generated by the Placement
+func GetDecisionsOfPlacement(placement *Placement, decisionGetter PlacementDecisionGetter) ([]*PlacementDecision, error) {
```
qiujian16: I am not sure if it is helpful. I think you would want to return a sets.String with all cluster names. Considering it should be used during schedule, you probably want to return the clusters added to the decision, the clusters removed from the decisions.

haoqing0110: @qiujian16 yes, update this func to GetDecisionClustersOfPlacement which will return cluster names of one placement's decision.

cluster/v1beta1/helpers.go
```diff
@@ -82,3 +83,85 @@ func BuildClusterSelector(clusterSet *ManagedClusterSet) (labels.Selector, error
...
+}
+
+func (pdcr *PlacementDecisionClustersResponse) Deleted() sets.String {
+	return pdcr.deletedDecisionClusters
+}
+
+func NewPlacementDecisionClustersTracker(placement *Placement, pdl PlacementDecisionGetter) (*PlacementDecisionClustersTracker, error) {
+	return &PlacementDecisionClustersTracker{
+		placement:               placement,
+		placementDecisionGetter: pdl,
```
qiujian16: you would need existing decision as an optional input also, right?

cluster/v1beta1/helpers.go
```diff
@@ -82,3 +83,85 @@ func BuildClusterSelector(clusterSet *ManagedClusterSet) (labels.Selector, error
...
+	placementDecisionGetter PlacementDecisionGetter
+	decisionClusters        sets.String
+}
+
+type PlacementDecisionClustersResponse struct {
+	addeddDecisionClusters  sets.String
+	deletedDecisionClusters sets.String
+}
+
+func (pdcr *PlacementDecisionClustersResponse) Added() sets.String {
```
qiujian16: this is racy consider you have to gorouting visit this at the same time.

cluster/v1beta1/helpers.go
```diff
@@ -82,3 +83,85 @@ func BuildClusterSelector(clusterSet *ManagedClusterSet) (labels.Selector, error
...
+
+func NewPlacementDecisionClustersTracker(placement *Placement, pdl PlacementDecisionGetter) (*PlacementDecisionClustersTracker, error) {
+	return &PlacementDecisionClustersTracker{
+		placement:               placement,
+		placementDecisionGetter: pdl,
+	}, nil
+}
+
+// get() return the placement decision cluster names, and do not persist it in tracker's decisionClusters
+func (pdct *PlacementDecisionClustersTracker) get() (sets.String, error) {
```
qiujian16: I think you could have a Get()(sets.string, sets.sting, error) that returns added and removed. In the func, you would need a lock to avoid racy; get the update-to-date decision, compared with and update existing and return.

Refresh/Added/Deleted may not be needed.

haoqing0110: Removed Refresh/Added/Deleted and add rw lock to avoid racy.

cluster/v1beta1/helpers.go
```diff
@@ -108,3 +110,70 @@ func GetBoundManagedClusterSetBindings(namespace string,
...
+}
+
+func NewPlacementDecisionClustersTracker(placement *Placement, pdl PlacementDecisionGetter, existDecisionClusters sets.String) *PlacementDecisionClustersTracker {
+	pdct := &PlacementDecisionClustersTracker{
+		placement:               placement,
+		placementDecisionGetter: pdl,
+	}
+	if existDecisionClusters != nil {
+		pdct.decisionClusters = existDecisionClusters
+	}
```
qiujian16, line 131: should you ini the decisitionClusters when existDecisionClusters is nil?

also maybe existingScheduledClusters

qiujian16 (commented):
/approve

one nit comment

cluster/v1beta1/helpers.go
haoqing0110: fixed.

qiujian16:
/lgtm
/hold

in case you need others to review

PR #159, opened 2022-05-26
Add GetClusterSetName to get the name of the ManagedClusterSet from the ManagedCluster's clusterSet label.

mikeshng:
Signed-off-by: Mike Ng <ming@redhat.com>

### DIFF
--- a/cluster/v1beta1/helpers.go
+++ b/cluster/v1beta1/helpers.go
@@ -9,6 +9,10 @@ import (
 	v1 "open-cluster-management.io/api/cluster/v1"
 )
 
+const (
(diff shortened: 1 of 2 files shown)


mikeshng:
@qiujian16 could you please take a look and provide some feedback? Thanks.

qiujian16:
this won't work when we add labeSelector based clusterset. I think we should use GetClusterSetsOfCluster.

mikeshng:
> this won't work when we add labeSelector based clusterset. I think we should use GetClusterSetsOfCluster.

Thanks. Closing.

PR #162, opened 2022-06-08
move registration feature gates to api, can be used by registration-o…

ivan-cai:
move registration feature gates to api, can be used by registration-operator for check input feature gates if valid.

Signed-off-by: ivan-cai <caijing.cai@alibaba-inc.com>

### DIFF
--- a/feature/feature.go
+++ b/feature/feature.go
@@ -1,0 +1,58 @@
+package feature
+
+import (
+	"k8s.io/component-base/featuregate"
+)
+
+const (
+	// Every feature gate should add method here following this template:
+	//
+	// // owner: @username
+	// // alpha: v1.X
+	// MyFeature featuregate.Feature = "MyFeature"
+
+	// ClusterClaim will start a new controller in the spoke-agent to manage the cluster-claim
+	// resources in the managed cluster.
+	//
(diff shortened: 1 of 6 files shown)


ivan-cai:
@qiujian16 As we discussed in  [registration-operator pr 230](https://github.com/open-cluster-management-io/registration-operator/pull/230)

qiujian16 (commented):
hrm, could we just define the feature name or feature spec in api? I do not think we should have init here, this could be messy when others import api given feature is a global var.

ivan-cai:
> hrm, could we just define the feature name or feature spec in api? I do not think we should have init here, this could be messy when others import api given feature is a global var.

@qiujian16 I have delete `init()` and `DefaultSpokeMutableFeatureGate` & `DefaultHubMutableFeatureGate`. Plz review again.

qiujian16:
thanks, one more nit thing. How about we have a separate package like `feature` to put it? I do not think it relates to a certain version of API, right?

ivan-cai:
> thanks, one more nit thing. How about we have a separate package like `feature` to put it? I do not think it relates to a certain version of API, right?

@qiujian16 Yes, I have changed. Plz review again, thanks!

Issue #165, opened 2022-06-23
Support hub api-server host alias


promacanthus:
In some cases, the server in the Hub cluster kubeconfig file is a domain name. For example, as shown in the example below.

```yaml
apiVersion: v1
kind: Config
clusters:
- cluster:
    server: https://xxx.yyy.com
```

The contents of this kubeconfig file will be stored in the namespace(open-cluster-manager-agent) in the secret(bootstrap-hub-kubeconfig) as the startup configuration when the managed cluster is registered.

The registration-agent will use this bootstrap-hub-kubeconfig to create CRD(managedCluster) and CSR, but the following error message will be encountered in the logs.

```shell
E0623 03:02:35.268397       1 reflector.go:138] k8s.io/client-go@v0.23.0/tools/cache/reflector.go:167: Failed to watch *v1.CertificateSigningRequest: failed to list *v1.CertificateSigningRequest: Get "https://xxx.yyy.com/apis/certificates.k8s.io/v1/certificatesigningrequests?limit=500&resourceVersion=0": dial tcp: lookup xxx.yyy.com on 10.xx.yy.zz:53: no such host
```

This is because the domain name of the Hub Cluster API Server is not available in the managed cluster's DNS servers. The solution to this problem is simply to write the host alias to the registration-agent pod. This is shown below.

```yaml
spec:
hostAliases:
- ip: "xx.xx.xx.xx"
  hostnames:
  -  "xxx.yyy.com"
```
----------

For the above problem and solution, the API needs to be modified, and the specific design and implementation are as follows.

Add a new field named HubApiServerHostAlias to the KlusterletSpec.

```golang
type KlusterletSpec struct {

	// ...

	// HubApiServerHostAlias contains the host alias for hub api server
	// +optional
	HubApiServerHostAlias *v1.HostAlias `json:"hub_api_server_host_alias,omitempty"`
}
```

**[Registration-operator](https://github.com/open-cluster-management-io/registration-operator)** Repo

Add a new rendering logic to  [manifests/klusterlet/management/klusterlet-registration-deployment.yaml](https://github.com/open-cluster-management-io/registration-operator/blob/main/manifests/klusterlet/management/klusterlet-registration-deployment.yaml) template.

```yaml
spec:
{{if .HostAlias }}
hostAliases:
- ip: {{ .HostAlias.Ip }}
  hostnames:
  {{range .HostAlias.Hostnames }}
  - {{ . }}
  {{end}}
{{end}}
```

[...]

PR #166, opened 2022-06-28
Add HostAlias in KlusterletDeployOption

promacanthus:
#165 

To support registration-agent and work-agent use domain to communicate with hub api server.


operator/v1/types.go
```diff
@@ -133,6 +133,11 @@ type KlusterletDeployOption struct {
 	// Note: Do not modify the Mode field once it's applied.
 	// +optional
 	Mode InstallMode `json:"mode"`
+
```
qiujian16: KlusterletDeployOption is to define whether klusterlet can be run in hosted or default mode. I would think HubApiServerHostAlias should be put in the spec. it cannot be `hub_api_server_host_alias`, it needs to follow the camel format such as "hubAPIServerHostAliase"

qiujian16 (commented):
thanks, I understand the problem to solve. But I do not think adding host aliase field can fully resolve the issue. If you have an addon that needs to connect to the same hub, you will meet the same problem and need to take care of the host setup on addon deployment.

I wonder if it is better to resolve it in system wide, e.g. setting up a dns record in dns server of kube so we do not need to set it per deployment. FYI @elgnay I think we see this issue before.

operator/v1/types.go
promacanthus: I had a wrong understanding of `KlusterletDeployOption`. I will put `HubApiServerHostAlias` in the spec and format as `hubAPIServerHostAlias`.

elgnay:
Agreed with @qiujian16. Besides the registration agent and work agent, klusterlet operator itself also communicates with the hub kube apiserver (https://github.com/open-cluster-management-io/registration-operator/blob/803bfb374805414222e2aff669b73588dffe3537/pkg/operators/klusterlet/controllers/ssarcontroller/klusterlet_ssar_controller.go#L159),  it's a little bit complicated for klusterlet operator to handle the HostAlias in the klusterlet crs.

operator/v1/types.go
```diff
@@ -327,6 +327,11 @@ type KlusterletSpec struct {
 	// RegistrationConfiguration contains the configuration of registration
 	// +optional
 	RegistrationConfiguration *RegistrationConfiguration `json:"registrationConfiguration,omitempty"`
+
+	// HubApiServerHostAlias contains the host alias for hub api server.
+	// registration-agent and work-agent will use it to communicate with hub api server.
+	// +optional
+	HubApiServerHostAlias *v1.HostAlias `json:"hubApiServerHostAlias,omitempty"`
```
qiujian16: does it need to be a pointer? what if user set one with empty IP or host name in it? we probably should not use v1 struct but have our own with more validation

promacanthus: I think a nil pointer indicates that agents should not use host alias. Add a new struct HubApiServerHostAlias with ip and hostname field. I think only need to set one ip and hostname for hub api server.

```golang
type HubApiServerHostAlias struct {
	// IP address of the host file entry.
	// +required
	// +kubebuilder:validation:Pattern=`^((0|[1-9]\d|1\d\d|2[0-4]\d|25[0-5])\.){3}(0|[1-9]\d|1\d\d|2[0-4]\d|25[0-5])$`
	IP string `json:"ip,omitempty" protobuf:"bytes,1,opt,name=ip"`

	// Hostname for the above IP address.
	// +required
	// +kubebuilder:validation:Pattern=`^(([a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9\-]*[a-zA-Z0-9])\.)*([A-Za-z0-9]|[A-Za-z0-9][A-Za-z0-9\-]*[A-Za-z0-9])$`
	Hostname string `json:"hostnames,omitempty" protobuf:"bytes,2,rep,name=hostnames"`
}
```

operator/v1/types.go
```diff
@@ -354,6 +359,22 @@ type NodePlacement struct {
...
 }
 
+// HubApiServerHostAlias holds the mapping between IP and hostnames that will be injected as an entry in the
+// pod's hosts file.
+type HubApiServerHostAlias struct {
+	// IP address of the host file entry.
+	// +required
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:Pattern=`^((0|[1-9]\d|1\d\d|2[0-4]\d|25[0-5])\.){3}(0|[1-9]\d|1\d\d|2[0-4]\d|25[0-5])$`
+	IP string `json:"ip,omitempty" protobuf:"bytes,1,opt,name=ip"`
```
qiujian16: no need to add protobuf, it is not supported in crd.

promacanthus: Done.   And i  think required field do not need omitempty tag.

qiujian16:
thanks,I think it looks good. verify fails. Please also noted that klusterlet-operator itself will also use this host aliase to visit hub for status check. This is implementation details though.

qiujian16:
btw, it will be good if you can add some tests https://github.com/open-cluster-management-io/api/blob/main/test/integration/klusterlet_test.go just to ensure the schema validation works as what your desired.

promacanthus:
> thanks,I think it looks good. verify fails. Please also noted that klusterlet-operator itself will also use this host aliase to visit hub for status check. This is implementation details though.

Thanks. I will keep this in mind as i think about implementation details.

promacanthus:
> btw, it will be good if you can add some tests https://github.com/open-cluster-management-io/api/blob/main/test/integration/klusterlet_test.go just to ensure the schema validation works as what your desired.

I will add some test cases to verify it.

qiujian16:
@Promacanthus verify still fails. I think it should generate deepcopy but I do not see it in this PR.

promacanthus:
> @Promacanthus verify still fails. I think it should generate deepcopy but I do not see it in this PR.

The problem has been found and fixed in the latest commit.

PR #167, opened 2022-06-30
add-on configuration

skeeey:


addon/v1alpha1/types_addondeploymentconfig.go
```diff
@@ -0,0 +1,68 @@
...
+
+type AddOnDeploymentConfigSpec struct {
+	// CustomizedVariables is a list of name-value variables for the current add-on deployment.
+	// The add-on implementation can use these variables to render its add-on deployment.
+	// The default is an empty list.
+	// +optional
+	CustomizedVariables []CustomizedVariable `json:"customizedVariables,omitempty"`
+
+	// NodePlacement enables explicit control over the scheduling of the add-on.
+	// +optional
```
qiujian16, line 37: node scheduling of the addon agent on the spoke? we should indicate that all agent pods on the addon are expected to comply with the node placement.

addon/v1alpha1/types_addondeploymentconfig.go
```diff
@@ -0,0 +1,68 @@
...
+	// spec represents a desired configuration for an add-on.
+	// +required
+	Spec AddOnDeploymentConfigSpec `json:"spec"`
+}
+
+type AddOnDeploymentConfigSpec struct {
+	// CustomizedVariables is a list of name-value variables for the current add-on deployment.
+	// The add-on implementation can use these variables to render its add-on deployment.
+	// The default is an empty list.
+	// +optional
```
qiujian16, line 27: listmap=Name

addon/v1alpha1/types_managedclusteraddon.go
```diff
@@ -147,6 +160,24 @@ type ObjectReference struct {
...
 }
 
+// ConfigReference is a reference to the current add-on configuration.
+// This resource is used to locate the configuration resource for the current add-on.
+type ConfigReference struct {
+	// This field is synced from ClusterManagementAddOn configGroupResource field.
+	ConfigGroupResource `json:",inline"`
+
+	// This field is synced from ClusterManagementAddOn defaultConfig and ManagedClusterAddOn config fields.
+	// If both them are defined, the ManagedClusterAddOn config will overwrite the ClusterManagementAddOn
```
qiujian16: both of them

addon/v1alpha1/types_clustermanagementaddon.go
```diff
@@ -35,41 +35,73 @@ type ClusterManagementAddOn struct {
...
 	// +optional
-	AddOnMeta AddOnMeta `json:"addOnMeta"`
+	AddOnMeta AddOnMeta `json:"addOnMeta,omitempty"`
 
 	// addOnConfiguration is a reference to configuration information for the add-on.
 	// In scenario where a multiple add-ons share the same add-on CRD,
 	// multiple ClusterManagementAddOn resources need to be created and reference the same AddOnConfiguration.
 	// +optional
-	AddOnConfiguration ConfigCoordinates `json:"addOnConfiguration"`
+	AddOnConfiguration ConfigCoordinates `json:"addOnConfiguration,omitempty"`
```
qiujian16, line 44: i think we should mention this is the default configuration of the addon?

addon/v1alpha1/types_managedclusteraddon.go
```diff
@@ -94,12 +99,17 @@ type ManagedClusterAddOnStatus struct {
...
 
+	// Deprecated: Use configReference instead
 	// addOnConfiguration is a reference to configuration information for the add-on.
 	// This resource is use to locate the configuration resource for the add-on.
 	// +optional
-	AddOnConfiguration ConfigCoordinates `json:"addOnConfiguration"`
+	AddOnConfiguration ConfigCoordinates `json:"addOnConfiguration,omitempty"`
+
+	// configReference is a reference to the add-on configuration.
+	// +optional
```
qiujian16, line 113: this will override the configuration reference define in clustermanagementaddon

addon/v1alpha1/types_addondeploymentconfig.go
skeeey: yes, make the description more clear

addon/v1alpha1/types_managedclusteraddon.go
skeeey: corrected

addon/v1alpha1/types_clustermanagementaddon.go
skeeey: we have described the default config scenario  in the `AddOnConfiguration.DefaultConfig`

addon/v1alpha1/types_addondeploymentconfig.go
skeeey: added

addon/v1alpha1/types_managedclusteraddon.go
skeeey: added the description

addon/v1alpha1/types_clustermanagementaddon.go
```diff
@@ -35,41 +35,73 @@ type ClusterManagementAddOn struct {
...
+	DefaultConfig ConfigReferent `json:"defaultConfig,omitempty"`
+}
+
+// ConfigGroupResource represents the GroupResource of the add-on configuration
+type ConfigGroupResource struct {
+	// group of the add-on configuration.
+	Group string `json:"group"`
+
+	// resource of the add-on configuration.
+	Resource string `json:"resource"`
```
qiujian16, line 96: required?

addon/v1alpha1/types_managedclusteraddon.go
```diff
@@ -147,6 +161,24 @@ type ObjectReference struct {
...
+	// This field is synced from ClusterManagementAddOn configGroupResource field.
+	ConfigGroupResource `json:",inline"`
+
+	// This field is synced from ClusterManagementAddOn defaultConfig and ManagedClusterAddOn config fields.
+	// If both of them are defined, the ManagedClusterAddOn config will overwrite the ClusterManagementAddOn
+	// defaultConfig.
+	ConfigReferent `json:",inline"`
+
+	// version of the add-on configuration.
+	Version string `json:"version"`
```
qiujian16: required?

addon/v1alpha1/types_clustermanagementaddon.go
```diff
@@ -35,41 +35,73 @@ type ClusterManagementAddOn struct {
...
+
+// ConfigReferent represents the namespace and name for an add-on configuration.
+type ConfigReferent struct {
+	// namespace of the add-on configuration.
+	// If this field is not set, the configuration is in the cluster scope.
+	// +optional
+	Namespace string `json:"namespace,omitempty"`
+
+	// name of the add-on configuration.
+	Name string `json:"name"`
```
qiujian16, line 110: required

addon/v1alpha1/types_addondeploymentconfig.go
```diff
@@ -0,0 +1,74 @@
...
+	NodePlacement NodePlacement `json:"nodePlacement,omitempty"`
+}
+
+// CustomizedVariable represents a customized variable for add-on deployment.
+type CustomizedVariable struct {
+	// Name of this variable.
+	// +kubebuilder:validation:Required
+	// +required
+	Name string `json:"name"`
```
qiujian16, line 48: should it has some regex validation?

addon/v1alpha1/types_addondeploymentconfig.go
```diff
@@ -0,0 +1,74 @@
...
+
+	// Value of this variable.
+	Value string `json:"value"`
```
qiujian16, line 52: you may want to add a maxlength validation.

qiujian16 (commented):
we should add some tests for this api and also addon apis.

addon/v1alpha1/types_clustermanagementaddon.go
skeeey: yes, I think it's required

addon/v1alpha1/types_managedclusteraddon.go
skeeey: it's required

addon/v1alpha1/types_clustermanagementaddon.go
skeeey: it's required

addon/v1alpha1/types_addondeploymentconfig.go
skeeey: limit the length to 255

addon/v1alpha1/types_addondeploymentconfig.go
skeeey: using c_identifier as the regex validation and limit the length to 255

skeeey:
/hold

for adding test

addon/v1alpha1/types_addondeploymentconfig.go
```diff
@@ -0,0 +1,77 @@
...
+type CustomizedVariable struct {
+	// Name of this variable.
+	// +required
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MaxLength=255
+	// +kubebuilder:validation:Pattern=^[a-zA-Z_][_a-zA-Z0-9]*$
+	Name string `json:"name"`
+
+	// Value of this variable.
+	// +kubebuilder:validation:MaxLength=255
```
deads2k: maxlength is very good. I suggest 1024 at least.  255  is tight for images+SHA for instance.

addon/v1alpha1/types_addondeploymentconfig.go
```diff
@@ -0,0 +1,77 @@
...
+	Value string `json:"value"`
+}
+
+// NodePlacement describes node scheduling configuration for the pods.
+type NodePlacement struct {
+	// NodeSelector defines which Nodes the Pods are scheduled on. The default is an empty list.
```
deads2k: Empty list means, "match all nodes", right? If so, clarify in doc.

addon/v1alpha1/types_addondeploymentconfig.go
```diff
@@ -0,0 +1,77 @@
...
+	// +optional
+	NodeSelector map[string]string `json:"nodeSelector,omitempty"`
+
+	// Tolerations is attached by pods to tolerate any taint that matches
+	// the triple <key,value,effect> using the matching operator <operator>.
+	// The default is an empty list.
```
deads2k, line 66: empty list means, "tolerate nothing" right?  If so, clarify in doc.

addon/v1alpha1/types_managedclusteraddon.go
```diff
@@ -147,6 +162,26 @@ type ObjectReference struct {
...
+type ConfigReference struct {
+	// This field is synced from ClusterManagementAddOn configGroupResource field.
+	ConfigGroupResource `json:",inline"`
+
+	// This field is synced from ClusterManagementAddOn defaultConfig and ManagedClusterAddOn config fields.
+	// If both of them are defined, the ManagedClusterAddOn config will overwrite the ClusterManagementAddOn
+	// defaultConfig.
+	ConfigReferent `json:",inline"`
+
+	// version of the add-on configuration.
```
deads2k: When the controller sets this value, how does it determine what the version is?  Is it an API version or more like a revision of the configuration like a generation.  If it's an API version, why does the serialization format matter in status?

deads2k:
some clarifications requested.  The version field isn't completely clear to me. If it's an apiversion, normally the serialization version isn't critical for status or references, so I'm curious how it will be consumed.

addon/v1alpha1/types_addondeploymentconfig.go
skeeey: set to 1024

addon/v1alpha1/types_addondeploymentconfig.go
skeeey: right, add it to doc

addon/v1alpha1/types_addondeploymentconfig.go
skeeey: yes, add it to doc

addon/v1alpha1/types_managedclusteraddon.go
skeeey: it's the api version, the addon framework use rest mapper to discovery the config api version with the config resource and group that are defined in the clustermanagementaddon, and add the version to managedclusteraddon status.

For an addon implementation, it can get the config full meta info (group, resource, version and name/namespace) from its managedclusteraddon, then the implementation use them to get the config cr, and render its addon agent

it will be easy to get the addon config cr for addon implementation with the config api version

skeeey: UPDATE

Remove the version from the managedclusteraddon status, for addon implementation, it should known the exact apiversion of the config, so the config apiversion is not necessary, and we will ask implementation to provide the config gvr to the addon framework, thus the framework will have an exact type to handle

PR #168, opened 2022-06-30
Add GOPATH, GOHOSTOS and GOHOSTARCH in makefile

promacanthus:
Sometimes we don't set the environment variables `GOPATH`, `GOHOSTOS`, and `GOHOSTARCH`, which will cause us to fail to execute the makefile.

Signed-off-by: Promacanthus <promacanthus@gmail.com>


Makefile
```diff
@@ -53,3 +53,10 @@ update-with-container: build-runtime-image
 	$(RUNTIME) run -ti --rm -v $(PWD):/go/src/open-cluster-management.io/api:z -w /go/src/open-cluster-management.io/api $(RUNTIME_IMAGE_NAME) make update-scripts update-codegen-crds
 
 include ./test/integration-test.mk
+
+check-env:
+ifeq ($(GOPATH),)
+	$(warning "environment variable GOPATH is empty, auto set from go env GOPATH") 
+export GOPATH=$(shell go env GOPATH)
+endif
+.PHONY: check-env
```
qiujian16: eof

promacanthus: Should i add one more empty line after the L62?

PR #182, opened 2022-10-14
add work helpers

zhiweiyin318:
Signed-off-by: Zhiwei Yin <zyin@redhat.com>

issue: https://github.com/open-cluster-management-io/addon-framework/issues/127

1. add work helpers 
2. upgrade k8s to v0.24.0 because the the old pkg "github.com/googleapis/gnostic"and the new "github.com/google/gnostic/extensions" conflict


qiujian16:
should this be a versioned helper? like putting this into work/v1/apply amd work/v1/builder?

zhiweiyin318:
> should this be a versioned helper? like putting this into work/v1/apply amd work/v1/builder?

@qiujian16  yes, versioned helper can help manage the api upgrade in the future.
done

qiujian16:
@zhiweiyin318 let's use tools/work/v1 as the package name all utils/work/v1

zhiweiyin318:
> @zhiweiyin318 let's use tools/work/v1 as the package name all utils/work/v1

@qiujian16 we have had  the `tools` dir  including gen swaggerdoc, so I use utils/work/v1 as tha package name.

PR #186, opened 2022-10-28
Introduce a PlaceWork API.

jnpacker:
* Initial commit
* Allows the specification of a ManifestWork.Spec
* Allows Placement targeting of the ManifestWork.Spec
* The controller will take the API Spec details and PlacementDecision and create new ManifestWork resources in the appropriate clusters
* It will also remove ManifestWorks from the appropriate clusters

Signed-off-by: Joshua Packer <jpacker@redhat.com>


jnpacker:
Related enhancement: https://github.com/open-cluster-management-io/enhancements/pull/51/files

place-work/v1alpha1/types.go
```diff
@@ -0,0 +1,134 @@
...
+	// 1. AppliedManifestWorks represents ManifestWorks have been distributed as per placement All, Partial, None, Problem
+	// 2. PlacementRefValid
+	Conditions []metav1.Condition `json:"conditions,omitempty"`
+
+	// ManifestWorkDelivery tracks each ManifestWork that was created,
+	// including that it was applied and its overall status
+	PlacedManifestWork []PlacedManifestWork `json:"placedManifestWork,omitempty"`
+
+	// Summary that reflects all relevant ManifestWorks
+	PlacedManifestWorkSummary []PlacedManifestWorkSummary `json:"summary"`
```
serngawy: Why this is an array ? we only have 1 manifestWork to apply single entry should be fine

place-work/v1alpha1/types.go
```diff
@@ -0,0 +1,134 @@
...
+
+	// Conditions contains the different condition statuses for distrbution of ManifestWork resources
+	// Valid condition types are:
+	// 1. AppliedManifestWorks represents ManifestWorks have been distributed as per placement All, Partial, None, Problem
+	// 2. PlacementRefValid
+	Conditions []metav1.Condition `json:"conditions,omitempty"`
+
+	// ManifestWorkDelivery tracks each ManifestWork that was created,
+	// including that it was applied and its overall status
+	PlacedManifestWork []PlacedManifestWork `json:"placedManifestWork,omitempty"`
```
serngawy: This could become to much data and hard to scale. why not just giving the status per cluster ex; applied/processing/available/degraded and for more info user can check the manifestWork itself. What do u think ?

serngawy (commented):
I added few comments and the **distributionStrategy** from the enhancement is missing in the API definition.

place-work/v1alpha1/types.go
jnpacker: I have to go back and check the synthetic scale test on the Controller when I put that pull request in, but i was able to handle distribution to 2000 clusters at the time.

jnpacker: Pretty much its 2000 X (ObjectRef + bool + timestamp + string)
It's a finite sized object, if we get the cluster count to high though it might become a problem.  When I revisit the synthetic scale test, I'll try it with 3000-4000 clusters to make sure and also work out the calculation.

jnpacker:
I'm looking for some thoughts on the Go / verify failure above...  `make verify` in my environment runs successfully, and from the error, the matching text looks similar...

My env. is Ubuntu, Golang 1.18.7 with controller-gen 0.6.0  (this matches what I see in the Go / verify log.

place-work/v1alpha1/doc.go
```diff
@@ -0,0 +1,9 @@
+// Package v1 contains API Schema definitions for the place-work v1 API group
+// +k8s:deepcopy-gen=package,register
+// +k8s:conversion-gen=open-cluster-management.io/api/place-work
+// +k8s:defaulter-gen=TypeMeta
+// +k8s:openapi-gen=true
+
+// +kubebuilder:validation:Optional
+// +groupName=work.open-cluster-management.io
```
qiujian16: if the group name is work, it is better to put into work dir.

place-work/v1alpha1/types.go
```diff
@@ -0,0 +1,134 @@
...
+// EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
+// NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
+
+// PlaceManifestWorkSpec defines the desired state of PlaceManifestWork
+type PlaceManifestWorkSpec struct {
+	// INSERT ADDITIONAL SPEC FIELDS - desired state of cluster
+	// Important: Run "make" to regenerate code after modifying this file
+
+	// ManifestWorkSpec is the ManifestWorkSpec that will be used to generate a per-cluster ManifestWork
+	ManifestWorkSpec work.ManifestWorkSpec `json:"manifestWorkSpec"`
```
qiujian16: manifestWorkTemplate?

place-work/v1alpha1/types.go
```diff
@@ -0,0 +1,134 @@
...
+
+	// PacementRef is the name of the Placement resource, from which a PlacementDecision will be found and used
+	// to distribute the ManifestWork
+	PlacementRef *corev1.LocalObjectReference `json:"placementRef,omitempty"`
```
qiujian16: it is better to create our own placement ref instead of using corev1 one.

place-work/v1alpha1/types.go
qiujian16: I think we might start without this field. IT could be too many items and introduce many update calls

place-work/v1alpha1/types.go
```diff
@@ -0,0 +1,134 @@
...
+	PlacedManifestWork []PlacedManifestWork `json:"placedManifestWork,omitempty"`
+
+	// Summary that reflects all relevant ManifestWorks
+	PlacedManifestWorkSummary []PlacedManifestWorkSummary `json:"summary"`
+}
+
+type PlacedManifestWorkSummary struct {
+	Total       int `json:"total"`
+	Progressing int `json:"progressing"`
+	Available   int `json:"available"`
```
qiujian16: need some comments for each field. Are these optional or required?

place-work/v1alpha1/types.go
```diff
@@ -0,0 +1,134 @@
...
+type ConditionType string
+
+const (
+	PlacementDecisionVerifiedAsExpected = "PlacementDecisionVerified"
+	PlacementDecisionNotFound           = "PlacementDecisionNotFound"
+	PlacementDecisionEmpty              = "PlacementDecisionEmpty"
+
+	AsExpected = "AsExpected"
+	Processing = "Processing"
+	Partial    = "Partial"
```
qiujian16: need some comments on the intention of Partial

qiujian16 (commented):
need to also add client-gen marker so `make update` can generate client code

qiujian16:
cc @ivan-cai

place-work/v1alpha1/types.go
jnpacker: So i've been thinking about options to trigger the updates via a backoff algorithm, which I still think is problematic, or by triggering the Place-Work-controller, but making sure not to trigger it if the last time it ran is Less Than.

1. I'm pretty sure the size was not a factor in my synthetic scale test as its a fixed size.
2. Initial controller just reconciles on a time loop, like ManifestWork.

place-work/v1alpha1/doc.go
jnpacker: Will move it to work.ocm.io

place-work/v1alpha1/types.go
jnpacker: Changed.

place-work/v1alpha1/types.go
jnpacker: Done.

place-work/v1alpha1/types.go
jnpacker: AsExpected is all applied.
Prcoessing is while applying or reconciling.
Partial should probably be NotAsExpected.

jnpacker: These are conditions for the place-work resource.

place-work/v1alpha1/types.go
```diff
@@ -0,0 +1,134 @@
...
+	// Work is an objectReference to the actual ManifestWork resource
+	Work corev1.ObjectReference `json:"workRef"`
+
+	// +kubebuilder:default=false
+	Applied bool `json:"applied"`
+
+	// ManifestWorkStatus ManifestWorkStatus.Condition can be
+	// +kubebuilder:validation:Enum=Progressing;Available;Degraded;Applied
+	// +optional
+	ManifestWorkStatus ManifestWorkStatus `json:"status"`
```
jnpacker: I was hoping to store Work* (WorkProgressing, WorkApplied, WorkAvailable or WorkDegraded from each MW as stored in the conditions.

For Example if condition type: Applied, status = false, and type: Available, status = Unknown  We would set the status in an order based on False or Unknown.
Applied = false   the value would be applied
1. Show Applied until it turns true (then move to next condition)
2. Show Progressing if present and true (overrides Applied)
3. Show Available (after passing 1 & 2)
4. Degraded == true, display instead of Available

Or we could look at the size of holding all the conditions, this might be expensive to process as well.

place-work/v1alpha1/types.go
qiujian16: but who will consume this field? It seems to me that having summary field is sufficient, and since we can build a linkage between placementWork and manifestwork, user can get more details from manifestwork. Similar as deployment/pod.

place-work/v1alpha1/types.go
jnpacker: But this approach might be to ambitious.

qiujian16: I would treat these condition as orthogonal. e.g. A resource can be applied=true available=false (because someone on spoke accidentally remove it) or applied=false available=true (because the latest update on the resource fails).

We currently have not implemented progressing and degraded in manifestwork yet...
Progressing would be a transition state. e.g. deployment is rolling out, so actual replica is larger than the desired replica.
Degraded would be a condition that the resource is available, but not fully functioning. e.g. deployment resource exists on the spoke, but the actual replica is smaller than desired replica.

jnpacker: I commented it out as a TODO for now based on your feedback.

jnpacker:
Incorporated @qiujian16's suggestions.

work/v1alpha1/types.go
```diff
@@ -0,0 +1,150 @@
...
+	// 1. AppliedManifestWorks represents ManifestWorks have been distributed as per placement All, Partial, None, Problem
+	// 2. PlacementRefValid
+	Conditions []metav1.Condition `json:"conditions,omitempty"`
+
+	// ManifestWorkDelivery tracks each ManifestWork that was created,
+	// including that it was applied and its overall status
+	PlacedManifestWork []PlacedManifestWork `json:"placedManifestWork,omitempty"`
+
+	// Summary that reflects all relevant ManifestWorks
+	PlacedManifestWorkSummary []PlacedManifestWorkSummary `json:"summary"`
```
serngawy: This one still not clear, why is it as slice of PlacedManifestWorkSummary ? Do we will break down summary for all manifests defined under the manifestWrok ?

place-work/v1alpha1/types.go
serngawy: may be just return the `manifestWork.Status.ResourceStatus.StatusFeedback` If the user defines the feedbackRules in the `ManifestWork.spec` , what do u think ? otherwise I'm more with Jian its too much info that will cause issues than help

work/v1alpha1/types.go
```diff
@@ -0,0 +1,150 @@
...
+	Degraded    int `json:"degraded"`
+	Applied     int `json:"Applied"`
+}
+
+type ManifestWorkStatus string
+
+type PlacedManifestWork struct {
+
+	// Work is an objectReference to the actual ManifestWork resource
+	Work corev1.ObjectReference `json:"workRef"`
```
qiujian16: it is better to use own reference object rather than what in corev1.

work/v1alpha1/types.go
```diff
@@ -0,0 +1,150 @@
...
+
+	// +kubebuilder:default=false
+	Applied bool `json:"applied"`
```
qiujian16: I would have a conditions here, where we can record last observed generation, instead of applied/timestamp

work/v1alpha1/types.go
```diff
@@ -0,0 +1,150 @@
...
+}
+
+// localPlacementReference is the name of a Placement resource in current namespace
+type LocalPlacementReference struct {
+	// Name of the Placement resource in the current namespace
+	Name string `json:"name"`
+}
+
+type PlacedManifestWorkSummary struct {
+	Total       int `json:"total"`
```
qiujian16: still needs doc/comment for each field here. Is it total based on manifest or work?

work/v1alpha1/types.go
jnpacker: How about now?

work/v1alpha1/types.go
jnpacker: Done. Will need to see how this affects the synthetic scale.

work/v1alpha1/types.go
jnpacker: done.

work/v1alpha1/types.go
jnpacker: done.

work/v1alpha1/types.go
```diff
@@ -0,0 +1,158 @@
...
+	// Work is a name & namespace reference to the actual ManifestWork resource
+	Work ManifestWorkReference `json:"workRef"`
+
+	// Condition stores the last known status of the ManifestWork
+	// The ManifestWork generation can be included as well as the last known condition
+	// +optional
+	Condition metav1.Condition `json:"condition,omitempty"`
+
+	// Timestamp is the time when the ManifestWork was last modified
+	Timestamp string `json:"timestamp"`
```
qiujian16: I do not think we still need this timestamp? conditions has sufficient info

work/v1alpha1/types.go
```diff
@@ -0,0 +1,158 @@
...
+// PlacedManifestWork is a minimal status of the ManifestWork created for each targeted clusters, 0..n
+type PlacedManifestWork struct {
+
+	// Work is a name & namespace reference to the actual ManifestWork resource
+	Work ManifestWorkReference `json:"workRef"`
+
+	// Condition stores the last known status of the ManifestWork
+	// The ManifestWork generation can be included as well as the last known condition
+	// +optional
+	Condition metav1.Condition `json:"condition,omitempty"`
```
qiujian16: Conditions []metav1.Condition ?

work/v1alpha1/types.go
```diff
@@ -0,0 +1,158 @@
...
+
+	// Conditions contains the different condition statuses for distrbution of ManifestWork resources
+	// Valid condition types are:
+	// 1. AppliedManifestWorks represents ManifestWorks have been distributed as per placement All, Partial, None, Problem
+	// 2. PlacementRefValid
+	Conditions []metav1.Condition `json:"conditions,omitempty"`
+
+	// ManifestWorkDelivery tracks each ManifestWork that was created,
+	// including that it was applied and its overall status
+	PlacedManifestWork []PlacedManifestWork `json:"placedManifestWork,omitempty"`
```
qiujian16: I am still not sure the intention of this field, should we have a label selector in the PlaceManifestWorkSpec, so the controller always set those labels on the ManifestWork? With that we can always use label selector to find all related manifestworks.

work/v1alpha1/types.go
jnpacker: I only planned to represent one condition that replaces applied, in other words keep the latest applicable condition. Storing multiple conditions by thousands of clusters might likely be a problem as @serngawy suggests.

work/v1alpha1/types.go
jnpacker: Has all the relevant details of the child.
1. Reference
2. Latest sync'd status

So if someone disrupts the label as well to get some basic details of the status.

place-work/v1alpha1/types.go
jnpacker: The StatusFeedback from my perspective is a bigger risk then These details, but I'll remove them for now.

work/v1alpha1/types.go
jnpacker: Removing.

work/v1alpha1/types.go
jnpacker: Removing.

place-work/v1alpha1/types.go
jnpacker: Agreed.

qiujian16:
/approve
seems a good starting point to me. Only a nit comment that we may need to mention in the gdoc that this should be created in the same namespace of placement.

work/v1alpha1/types.go
```diff
@@ -0,0 +1,129 @@
...
+	// Spec reperesents the desired ManifestWork payload and Placement reference to be reconciled
+	Spec PlaceManifestWorkSpec `json:"spec,omitempty"`
+
+	// Status represent the current status of Placing ManifestWork resources
+	Status PlaceManifestWorkStatus `json:"status,omitempty"`
+}
+
+// PlaceManifestWorkSpec defines the desired state of PlaceManifestWork
+type PlaceManifestWorkSpec struct {
+	// ManifestWorkSpec is the ManifestWorkSpec that will be used to generate a per-cluster ManifestWork
```
ivan-cai: ManifestWorkSpec  may be  changed to ManifestWorkTemplate

ivan-cai:
Do we have any plans to support differentiated configurations across clusters? @qiujian16 @jnpacker 

Sometimes，we may be want to distribute some objects like ResourceQuota/LimitRanger to multiple clusters，but the object value is different in different clusters.

qiujian16:
> Do we have any plans to support differentiated configurations across clusters? @qiujian16 @jnpacker
> 
> Sometimes，we may be want to distribute some objects like ResourceQuota/LimitRanger to multiple clusters，but the object value is different in different clusters.

yes, i think there are use cases like:
- value overrides.
- distribution policy. e.g. replicate to all clusters or select one and switch when the original one fails.

I think these could be the followup. WDYT?

ivan-cai:
> > Do we have any plans to support differentiated configurations across clusters? @qiujian16 @jnpacker
> > Sometimes，we may be want to distribute some objects like ResourceQuota/LimitRanger to multiple clusters，but the object value is different in different clusters.
> 
> yes, i think there are use cases like:
> 
> * value overrides.
> * distribution policy. e.g. replicate to all clusters or select one and switch when the original one fails.
> 
> I think these could be the followup. WDYT?

yes, value overrides and distribution policy may be the more important use cases.

work/v1alpha1/types.go
jnpacker: Done.

hack/update-deepcopy.sh
```diff
@@ -10,6 +10,6 @@ verify="${VERIFY:-}"
 GOFLAGS="" bash ${CODEGEN_PKG}/generate-groups.sh "deepcopy" \
   open-cluster-management.io/api/generated \
   open-cluster-management.io/api \
-  "cluster:v1 cluster:v1alpha1 cluster:v1beta1 cluster:v1beta2 work:v1 operator:v1 addon:v1alpha1" \
+  "cluster:v1 cluster:v1alpha1 cluster:v1beta1 work:v1 work:v1alpha1 operator:v1 addon:v1alpha1" \
```
serngawy: I think cluster:v1beta2 removed by mistake, cluster:v1beta2 API still there.

jnpacker: agreed.

mikeshng:
@qiujian16 I think we addressed all the community feedback from Ivan and Mohamed.

Could you please lgtm on this PR so we can get this merged and start the implementation? Thank you.

jnpacker:
> Do we have any plans to support differentiated configurations across clusters?
Definitely something we could consider

qiujian16:
/lgtm
/hold

I think we should also update the enhancement proposal based on this API shape. Also future work about 
1. configuration override
2. distribution strategy 
should be mentioned in the proposal so we do not forget them.

Otherwise feel free to release the hold.

PR #187, opened 2022-11-07
add merge strategy for cma supportconfigs

skeeey:

### DIFF
--- a/addon/v1alpha1/types_clustermanagementaddon.go
+++ b/addon/v1alpha1/types_clustermanagementaddon.go
@@ -48,6 +48,9 @@ type ClusterManagementAddOnSpec struct {
 	// An empty list means the add-on does not require configurations.
 	// The default is an empty list
 	// +optional
+	// +listType=map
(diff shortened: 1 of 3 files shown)


addon/v1alpha1/types_clustermanagementaddon.go
```diff
@@ -93,6 +96,8 @@ type ConfigCoordinates struct {
 // ConfigGroupResource represents the GroupResource of the add-on configuration
 type ConfigGroupResource struct {
 	// group of the add-on configuration.
+	// +required
+	// +kubebuilder:default=""
```
qiujian16, line 100: is this needed? if you set required, default is not necessary, isn't it?

skeeey: mark the group to `option`, but we need to keep `default`, because if one field is marked to `listMapKey` , it must have a default or be a required

qiujian16 (commented):
/approve
/lgtm

PR #196, opened 2022-12-09
Put all common labels/annotations in api repo

qiujian16:
We should treat these labels and annotations as part of APIs


elgnay:
Looks good to me as well.
/unhold

PR #205, opened 2023-02-10
Rename placemanifestwork to manifestworkset

qiujian16:
also change the placementRef to a list of placementRef


serngawy:
sounds good to me, will wait for others to give a review

work/v1alpha1/types_manifestworkset.go
```diff
@@ -120,10 +124,10 @@ const (
 	// PlacementDecisionVerified indicates if Placement is valid
 	//
 	// Reason: AsExpected, PlacementDecisionNotFound, PlacementDecisionEmpty or NotAsExpected
-	PlacementDecisionVerified PlaceWorkConditionType = "PlacementVerified"
+	PlacementDecisionVerified string = "PlacementVerified"
```
serngawy: Better to add "condition" in the naming as the condition type is removed.

work/v1alpha1/types_manifestworkset.go
```diff
@@ -120,10 +124,10 @@ const (
...
 
 	// ManifestWorkApplied confirms that a ManifestWork has been created in each cluster defined by PlacementDecision
 	//
 	// Reason: AsExpected, NotAsExpected or Processing
-	ManifestworkApplied PlaceWorkConditionType = "ManifestworkApplied"
+	ManifestworkApplied string = "ManifestworkApplied"
```
serngawy: same, better to add "condition" in the naming as the condition type is removed.

jnpacker:
I just want to point out that we have made this a little confusing with the use of the word Set.  

So the term Set means: 
ManifestWorkSet - Literally reads as a set of ManifestWork (which it's not)  [Maybe ManifestWorkClusterSet]
ManagedClusterSet - Literally reads as a set of ManagedClusters, which it is
PolicySet - Literally reads as a set of Policies, which it is.

qiujian16:
/hold

I agree set is overly used here.

still need to decide the final name

ManifestWorkClusterSet?
ManifestWorkReplicaSet?
ManifestDeployment?

serngawy:
> 
could be ManifestWorkReplica

serngawy:
We need to finish this and agree on naming (There are a pause/re-do work will happen after);  My vote is for
- ManifestWorkReplicaSets 
@qiujian16 @jnpacker

jnpacker:
Actually I think your onto something @serngawy , but using the term replica is confusing compared to a Kube ReplicaSet.

How about ManifestWorkPlacementSet, it's a little long, but we can then use the definition from Kube ReplicaSet (I've replaced ReplicaSet and Pods with `ManifestWorkPlacementSet` and `ManifestWorks`:

A `ManifestWorkPlacementSet` ~~Replicaset~~ is defined with fields, including a selector that specifies how to identify `Clusters` ~~Pods~~ it can acquire `(via Placement)`, ~~a number of replicas indicating how many Pods it should be maintaining,~~ and a `ManifestWork` ~~pod~~ template specifying the data of new `ManifestWorks` ~~Pods~~ it should create to meet the `Placement` ~~number of replicas~~ criteria. A `ManifestWorkPlacementSet` ~~ReplicaSet~~ then fulfills its purpose by creating and deleting `ManifestWorks` ~~Pods~~ as needed to reach the desired `Placement` ~~number~~. When a `ManifestWorkPlacementSet` ~~ReplicaSet~~ needs to create new `ManifestWorks` ~~Pods~~, it uses its `ManifestWork` ~~Pod~~ template.

serngawy:
Well,  ManifestWrokReplicaSet is a replicated manifesWork defined by the ManifestWorkTemplate and the Set defined by the placement. So as k8s user familiar with ReplicaSet term  I can relate that manifestWorkReplicaSet is a Set of replicated manifestWork.  I feel ManifetWorkReplicaSet make more sense but lets see other folks opinion.

jnpacker:
@qiujian16 Either ManifestWorkReplicaSet or ManifestWorkPlacementSet works for me.

qiujian16:
I would vote for ManifestWorkReplicaSet, will update this PR soon.

serngawy (approved):

ivan-cai:
I would vote for  ManifestWorkReplicaSet or ManifestDeployment

PR #206, opened 2023-02-10
Set omitempty on the ManifestWork UpdateStrategy field

mprahl:
When a ManifestWork without UpdateStrategy set is marshalled into JSON, it causes the JSON value of "updateStrategy": null. This works on newer versions of Kubernetes, however, on older versions such as v1.19.16, the following error is returned when creating a ManifestWork: The ManifestWork "addon-config-policy-controller-pre-delete-hosting-cluster2" is invalid: spec.manifestConfigs.updateStrategy: Invalid value: "null": spec.manifestConfigs.updateStrategy in body must be of type object: "null"

The best way to avoid this is to add omitempty to the JSON tag so that it's completely not present in the JSON representation.

This bug was encountered when using a predelete hook in the addon framework. So this change will need to be updated in at least the addon framework and the work agent.

Relates:
https://issues.redhat.com/browse/ACM-3233
https://issues.redhat.com/browse/ACM-2923

Signed-off-by: mprahl <mprahl@users.noreply.github.com>

### DIFF
--- a/work/v1/types.go
+++ b/work/v1/types.go
@@ -92,9 +92,9 @@ type ManifestConfigOption struct {
 	FeedbackRules []FeedbackRule `json:"feedbackRules,omitempty"`
 
 	// UpdateStrategy defines the strategy to update this manifest. UpdateStrategy is Update
-	// if it is not set,
-	// optional
-	UpdateStrategy *UpdateStrategy `json:"updateStrategy"`
+	// if it is not set.
+	// +optional
+	UpdateStrategy *UpdateStrategy `json:"updateStrategy,omitempty"`
 }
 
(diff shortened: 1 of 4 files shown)


work/v1/types.go
```diff
@@ -94,7 +94,7 @@ type ManifestConfigOption struct {
 	// UpdateStrategy defines the strategy to update this manifest. UpdateStrategy is Update
 	// if it is not set,
 	// optional
-	UpdateStrategy *UpdateStrategy `json:"updateStrategy"`
+	UpdateStrategy *UpdateStrategy `json:"updateStrategy,omitempty"`
```
JustinKuli: Should this also be `// +optional`?

mprahl: Good point. It doesn't change the behavior,  but it's good to be consistent with the other fields.

PR #208, opened 2023-02-15
addon rollout strategy

haoqing0110:
Ref: https://github.com/open-cluster-management-io/OCM/issues/114


addon/v1alpha1/types_clustermanagementaddon.go
```diff
@@ -175,10 +185,119 @@ type PlacementStrategy struct {
...
+
+	// desiredConfigSpecHash record the desired config spec hash.
+	DesiredConfigSpecHash *ConfigSpecHash `json:"desiredConfigSpecHash"`
+
+	// lastKnownGoodConfigSpecHash record the last known good config spec hash.
+	// For fresh install or rollout with type UpdateAll or RollingUpdate, the
+	// lastKnownGoodConfigSpecHash is the same as lastAppliedConfigSpecHash.
+	// For rollout with type RollingUpdateWithCanary, the lastKnownGoodConfigSpecHash
+	// is the last successfully applied config spec hash of the canary placement.
+	LastKnownGoodConfigSpecHash *ConfigSpecHash `json:"lastKnownGoodConfigSpecHash"`
```
haoqing0110: The name `lastKnownGood` is controversial, might need a better name.

addon/v1alpha1/types_addonhubconfig.go
```diff
@@ -0,0 +1,50 @@
+package v1alpha1
```
qiujian16: for addonhubconfig, I think we should put in a separated PR

addon/v1alpha1/types_clustermanagementaddon.go
```diff
@@ -175,10 +185,145 @@ type PlacementStrategy struct {
...
+}
+
+// InstallConfigReference is a reference to the current add-on configuration.
+// This resource is used to record the configuration resource for the current add-on.
+type InstallConfigReference struct {
+	// This field is synced from ClusterManagementAddOn Configurations.
+	ConfigGroupResource `json:",inline"`
+
+	// desiredConfigSpecHash record the desired config spec hash.
+	DesiredConfigSpecHash *ConfigSpecHash `json:"desiredConfigSpecHash"`
```
qiujian16: it seems to me the name should be desiredConfig rather than desiredConfigSpecHash, and in the desiredConfig, we have specHash/name/namespace

addon/v1alpha1/types_clustermanagementaddon.go
```diff
@@ -127,6 +128,15 @@ type ConfigReferent struct {
 	Name string `json:"name"`
 }
 
+// ConfigSpecHash represents the namespace,name and spec hash for an add-on configuration.
+type ConfigSpecHash struct {
+	// namespace and name for an add-on configuration.
+	ConfigReferent `json:",inline"`
+
+	// spec hash for an add-on configuration.
+	ConfigSpecHash string `json:"configSpecHash"`
```
zhujian7: use `SpecHash` directly?

addon/v1alpha1/types_clustermanagementaddon.go
```diff
@@ -175,10 +185,145 @@ type PlacementStrategy struct {
 	// User can override the configuration by updating the managedClusterAddon directly.
 	// +optional
 	Configs []AddOnConfig `json:"configs,omitempty"`
+	// The rollout strategy to apply addon configurations change.
+	// The rollout strategy only watches the addon configurations defined in ClusterManagementAddOn.
+	// +optional
+	RolloutStrategy RolloutStrategy `json:"rolloutStrategy,omitempty"`
```
zhujian7, line 192: should this be a pointer？

addon/v1alpha1/types_clustermanagementaddon.go
zhujian7: since it is only useful for `RollingUpdateWithCanary`, how about something like `LastCanaryAppliedConfig`? it is more accurate but not scalable.

addon/v1alpha1/types_clustermanagementaddon.go
skeeey: default `UpdateAll`?

addon/v1alpha1/types_clustermanagementaddon.go
```diff
@@ -175,10 +185,145 @@ type PlacementStrategy struct {
...
+	// The maximum concurrently updating number of addons.
+	// Value can be an absolute number (ex: 5) or a percentage of desired addons (ex: 10%).
+	// Absolute number is calculated from percentage by rounding up.
+	// Defaults to 25%.
+	// Example: when this is set to 30%, once the addon configs change, the addon on 30% of the selected clusters
+	// will adopt the new configs. When the addons with new configs are healthy, the addon on the remaining clusters
+	// will be further updated.
+	// +kubebuilder:default:="25%"
+	// +optional
+	MaxConcurrency *intstr.IntOrString `json:"maxConcurrency,omitempty"`
```
skeeey: this has default value, do we need a pointer?

addon/v1alpha1/types_clustermanagementaddon.go
haoqing0110: Good suggestion!

addon/v1alpha1/types_addonhubconfig.go
haoqing0110: removed from this PR

addon/v1alpha1/types_clustermanagementaddon.go
haoqing0110: add `	// +kubebuilder:default={type: UpdateAll}`

addon/v1alpha1/types_clustermanagementaddon.go
haoqing0110: remove pointer

haoqing0110: add the pointer back, since when it's a structure, won't generate a default value.

addon/v1alpha1/types_managedclusteraddon.go
```diff
@@ -199,6 +202,14 @@ type ConfigReference struct {
 
 	// lastObservedGeneration is the observed generation of the add-on configuration.
 	LastObservedGeneration int64 `json:"lastObservedGeneration"`
```
zhujian7, line 191: Will the `LastObservedGeneration` be replaced by the `LastAppliedConfig` in the future？ should we mark it as deprecated？

addon/v1alpha1/types_clustermanagementaddon.go
```diff
@@ -175,10 +185,146 @@ type PlacementStrategy struct {
 	// User can override the configuration by updating the managedClusterAddon directly.
 	// +optional
 	Configs []AddOnConfig `json:"configs,omitempty"`
+	// The rollout strategy to apply addon configurations change.
+	// The rollout strategy only watches the addon configurations defined in ClusterManagementAddOn.
+	// +kubebuilder:default={type: UpdateAll}
+	// +optional
+	RolloutStrategy *RolloutStrategy `json:"rolloutStrategy,omitempty"`
```
qiujian16: why pointer? I am starting to think it should not be a pointer, InstallStrategy should not be pointer either

addon/v1alpha1/types_clustermanagementaddon.go
qiujian16: I think it should. it just cannot get default value if you create via client

addon/v1alpha1/types_clustermanagementaddon.go
haoqing0110: The `kubebuilder:default={type: UpdateAll}` won't work when RolloutStrategy is a structure, like this integration testing case shows https://github.com/open-cluster-management-io/api/pull/208/files#diff-0f15edf563bb9b75e4f6b5934e42dffcc3db1c80238ae2970cafce38a8959c54

addon/v1alpha1/types_clustermanagementaddon.go
haoqing0110: @qiujian16 Do we want to ignore the case when creating via client? and also remove the integration testing case?

addon/v1alpha1/types_clustermanagementaddon.go
qiujian16: yeah, that is because when you use client to create the resource, this field is init by default. But when you use yaml without setting this field, it should still work

haoqing0110: Set both InstallStrategy and RolloutStrategy to structure, and modified some integration tests.

addon/v1alpha1/types_clustermanagementaddon.go
```diff
@@ -175,10 +185,146 @@ type PlacementStrategy struct {
...
+	// +kubebuilder:default={type: UpdateAll}
+	// +optional
+	RolloutStrategy RolloutStrategy `json:"rolloutStrategy,omitempty"`
+}
+
+// RolloutStrategy represents the rollout strategy of the add-on configuration.
+type RolloutStrategy struct {
+	// Type is the type of the rollout strategy, it supports UpdateAll, RollingUpdate and RollingUpdateWithCanary:
+	// - UpdateAll: when configs change, apply the new configs to all the selected clusters at once.
+	//   This is the default strategy.
```
deads2k, line 199: Just to confirm (no need to update the doc), this is the default for backwards compatibility, correct?  If so, would we consider changing the default to RollingUpdate in beta?

addon/v1alpha1/types_clustermanagementaddon.go
```diff
@@ -175,10 +185,146 @@ type PlacementStrategy struct {
...
+	//
+	//   The canary placement does not have to be a subset of the install placement, and it is more like a
+	//   reference for finding and checking canary clusters before upgrading all. To trigger the rollout
+	//   on the canary clusters, you can define another rollout strategy with the type RollingUpdate, or even
+	//   manually upgrade the addons on those clusters.
+	//
+	// +kubebuilder:validation:Enum=UpdateAll;RollingUpdate;RollingUpdateWithCanary
+	// +kubebuilder:default:=UpdateAll
+	// +optional
+	Type string `json:"type"`
```
deads2k, line 224: excellent doc on this field.

addon/v1alpha1/types_clustermanagementaddon.go
```diff
@@ -175,10 +185,146 @@ type PlacementStrategy struct {
...
+	// AddonRolloutStrategyRollingUpdate is the addon rollout strategy representing wait and check
+	// if add-ons on the canary have applied the new configs, then apply the new configs to
+	// all the selected clusters with the concurrency rate.
+	AddonRolloutStrategyRollingUpdateWithCanary string = "RollingUpdateWithCanary"
+)
+
+// RollingUpdate represents the behavior to rolling update add-on configurations
+// on the selected clusters.
+type RollingUpdate struct {
+	// The maximum concurrently updating number of addons.
```
deads2k: maximum concurrently updating number of clusters?  This struct is for a single add-on on multiple managedclusters, correct?

addon/v1alpha1/types_clustermanagementaddon.go
```diff
@@ -175,10 +185,146 @@ type PlacementStrategy struct {
...
+	// +required
+	Placement PlacementRef `json:"placement,omitempty"`
+
+	// the behavior to rolling update add-on configurations.
+	RollingUpdate `json:",inline"`
 }
 
 // ClusterManagementAddOnStatus represents the current status of cluster management add-on.
 type ClusterManagementAddOnStatus struct {
+	// defaultconfigReferences is a list of current add-on default configuration references.
```
deads2k, line 277: Are these the configurations that will be applied if a new managedcluster appears in the placement?

deads2k:
a few minor questions.  I like the shape

addon/v1alpha1/types_clustermanagementaddon.go
haoqing0110: Yes, it's for backwards compatibility. Good suggestion, we could consider using RollingUpdate as default in beta.

addon/v1alpha1/types_clustermanagementaddon.go
haoqing0110: yes, it's for a single add-on on multiple managedclusters. Update the comments.

addon/v1alpha1/types_clustermanagementaddon.go
haoqing0110: yes, the ClusterManagementAddOnStatus records the configurations with spec hash, a new added managedcluster will use these configs.

qiujian16:
/approve
/lgtm
/hold
looks good to me. hold in case any additional review is needed.

haoqing0110:
/unhold
Feel free to leave more comments if you have an interest in this API.

PR #213, opened 2023-02-27
add addon condition types and reasons

zhiweiyin318:


zhiweiyin318:
some condition types are defined in API repo, others are defined in addon-framework repo, so put them together in API repo with the reasons .

/assign @zhujian7 
/assign @qiujian16

addon/v1alpha1/types_managedclusteraddon.go
```diff
@@ -280,7 +268,134 @@ const (
...
+	// on hosting cluster.
+	AddonHostingPreDeleteHookFinalizer = "addon.open-cluster-management.io/hosting-addon-pre-delete"
+	// AddonHostingManifestFinalizer is the finalizer for an addon which has deployed manifests on the external
+	// hosting cluster in Hosted mode.
+	AddonHostingManifestFinalizer = "addon.open-cluster-management.io/hosting-manifests-cleanup"
+)
+
+const (
+	// AddonInstallModeBuiltinValueKey is the key of the build in value to represent the addon install mode, addon developers
+	// can use this built in value in manifests.
```
qiujian16: Is the value they put in helm chart? I do not think we should put this in api repo if it is not set in any resources.

addon/v1alpha1/types_managedclusteraddon.go
```diff
@@ -280,7 +268,134 @@ const (
...
-	// HostedManifestLocationNoneValue is a value of the annotation HostedManifestLocationAnnotationKey,,
+	// HostedManifestLocationNoneValue is a value of the annotation HostedManifestLocationAnnotationKey,
 	// indicates the manifest will not be deployed in Hosted mode.
 	HostedManifestLocationNoneValue = "none"
+
+	// Finalizers on the managedClusterAddon.
+
+	// AddonDeprecatedPreDeleteHookFinalizer is the finalizer for an addon which has deployed hook objects.
+	// Deprecated: please use the AddonPreDeleteHookFinalizer instead.
+	AddonDeprecatedPreDeleteHookFinalizer = "cluster.open-cluster-management.io/addon-pre-delete"
```
qiujian16, line 279: why we add deprecated?

zhiweiyin318: I want to use `addon.open-cluster-management.io/addon-pre-delete` instead of `cluster.open-cluster-management.io/addon-pre-delete`

addon/v1alpha1/types_managedclusteraddon.go
zhiweiyin318: yes, removed.

addon/v1alpha1/types_managedclusteraddon.go
qiujian16: do we need to put deprecated in the api repo given they are deprecated?

PR #217, opened 2023-03-08
add pattern for clustername

ycyaoxdu:
related issue: https://github.com/open-cluster-management-io/clusteradm/issues/318


ycyaoxdu:
/assign @qiujian16 
cc @mikeshng

qiujian16:
not quite sure if it is breaking. We need a discussion on this
Also since it needs to be comply with namespace format, we might also set maxLength to 63?

qiujian16:
@ycyaoxdu would you rebase the PR I think we can merge this one

PR #227, opened 2023-03-22
upgrade sets to use generic and add cilint

ycyaoxdu:
add golangci-lint and upgrade sets.String to sets.Set[string]

### DIFF
--- a/utils/work/v1/workbuilder/workbuilder_test.go
+++ b/utils/work/v1/workbuilder/workbuilder_test.go
@@ -10,10 +10,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
-	"k8s.io/client-go/kubernetes/scheme"
 	workapiv1 "open-cluster-management.io/api/work/v1"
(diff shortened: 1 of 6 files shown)


utils/work/v1/workbuilder/workbuilder_test.go
```diff
@@ -105,16 +102,16 @@ func newFakeManifest(object runtime.Object) workapiv1.Manifest {
...
 
-func newFakeWorkClient(works []workapiv1.ManifestWork) client.Client {
-	s := scheme.Scheme
-	s.AddKnownTypes(workapiv1.GroupVersion, &workapiv1.ManifestWork{}, &workapiv1.ManifestWorkList{})
-	var objects []runtime.Object
-	for i := 0; i < len(works); i++ {
-		objects = append(objects, &works[i])
-	}
-	return fake.NewClientBuilder().WithScheme(s).WithRuntimeObjects(objects...).Build()
+// func newFakeWorkClient(works []workapiv1.ManifestWork) client.Client {
```
qiujian16: why we comment this?

ycyaoxdu: it is not used, not sure if we need to remove this function? @qiujian16

qiujian16: yes we should remove it if it is not used.

PR #228, opened 2023-03-29
Remove Enable field in ClusterManager API

qiujian16:


operator/v1/types.go
```diff
@@ -98,14 +97,6 @@ type WorkConfiguration struct {
...
 
 type AddOnManagerConfiguration struct {
-	// Mode is either Enable, Disable, "" where "" is Disable by default.
-	// In Enable mode, the component will be installed.
-	// In Disable mode, the component will not be installed.
-	// +kubebuilder:default:=Disable
-	// +kubebuilder:validation:Enum:=Enable;Disable
-	// +optional
-	Mode ComponentModeType `json:"mode,omitempty"`
-
```
haoqing0110, line 108: https://github.com/open-cluster-management-io/api/blob/0fa9a2459a0e0222145d8b352acc0c0dbb49fa60/operator/v1/types.go#L111-L117 is also added to enable addon manager, it is still necessary to keep？

qiujian16: removed

PR #232, opened 2023-04-17
update work annotations

haoqing0110:


utils/work/v1/workapplier/workapplier.go
```diff
@@ -62,11 +62,14 @@ func (w *WorkApplier) Apply(ctx context.Context, work *workapiv1.ManifestWork) (
...
 	}
 
-	if ManifestWorkSpecEqual(work.Spec, existingWork.Spec) {
+	if ManifestWorkEqual(work, existingWork) {
 		return existingWork, nil
 	}
 
 	oldData, err := json.Marshal(&workapiv1.ManifestWork{
+		ObjectMeta: metav1.ObjectMeta{
+			Annotations: existingWork.Annotations,
```
qiujian16, line 71: should we also compare labels and ownerrefs?

haoqing0110: @qiujian16 I don't see the labels and ownerrefs will change during update. Not sure if I miss some cases and need to compare them?

PR #233, opened 2023-04-17
fix work hash confilict and compare mainfests

zhiweiyin318:


utils/work/v1/workapplier/workapplier.go
```diff
@@ -121,7 +126,33 @@ func manifestsEqual(new, old []workapiv1.Manifest) bool {
...
+}
+
+func manifestUnstructured(manifest workapiv1.Manifest) *unstructured.Unstructured {
+	newManifest := &unstructured.Unstructured{}
+	if err := newManifest.UnmarshalJSON(manifest.Raw); err != nil {
+		klog.Errorf("failed to unmarshal new manifest. err:%v", err)
+		return newManifest
+	}
+
+	unstructured.RemoveNestedField(newManifest.Object, "metadata", "creationTimestamp")
```
qiujian16: need to add some comments here

utils/work/v1/workapplier/workapplier.go
```diff
@@ -121,7 +126,33 @@ func manifestsEqual(new, old []workapiv1.Manifest) bool {
...
+	unstructured.RemoveNestedField(newManifest.Object, "spec", "template", "metadata", "creationTimestamp")
+
+	return newManifest
+}
+
+// mapExisted is used to check if required is included in existing.
+func mapExisted(required, existing map[string]string) bool {
```
qiujian16: I think it is more like `shouldUpdateMap` ?

utils/work/v1/workapplier/workapplier.go
zhiweiyin318: added

utils/work/v1/workapplier/workapplier.go
zhiweiyin318: yes, changed the func name.

zhiweiyin318:
@qiujian16   please review again.  addressed the comments we discussed.

PR #235, opened 2023-04-26
Return json raw string in feedback result

qiujian16:


qiujian16:
https://github.com/open-cluster-management-io/work/issues/196

PR #239, opened 2023-05-18
Add addon template api

zhujian7:


addon/v1alpha1/types_addontemplate.go
```diff
@@ -0,0 +1,181 @@
...
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+// +kubebuilder:resource:scope="Cluster"
+// +kubebuilder:printcolumn:name="ADDON NAME",type=string,JSONPath=`.spec.addonName`
+
+// AddOnTemplate is the Custom Resource object, it is used to describe
+// how to deploy the addon agent and how to register the addon.
+//
+// AddOnTemplate is a cluster-scoped resource, and will only be used
+// on the hub cluster.
+type AddOnTemplate struct {
```
qiujian16, line 21: ManagedClusterAddOnTemplate?

addon/v1alpha1/types_addontemplate.go
```diff
@@ -0,0 +1,181 @@
...
+	// +optional
+	HubPermissions []HubPermissionConfig `json:"hubPermissions,omitempty"`
+}
+
+// HubPermissionConfig configures the permission of the addon agent to access the hub cluster.
+// Will create a RoleBinding in the same namespace as the managedClusterAddon to bind the user
+// provided ClusterRole/Role to the "system:open-cluster-management:cluster:<cluster-name>:addon:<addon-name>"
+// Group.
+type HubPermissionConfig struct {
+	// Type of the permissions setting. It defines how to bind the roleRef
```
qiujian16: we need to describe the meaning of CurrentCluster and singlenamespace

qiujian16 (commented):
/approve

two minor comments

addon/v1alpha1/types_addontemplate.go
zhujian7: described [here](https://github.com/open-cluster-management-io/api/pull/239/files#diff-cc2eb9f759617220330a98f0609b906346b9f986ccaf027c4449db17fbf1f4dcR66-R72), should I move them into the `HubPermissionConfig` struct？

addon/v1alpha1/types_addontemplate.go
zhujian7: should it start with "addon" to be consistent with addonDeploymentConfig?

qiujian16: hrm, that makes sense also

addon/v1alpha1/types_addontemplate.go
qiujian16: I think we should for generating the openapi doc

zhujian7: description added

PR #242, opened 2023-05-25
Add placement DecisionStrategy API

serngawy:


cluster/v1beta1/types_placement.go
```diff
@@ -94,6 +94,42 @@ type PlacementSpec struct {
...
+	// +optional
+	DecisionStrategy DecisionStrategy `json:"decisionStrategy,omitempty"`
+}
+
+// Subset of the created placementDecision with label.
+type DecisionGroup struct {
+	// Label to be added to the created placement Decisions
+	// +kubebuilder:validation:Required
+	// +required
+	Label string `json:"labels,omitempty"`
```
qiujian16: what does the value of this field look like

cluster/v1beta1/types_placement.go
```diff
@@ -329,11 +365,35 @@ const (
...
 
+// Present decision groups status based on the DecisionStrategy definition.
+type DecisionGroupStatus struct {
+	// Present the decision group index. If there is no decision strategy defined all placement decisions will be in group index 0
+	// +optional
+	DecisionGroupIndex int32 `json:"decisionGroupIndex"`
+
+	// Decision group label that is defined in the DecisionStrategy's DecisionGroup.
+	// +optional
+	DecisionGroupLabel string `json:"decisionGroupLabel"`
```
qiujian16: shouldn't this contain key/value?

cluster/v1beta1/types_placement.go
```diff
@@ -94,6 +94,42 @@ type PlacementSpec struct {
...
+	// Define subsets of the created placement decisions and add labels to the placementDecision.
+	// +optional
+	DecisionGroups []DecisionGroup `json:"decisionGroups,omitempty"`
+
+	// ClusterPercentagePerDecisionGroup is a percentage of the total selected clusters that will divide the clusters to groups.
+	// ex; for a total 100 clusters selected, ClusterPercentagePerDecisionGroup equal to 20% will divide the placement decision to 5 groups each group should have 20 clusters.
+	// +kubebuilder:validation:Enum=None;"5%";"10%";"15%";"20%";"25%";"30%";"50%";"100%"
+	// +kubebuilder:default:=None
+	// +optional
+	ClusterPercentagePerDecisionGroup string `json:"clusterPercentagePerDecisionGroup,omitempty"`
```
qiujian16: It is actually MaxCountPerGroup right? I still think we should only have one field that can be set as IntOrString

cluster/v1beta1/types_placementdecision.go
```diff
@@ -28,7 +28,12 @@ type PlacementDecision struct {
 
 // The placementDecsion label name holding the placement name
 const (
+	// Placement owner name.
 	PlacementLabel string = "cluster.open-cluster-management.io/placement"
+	// decision group index.
+	DecisionGroupIndexLabel string = "cluster.open-cluster-management.io/decisiongroupindex"
+	// decision group label.
+	DecisionGroupLabel string = "cluster.open-cluster-management.io/decisiongrouplabel"
```
qiujian16: I think we should allow user to customize the key/value instead of a fixe key

cluster/v1beta1/types_placement.go
serngawy: this is label value so the placementDecision will have label key as "cluster.open-cluster-management.io/decisiongrouplabel" and the user define here the value

cluster/v1beta1/types_placement.go
serngawy: okay, I wasn't sure that we can define enum and int for type IntOrString, will push a fix for it.

cluster/v1beta1/types_placement.go
serngawy: the key is API constant "cluster.open-cluster-management.io/decisiongrouplabel"

cluster/v1beta1/types_placementdecision.go
serngawy: better to set the key as OCM API constant so controllers can identify it.

cluster/v1beta1/types_placement.go
serngawy: I see, label may cause confusion. May be better to change the field to name or groupName as we don't want the user to add multiple labels.

cluster/v1beta1/types_placementdecision.go
```diff
@@ -28,7 +28,12 @@ type PlacementDecision struct {
 
 // The placementDecsion label name holding the placement name
 const (
+	// Placement owner name.
 	PlacementLabel string = "cluster.open-cluster-management.io/placement"
+	// decision group index.
+	DecisionGroupIndexLabel string = "cluster.open-cluster-management.io/decisiongroupindex"
```
qiujian16:
```suggestion
	DecisionGroupIndexLabel string = "cluster.open-cluster-management.io/decision-group-index"
```

cluster/v1beta1/types_placementdecision.go
```diff
@@ -28,7 +28,12 @@ type PlacementDecision struct {
...
+	// decision group label.
+	DecisionGroupName string = "cluster.open-cluster-management.io/decisiongroupname"
```
qiujian16:
```suggestion
	DecisionGroupName string = "cluster.open-cluster-management.io/decision-group-name"
```

cluster/v1beta1/types_placement.go
qiujian16: https://github.com/open-cluster-management-io/api/blob/main/addon/v1alpha1/types_clustermanagementaddon.go#L260

cluster/v1beta1/types_placement.go
```diff
@@ -94,6 +94,39 @@ type PlacementSpec struct {
...
+	Name string `json:"name,omitempty"`
+
+	// LabelSelector to select clusters subset by label.
+	// +kubebuilder:validation:Required
+	// +required
+	ClusterSelector ClusterSelector `json:"groupClusterSelector,omitempty"`
+}
+
+// Group the created placementDecision into decision groups based on the number of clusters per decision group.
+type DecisionStrategy struct {
```
qiujian16: I would think this like 

```
type DecisionStrategy struct {
    GroupStrategy GroupStrategy
}

type GroupStrategy struct {
    DecisionGroups []DecisionGroup `json:"decisionGroups,omitempty"`
    ClustersPerDecisionGroup string `json:"clustersPerDecisionGroup,omitempty"`
}
```

cluster/v1beta1/types_placement.go
```diff
@@ -94,6 +95,40 @@ type PlacementSpec struct {
 	// certain taints to be selected by placements with matching tolerations.
 	// +optional
 	Tolerations []Toleration `json:"tolerations,omitempty"`
+
+	// Decision Strategy divide the created placement decision to groups and define number of clusters per decision group.
+	// +optional
+	DecisionStrategy DecisionStrategy `json:"decisionStrategy,omitempty"`
+}
+
+// Subset of the created placementDecision with label.
```
qiujian16: we should add this the clusters will be put into the decisions with a decision group label .....

cluster/v1beta1/types_placement.go
```diff
@@ -94,6 +95,40 @@ type PlacementSpec struct {
...
+type DecisionGroup struct {
+	// name to be added to the created placement Decisions labels with label key cluster.open-cluster-management.io/decisiongroupname
+	// +kubebuilder:validation:Required
+	// +required
+	Name string `json:"name,omitempty"`
```
qiujian16:
```suggestion
	GroupLabelValue string `json:"groupLabelValue,omitempty"`
```

cluster/v1beta1/types_placement.go
```diff
@@ -94,6 +95,40 @@ type PlacementSpec struct {
 	// certain taints to be selected by placements with matching tolerations.
 	// +optional
 	Tolerations []Toleration `json:"tolerations,omitempty"`
+
+	// Decision Strategy divide the created placement decision to groups and define number of clusters per decision group.
```
qiujian16:
```suggestion
	// Decision Strategy divide the created placement decision to groups and define number of clusters per decision group. Decision result will be sorted by name and filled into decisions if it is not set.
```

cluster/v1beta1/types_placement.go
```diff
@@ -94,6 +95,40 @@ type PlacementSpec struct {
...
+
+	// LabelSelector to select clusters subset by label.
+	// +kubebuilder:validation:Required
+	// +required
+	ClusterSelector ClusterSelector `json:"groupClusterSelector,omitempty"`
+}
+
+// Group the created placementDecision into decision groups based on the number of clusters per decision group.
+type DecisionStrategy struct {
+	// Define subsets of the created placement decisions and add labels to the placementDecision.
```
qiujian16:
```suggestion
	// DecisionGroups represents a list of predefined groups to put decision results. 
```

cluster/v1beta1/types_placement.go
```diff
@@ -94,6 +95,40 @@ type PlacementSpec struct {
...
+	// +optional
+	DecisionGroups []DecisionGroup `json:"decisionGroups,omitempty"`
+
+	// ClustersPerDecisionGroup is a specific number or percentage of the total selected clusters.
+	// The specific number will divide the placementDecisions to decisionGroups each group has max number of clusters equal to that specific number.
+	// The percentage will divide the placementDecisions to decisionGroups each group has max number of clusters based on the total num of selected clusters and percentage.
+	// ex; for a total 100 clusters selected, ClustersPerDecisionGroup equal to 20% will divide the placement decision to 5 groups each group should have 20 clusters.
```
qiujian16, line 127: should clarify that If DecisionGroups is set, the matched decision result will be put into them at first. The left decision results will be put into the decision groups based on this field

cluster/v1beta1/types_placement.go
```diff
@@ -329,11 +364,35 @@ const (
...
 
+// Present decision groups status based on the DecisionStrategy definition.
+type DecisionGroupStatus struct {
+	// Present the decision group index. If there is no decision strategy defined all placement decisions will be in group index 0
+	// +optional
+	DecisionGroupIndex int32 `json:"decisionGroupIndex"`
+
+	// Decision group name that is defined in the DecisionStrategy's DecisionGroup.
+	// +optional
+	DecisionGroupName string `json:"decisionGroupName"`
```
qiujian16, line 384:
```suggestion
	GroupLabelValue string `json:"groupLableValue"`
```

cluster/v1beta1/types_placement.go
```diff
@@ -329,11 +364,35 @@ const (
...
+
+	// List of placement decisions names associated with the decision group
+	// +optional
+	PlacementDecisions []string `json:"placementDecisions"`
```
qiujian16: Decisions

cluster/v1beta1/types_placement.go
serngawy: done

cluster/v1beta1/types_placement.go
serngawy: done

cluster/v1beta1/types_placement.go
serngawy: done

cluster/v1beta1/types_placement.go
serngawy: done

cluster/v1beta1/types_placementdecision.go
serngawy: done

cluster/v1beta1/types_placementdecision.go
serngawy: kind of confusing the label key is decision-group-name not decision-group-labelvalue ? if yes, better to keep the fields up as group Name then ?

cluster/v1beta1/types_placement.go
serngawy: done

cluster/v1beta1/types_placementdecision.go
qiujian16: the label key is decision-group-name, while what you want user to set is the label value of this key. Group name is not clear since it may imply this is a resource name or a certain field, but it is actually a label value of the certain key

cluster/v1beta1/types_placement.go
serngawy: Do we need to have nested struct ? I believe decisionStrategy as single struct suppose to be good unless we will extended with something related.

cluster/v1beta1/types_placement.go
serngawy: I change it above to groupName instead of name/labelValue and mention that in the CRD definition  [L106](https://github.com/open-cluster-management-io/api/pull/242/files#diff-40673a7be25e1817d0abcd7d4c2c0706d9398d31b65c1c62ae5ce2f6b9521cfeR106)

serngawy (commented):
I added the rolloutStrategy struct to the helper class as per discussion with ACM teams. The rolloutStrategy struct will be part and consumed by other APIs like manifestWorkReplicatSet, policy, Addon,...etc

cluster/v1beta1/types_placement.go
qiujian16: nested will help us to extend in the future

cluster/v1beta1/types_placement.go
```diff
@@ -94,6 +95,41 @@ type PlacementSpec struct {
...
+	// +optional
+	DecisionGroups []DecisionGroup `json:"decisionGroups,omitempty"`
+
+	// ClustersPerDecisionGroup is a specific number or percentage of the total selected clusters.
+	// The specific number will divide the placementDecisions to decisionGroups each group has max number of clusters equal to that specific number.
+	// The percentage will divide the placementDecisions to decisionGroups each group has max number of clusters based on the total num of selected clusters and percentage.
+	// ex; for a total 100 clusters selected, ClustersPerDecisionGroup equal to 20% will divide the placement decision to 5 groups each group should have 20 clusters.
+	// Default is having all clusters in a single group.
+	// If the DecisionGroups field defined, it will be considered first to create the decisionGroups then the ClustersPerDecisionGroup will be used to determine the rest of decisionGroups.
+	// +kubebuilder:validation:Enum="5%";"10%";"15%";"20%";"25%";"30%";"40%";"50%";"100%"
```
qiujian16: do not use enum here. you can have a regex to validate

cluster/v1beta1/types_placement.go
```diff
@@ -94,6 +95,41 @@ type PlacementSpec struct {
...
+	// +optional
+	DecisionStrategy DecisionStrategy `json:"decisionStrategy,omitempty"`
+}
+
+// DecisionGroup define a subset of clusters that will be added to placementDecisions with groupName label.
+type DecisionGroup struct {
+	// Group name to be added to the created placement Decisions labels with label key cluster.open-cluster-management.io/decision-group-name
+	// +kubebuilder:validation:Required
+	// +required
+	GroupName string `json:"groupName,omitempty"`
```
qiujian16, line 110: add regex validation rule here, since it is a label value.

cluster/v1beta1/types_placement.go
```diff
@@ -94,6 +95,41 @@ type PlacementSpec struct {
...
 	Tolerations []Toleration `json:"tolerations,omitempty"`
+
+	// DecisionStrategy divide the created placement decision to groups and define number of clusters per decision group. Decision result will be sorted by name and filled into decisions if it is not set.
+	// +optional
+	DecisionStrategy DecisionStrategy `json:"decisionStrategy,omitempty"`
+}
+
+// DecisionGroup define a subset of clusters that will be added to placementDecisions with groupName label.
+type DecisionGroup struct {
+	// Group name to be added to the created placement Decisions labels with label key cluster.open-cluster-management.io/decision-group-name
```
qiujian16: Group name to be added as the value of the created PlacementDecisions labels with key ...

serngawy: done

cluster/v1beta1/types_placement.go
serngawy: done

cluster/v1beta1/types_placement.go
serngawy: done

cluster/v1beta1/types_placement.go
serngawy: done

qiujian16:
/approve

I am good with this change 

@haoqing0110 @zhiweiyin318 please take a final look and I think we can tag it.

cluster/v1beta1/types_placement.go
```diff
@@ -94,6 +94,50 @@ type PlacementSpec struct {
...
+	// The specific number will divide the placementDecisions to decisionGroups each group has max number of clusters equal to that specific number.
+	// The percentage will divide the placementDecisions to decisionGroups each group has max number of clusters based on the total num of selected clusters and percentage.
+	// ex; for a total 100 clusters selected, ClustersPerDecisionGroup equal to 20% will divide the placement decision to 5 groups each group should have 20 clusters.
+	// Default is having all clusters in a single group.
+	// If the DecisionGroups field defined, it will be considered first to create the decisionGroups then the ClustersPerDecisionGroup will be used to determine the rest of decisionGroups.
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:Pattern="^((100|[0-9]{1,2})%|[0-9]+)$"
+	// +kubebuilder:default:="100%"
+	// +optional
+	ClustersPerDecisionGroup string `json:"clustersPerDecisionGroup,omitempty"`
```
haoqing0110: Using `intstr.IntOrString` here?

cluster/v1beta1/helpers.go
```diff
@@ -175,3 +175,53 @@ func (pdct *PlacementDecisionClustersTracker) Existing() sets.Set[string] {
...
+	Type RolloutStrategyType `json:"type,omitempty"`
+
+	// List of the decision groups names or indexes to apply the workload first and fail if workload did not reach successful ((not proceed to apply workload to other decision groups/clusters).
+	// +optional
+	MandatoryDecisionGroups []MandatoryDecisionGroup `json:"mandatoryDecisionGroups,omitempty"`
+
+	// Timeout defined in minutes for how long workload applier controller will wait till workload reach successful state in the cluster. Only considered for Rollout Type Progressive and ProgressivePerGroup. Default is None meaning the workload applier will not proceed apply workload to other clusters if did not reach the successful state.
+	// +kubebuilder:default:=None
+	// +optional
+	Timeout string `json:"timeout,omitempty"`
```
zhujian7: what is the unit of the Timeout field? minute? it could be "1h", "30m"? or "1hour", "30minutes"?

qiujian16: good point, it should be second with int32 as type commonly

serngawy: good point thanks, I added a validation pattern for [0-9]+[h|m|s ] |None

cluster/v1beta1/types_placement.go
serngawy: I believe the validation pattern is good enough to do the same check. 
 actually I tried to set the type as intOrString but it gives me error while trying to build the API (error about having the validation pattern 100|[0-9]% with intOrString type) it did not build successful.

cluster/v1beta1/helpers.go
zhiweiyin318: should be `*int32`?  my understanding is that None and 0 have different meanings, right?

serngawy: yes, its string type to be able to set None

cluster/v1beta1/types_placement.go
haoqing0110: Using string should be fine from the perspective of functional implementation. 

One benefit of using `IntOrString` I can think is that it's in line with user habits, like k8s [deployment](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/api/apps/v1/types.go#L449) `maxSurge` and `maxUnavailable`, user don't need to warp "" for an integer in yaml. 
```
The Placement "demo1" is invalid: spec.decisionStrategy.groupStrategy.clustersPerDecisionGroup: Invalid value: "integer": spec.decisionStrategy.groupStrategy.clustersPerDecisionGroup in body must be of type string: "integer"
```

For the format check, another method is that we can check it in code like https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apimachinery/pkg/util/intstr/intstr.go#LL212C6-L212C32 and report error in placement "misconfigured" condition. Hope this can help.

serngawy: Nice, wasn't aware that intOrString type has function to validate the percentage. okay, changing it to intOrString

cluster/v1beta1/helpers.go
haoqing0110: FYI https://github.com/open-cluster-management-io/api/blob/main/cluster/v1beta1/types_placement.go#L320 which is nil by default. 
Are there user cases that would be waiting for minutes or even hours?

serngawy: yes, upgrade OCP takes at least 50min. other examples; install Subscription/operator (ACM), images download (preCache) take around 5min-10min. AND the timeout string can be validated using the ParseDuration https://pkg.go.dev/time#ParseDuration

haoqing0110:
LGTM
I'm not sure if helpers.go is the right place to put the common rollout strategy API, since it will be a general API, being part of addon/policy/etc, even a duck type. 
Can this PR get split and merge the placement decision strategy part first? 
cc @serngawy @qiujian16

serngawy:
> LGTM I'm not sure if helpers.go is the right place to put the common rollout strategy API, since it will be a general API, being part of addon/policy/etc, even a duck type. Can this PR get split and merge the placement decision strategy part first? cc @serngawy @qiujian16

As we discussed the rolloutStrategy struct is separated from the helper and moved to  another PR [0].  I believe the DecisionStrategy is good to merge.

[0] https://github.com/open-cluster-management-io/api/pull/244

PR #244, opened 2023-06-20
Add common RolloutStrategy API

serngawy:


serngawy:
Add @haoqing0110

cluster/v1alpha1/types_rolloutstrategy.go
```diff
@@ -0,0 +1,96 @@
...
+// proposal at https://github.com/open-cluster-management-io/enhancements/tree/main/enhancements/sig-architecture/64-placementStrategy
+
+// Rollout strategy to be used by workload applier controller.
+type RolloutStrategy struct {
+	// Rollout strategy Types are All, Progressive and ProgressivePerGroup
+	// 1) All means apply the workload to all clusters in the decision groups at once.
+	// 2) Progressive means apply the workload to the selected clusters progressively per cluster. The workload will not be applied to the next cluster unless one of the current applied clusters reach the successful state or timeout.
+	// 3) ProgressivePerGroup means apply the workload to decisionGroup clusters progressively per group. The workload will not be applied to the next decisionGroup unless all clusters in the current group reach the successful state or timeout.
+	// +kubebuilder:default:=All
+	// +optional
```
qiujian16, line 28: this is need to an enum validation

cluster/v1alpha1/types_rolloutstrategy.go
```diff
@@ -0,0 +1,96 @@
...
+	// Timeout default value is None meaning the workload applier will not proceed apply workload to other clusters if did not reach the successful state.
+	// Timeout must be defined in [0-9h]|[0-9m]|[0-9s] format examples; 2h , 90m , 360s
+	// +kubebuilder:validation:Pattern="^(([0-9])+[h|m|s])|None$"
+	// +kubebuilder:default:=None
+	// +optional
+	Timeout string `json:"timeout,omitempty"`
+}
+
+// MandatoryDecisionGroup set the decision group name or group index.
+type MandatoryDecisionGroup struct {
```
haoqing0110: How to handle the case when both GroupName and GroupIndex are defined in each MandatoryDecisionGroup? Can `intorstring` make it easier?

cluster/v1alpha1/types_rolloutstrategy.go
```diff
@@ -0,0 +1,96 @@
+package v1alpha1
+
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+// +k8s:deepcopy-gen=true
+
+// RolloutStrategy API used by consumer workload applier APIs such as ManifestWorkReplicaSet, Policy, ManagedClusterAddon ...etc
```
qiujian16: we should only indicate that this is for rolling out workload or configuration across clusters. We do not need to mention  ManifestWorkReplicaSet, Policy, ManagedClusterAddon. Otherwise, the doc will be added into each crd definition.

cluster/v1alpha1/types_rolloutstrategy.go
qiujian16: can we allow setting GroupName and GroupIndex at the same time?

cluster/v1alpha1/types_rolloutstrategy.go
```diff
@@ -0,0 +1,96 @@
...
+// RolloutProgressive is RolloutStrategyType
+type RolloutProgressive struct {
+	// List of the decision groups names or indexes to apply the workload first and fail if workload did not reach successful ((not proceed to apply workload to other decision groups/clusters).
+	// +optional
+	MandatoryDecisionGroups []MandatoryDecisionGroup `json:"mandatoryDecisionGroups,omitempty"`
+
+	// MaxConcurrency is the max number of clusters to deploy workload concurrently. The default value for MaxConcurrency is determined from the clustersPerDecisionGroup defined in the placement->DecisionStrategy.
+	// +kubebuilder:validation:Pattern="^((100|[0-9]{1,2})%|[0-9]+)$"
+	// +optional
+	MaxConcurrency string `json:"maxConcurrency,omitempty"`
```
haoqing0110: Have the below question for `maxconcurrency`, might need a clearer explanation.

1. Does the `maxconcurrency` also work on MandatoryDecisionGroup？
2. If a placement strategy is defined as below:
```
  DecisionStrategy:
    numberOfClustersPerDecisionGroup: 150
    DecisionGroups:
    - label: prod-canary-west
      clusterSelector:
        matchExpressions:
          - key: prod-canary-west
            operator: Exist
….
status:
  numberOfSelectedClusters: 300
  decisionGroups:
  - decisionGroupIndex: 0
    decisionGroupLabel: prod-canary-west
    placementDecisions:
    - ztp-placement-decision-0 // cluster1 to cluster 100
    - ztp-placement-decision-1 // cluster101 to cluster120
    clusterCount: 120
  - decisionGroupIndex: 1
    placementDecisions:
    - ztp-placement-decision-2 // cluster121 to cluster 220
    - ztp-placement-decision-3 // cluster221 to cluster 270
    clusterCount: 150 //
  - decisionGroupIndex: 2
    placementDecisions:
    - ztp-placement-decision-4  // cluster271 to cluster 300
    clusterCount: 30
```
When type is `rolloutProgressive` and not define any `mandatoryDecisionGroups`, what’s the expected upgrade behavior, will decision groups be ignored and upgrade cluster1 to cluster 150 first actually?

cluster/v1alpha1/types_rolloutstrategy.go
serngawy: When user set GroupName group selection should be based on GroupName even if user set GroupIndex.

serngawy: no, We should consider GroupName first then GroupIndex.

cluster/v1alpha1/types_rolloutstrategy.go
serngawy: Yes, based on the example above if the MandatoryDecisionGroup and Maxconcurrency not defined the rolloutProgressive should follow numberOfClustersPerDecisionGroup. Maxconcurrency to give the  option override the numberOfClustersPerDecisionGroup.

cluster/v1alpha1/types_rolloutstrategy.go
serngawy: okay, done

cluster/v1alpha1/types_rolloutstrategy.go
serngawy: okay, done

cluster/v1alpha1/types_rolloutstrategy.go
haoqing0110: Sounds good, is that possible to add this explanation in the API comments?

cluster/v1alpha1/types_rolloutstrategy.go
haoqing0110: Should this filed also be the type `intorstring` to be consistent with clustersPerDecisionGroup?

cluster/v1alpha1/types_rolloutstrategy.go
```diff
@@ -0,0 +1,83 @@
...
+
+	// +optional
+	Timeout Timeout `json:",inline"`
+}
+
+// Timeout to consider while applying the workload.
+type Timeout struct {
+	// Timeout define how long workload applier controller will wait till workload reach successful state in the cluster. Only considered for Rollout Type Progressive and ProgressivePerGroup.
```
haoqing0110: This comment is inconsistent with the above type `RolloutAll` which has timeout field.
```
// RolloutAll is a RolloutStrategy Type
type RolloutAll struct {
	// +optional
	Timeout Timeout `json:",inline"`
}
```

cluster/v1alpha1/types_rolloutstrategy.go
serngawy: okay, done.

cluster/v1alpha1/types_rolloutstrategy.go
serngawy: ops, removed.

cluster/v1alpha1/types_rolloutstrategy.go
serngawy: sure, done.

haoqing0110:
LGTM

cluster/v1alpha1/types_rolloutstrategy.go
```diff
@@ -0,0 +1,88 @@
+package v1alpha1
+
+import (
+	"k8s.io/apimachinery/pkg/util/intstr"
+)
+
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
```
qiujian16: this is not a runtime object so line 7 is not needed. see here https://pkg.go.dev/k8s.io/gengo/examples/deepcopy-gen

qiujian16:
/approve
/lgtm
/hold

@haoqing0110 are we fine if there is no DeepCopy?

haoqing0110:
DeepCopy() for the common rollout strategy is needed I think, since we treat it as a duck type and will add library or helper function only for this part. 
cc @serngawy, running `make update` should generate that.

serngawy:
> DeepCopy() for the common rollout strategy is needed I think, since we treat it as a duck type and will add library or helper function only for this part. cc @serngawy, running `make update` should generate that.

`$make update` didn't generate deepcopy. However, as the rolloutStrategy will be part of others applier APIs spec the deepcopy will be generated with the applier API.

cluster/v1alpha1/types_rolloutstrategy.go
haoqing0110: Still need some const  string like after adding enum validation.

```
const (
	//All means apply the workload to all clusters in the decision groups at once.
	All string = "All"
	//Progressive means apply the workload to the selected clusters progressively per cluster.
	Progressive string = "Progressive"
	//ProgressivePerGroup means apply the workload to the selected clusters progressively per group.
	ProgressivePerGroup string = "ProgressivePerGroup"
)
```

serngawy: ops, removed by mistake. done

cluster/v1alpha1/types_rolloutstrategy.go
```diff
@@ -0,0 +1,103 @@
...
+	// +optional
+	All *RolloutAll `json:"all,omitempty"`
+
+	// Progressive define required fields for RolloutStrategy type Progressive
+	// +optional
+	Progressive *RolloutProgressive `json:"progressive,omitempty"`
+
+	// ProgressivePerGroup define required fields for RolloutStrategy type ProgressivePerGroup
+	// +optional
+	ProgressivePerGroup *RolloutProgressivePerGroup `json:"progressivePerGroup,omitempty"`
```
dhaiducek, line 41: Is there a reason behind having these as separate keys? Could it not be streamlined as providing the `Type` and inlining all of the options? If that's an option, I can work on a PR. At minimum, I think it might be nice to have `Timeout` at the same level as `Type`?

```golang
type RolloutStrategy struct {
	// +kubebuilder:validation:Enum=All;Progressive;ProgressivePerGroup
	// +kubebuilder:default:=All
	// +optional
	Type string `json:"type,omitempty"`

	// +optional
	Timeout `json:",inline"`

	// +optional
	MandatoryDecisionGroups `json:",inline"`

	// +kubebuilder:validation:Pattern="^((100|[0-9]{1,2})%|[0-9]+)$"
	// +kubebuilder:validation:XIntOrString
	// +optional
	MaxConcurrency intstr.IntOrString `json:"maxConcurrency,omitempty"`
}
```

qiujian16: this is designed as an union type, a certain type is point to the a certain field. see here https://github.com/kubernetes/enhancements/blob/master/keps/sig-api-machinery/1027-api-unions/README.md. This could avoid unexpected wrong configurations. For instance, you would not want user set Type to Progressive and still set the MandatoryDecisionGroups.

serngawy: It has been done with union type to avoid miss leading the endUser as for ex; when the Type is ALL there is no need to define the MandatoryDecisionGroups and MaxConcurrency. As well as when the Type is ProgressivePerGroup no need to define MaxConcurrency. 
So make sense to let each Type has its own field.

dhaiducek: Thanks for the comments--it'd be nice if the type could be inferred from the key rather than having an additional `type` to avoid the stutter, which I think is what feels awkward about this to me, but it seems that's not possible. 🙁

dhaiducek: Actually, I don't see the union type even implemented yet in Kubernetes?
Kubebuilder doesn't support `OneOf` (yet?), but some Kustomize post-processing on the CRD could let consumers use the `OneOf` in the CRD rather than having a `Type` field. This would also avoid confusion since it'd be impossible to set the `Type` to one thing and then provide a set of other options (as I think can be done as it is now)?

```yaml
oneOf:
- required: [ "all" ]
- required: [ "progressive" ]
- required: [ "progressivePerGroup" ]
```

Though without Kubebuilder support, it'd have to be implemented by the consumers of the library, unfortunately.

PR #245, opened 2023-06-29
make the custom sign ca secret namespace fixed

zhujian7:


addon/v1alpha1/types_addontemplate.go
```diff
@@ -155,19 +155,17 @@ type CustomSignerRegistrationConfig struct {
...
 	// SigningCA represents the reference of the secret on the hub cluster to sign the CSR
+	// the secret name must be "custom-signer-secret", and the secret type must be "kubernetes.io/tls"
 	// +kubebuilder:validation:Required
 	SigningCA SigningCARef `json:"signingCA"`
 }
 
-// SigningCARef is the reference to the signing CA secret that must contain the
-// certificate authority data with key "ca.crt" and the private key data with key "ca.key"
+// SigningCARef is the reference to the signing CA secret which type must be "kubernetes.io/tls" and
+// the name must be "custom-signer-secret"
```
qiujian16: we might need to use a more specific secret name. e.g. ocm-addon-template-signer

addon/v1alpha1/types_addontemplate.go
```diff
@@ -155,19 +155,17 @@ type CustomSignerRegistrationConfig struct {
 	Subject *Subject `json:"subject,omitempty"`
 
 	// SigningCA represents the reference of the secret on the hub cluster to sign the CSR
+	// the secret name must be "ocm-addon-template-signer", and the secret type must be "kubernetes.io/tls"
 	// +kubebuilder:validation:Required
 	SigningCA SigningCARef `json:"signingCA"`
```
zhujian7, line 160: @qiujian16 since we removed the secret name field, do you think the SigningCARef is still necessary? or we just change it to signingCANamespace?

qiujian16: I am thinking whether it makes more sense to watch secret in a certain namespace, rather than the a fixed name.

PR #246, opened 2023-06-30
Add singleton mode for klusterlet

qiujian16:
with inprocess mode, agent will be run in the
same process with the klusterlet.


zhujian7:
LGTM

PR #249, opened 2023-07-04
enable addon management feature gate by default

zhujian7:


qiujian16:
let's ask for comment on the ocm slack before we switch.

PR #250, opened 2023-07-11
add decision group type

haoqing0110:
Today the filed `DecisionGroup` and `ClustersPerDecisionGroup` is on the same level,  and the name "ClustersPerDecisionGroup" makes user easy to misunderstand that it's used to define the cluster numbers of "DecisionGroup". 

To avoid misunderstanding, this PR wants to add a decision group type static and dynamic, to clarify that `DecisionGroup` and `ClustersPerDecisionGroup` are working separately.


serngawy:
Not sure,  where is a confusion come from ? The user can only set the ClustersPerDecisionGroup to divide the selected clusters to groups

serngawy (changes requested):
can you elaborate more on the confusion.

haoqing0110:
For example as below, the "DecisionGroups" "prod-canary-west" has 120 clusters, and it actually won't be divide by "numberOfClustersPerDecisionGroup". "numberOfClustersPerDecisionGroup" only works on the rest of "DecisionGroups" "prod-canary-west".

```
  DecisionStrategy:
    numberOfClustersPerDecisionGroup: 50
    DecisionGroups:
    - label: prod-canary-west
      clusterSelector:
        matchExpressions:
          - key: prod-canary-west
            operator: Exist
….
status:
  numberOfSelectedClusters: 200
  decisionGroups:
  - decisionGroupIndex: 0
    decisionGroupLabel: prod-canary-west
    placementDecisions:
    - ztp-placement-decision-0 // cluster1 to cluster 100
    - ztp-placement-decision-1 // cluster101 to cluster120
    clusterCount: 120
  - decisionGroupIndex: 1
    placementDecisions:
    - ztp-placement-decision-2 // cluster121 to cluster 170
    clusterCount: 50 //
  - decisionGroupIndex: 2
    placementDecisions:
    - ztp-placement-decision-4  // cluster171 to cluster 200
    clusterCount: 30
```

serngawy:
> For example as below, the "DecisionGroups" "prod-canary-west" has 120 clusters, and it actually won't be divide by "numberOfClustersPerDecisionGroup". "numberOfClustersPerDecisionGroup" only works on the rest of "DecisionGroups" "prod-canary-west".
> 
> ```
>   DecisionStrategy:
>     numberOfClustersPerDecisionGroup: 50
>     DecisionGroups:
>     - label: prod-canary-west
>       clusterSelector:
>         matchExpressions:
>           - key: prod-canary-west
>             operator: Exist
> ….
> status:
>   numberOfSelectedClusters: 200
>   decisionGroups:
>   - decisionGroupIndex: 0
>     decisionGroupLabel: prod-canary-west
>     placementDecisions:
>     - ztp-placement-decision-0 // cluster1 to cluster 100
>     - ztp-placement-decision-1 // cluster101 to cluster120
>     clusterCount: 120
>   - decisionGroupIndex: 1
>     placementDecisions:
>     - ztp-placement-decision-2 // cluster121 to cluster 170
>     clusterCount: 50 //
>   - decisionGroupIndex: 2
>     placementDecisions:
>     - ztp-placement-decision-4  // cluster171 to cluster 200
>     clusterCount: 30
> ```

Well, looks kind of strange case; How is the num of canary clusters is more than the rest of running clusters ? theoretically this could happen but practically I don't think this right. 
In other hand;  if that the case better to apply the numberOfClustersPerDecisionGroup to the decisionGroup so we could have many decision group with same GroupNameLabel and different GroupIndexLabel to satisfy the numberOfClustersPerDecisionGroup.

haoqing0110:
This example is not a real-world one, it is to explain why I think the name is confusing. The defined "DecisionGroups"  is actually not divided by "numberOfClustersPerDecisionGroup". 

1. If we don't expect the canary clusters to be more than the rest of the running clusters and don't expect the canary clusters to be divided, the current API name is confusing, that's why I create this PR. 
2. If we think once canary clusters is larger than the "numberOfClustersPerDecisionGroup" and need to be divided, I'm fine to keep the current API, just change the API explanation.

serngawy:
> This example is not a real-world one, it is to explain why I think the name is confusing. The defined "DecisionGroups" is actually not divided by "numberOfClustersPerDecisionGroup".
> 
> 1. If we don't expect the canary clusters to be more than the rest of the running clusters and don't expect the canary clusters to be divided, the current API name is confusing, that's why I create this PR.
> 2. If we think once canary clusters is larger than the "numberOfClustersPerDecisionGroup" and need to be divided, I'm fine to keep the current API, just change the API explanation.

Okay, that is fine then let's stay with option 2.

haoqing0110:
option 2 sounds make sense, close this PR. https://github.com/open-cluster-management-io/api/pull/253

PR #252, opened 2023-07-13
Add cluster annotations.

xuezhaojun:


operator/v1/types.go
```diff
@@ -382,6 +382,14 @@ type KlusterletSpec struct {
 	// +kubebuilder:validation:Pattern=^[a-z0-9]([-a-z0-9]*[a-z0-9])?$
 	ClusterName string `json:"clusterName,omitempty"`
 
+	// InitClusterLabels is a set of labels that are applied to the managed cluster created during the registration process.
+	// +optional
+	InitClusterLabels map[string]string `json:"initClusterLabels,omitempty"`
```
qiujian16: can use change this field, what if user update it? And should we ensure user cannot set arbitrary key on that?

xuezhaojun: The fields are assumed only to be used in the "init" in here: https://github.com/open-cluster-management-io/ocm/blob/41f430b0776a4a7dbb1006a2125838e2994c4b17/pkg/registration/spoke/registration/creating_controller.go#L65

After the bootstrap secret expired, we will never enter this block of code again, so "update" operation will do nothing.

--

I'm not sure what restrictions we need in this case. What key we expect user not use?

qiujian16: for instance clusterset label key. init is not quite clear, this only happens when the agent is trying to create the cluster. If user delete the cluster on the hub, and agent will recreate the cluster use this label. So we cannot say it is only used during init.

qiujian16: and I think it should be put under registration configuration field

xuezhaojun: Got it, let me think it through again.

xuezhaojun: Fixed, @qiujian16  PTAL

operator/v1/types.go
```diff
@@ -107,6 +107,13 @@ type RegistrationConfiguration struct {
 	//  	he can set featuregate/Foo=false before upgrading. Let's say the cluster-admin wants featuregate/Foo=false.
 	// +optional
 	FeatureGates []FeatureGate `json:"featureGates,omitempty"`
+
+	// ClusterAnnotations is a set of annotations that agent used when creating a managed cluster.
```
qiujian16:
```suggestion
	// ClusterAnnotations is annotations with the reserve prefix "agent.open-cluster-management.io" set on
	// ManagedCluster when creating only, other actors can update it afterwards.
```

operator/v1/types.go
```diff
@@ -107,6 +107,13 @@ type RegistrationConfiguration struct {
...
+	// After the managed cluster is created, modify this field won't affect the managed cluster any more.
```
qiujian16: also need a const definition of this annotationPrefix

xuezhaojun: Yes, adding one now...

xuezhaojun: Added the `ClusterAnnotationsKeyPrefix `, PTAL.

qiujian16: After the managed cluster is created, modify this field won''t affect the managed cluster any more. But if the managed cluster is deleted from the hub, it will be used again when agent attempt to create a managed cluster. The annotations must has prefix: "agent.open-cluster-management.io", otherwise, it will be ignored.'

I do not think you still need these.

xuezhaojun: Agree, the content deleted. PTAL.

qiujian16 (commented):
/approve
/lgtm

PR #254, opened 2023-07-20
rollout helper functions

haoqing0110:
An idea to treat the common rollout strategy API as a duck type, and provide some helper functions for the common rollout API.
- The input would be rollout strategy, the placement decision tracker, and the resource status on each managed cluster
-  The output would be a list of clusters that need to update.

https://github.com/open-cluster-management-io/ocm/issues/236
https://github.com/open-cluster-management-io/ocm/issues/237


haoqing0110:
cc @qiujian16

cluster/v1alpha1/helpers.go
```diff
@@ -0,0 +1,222 @@
...
+	// resource desired status is upgraded and last applied status upgrade failed
+	UpgradeFailed
+)
+
+// +k8s:deepcopy-gen=false
+type RolloutHandler struct {
+	// duck type rollout strategy
+	rolloutStrategy RolloutStrategy
+	// placement decision tracker
+	pdTpdTracker clusterv1beta1.PlacementDecisionClustersTracker
```
qiujian16: typo?

cluster/v1alpha1/helpers.go
```diff
@@ -0,0 +1,222 @@
...
+	// Record the resource rollout status on each managed cluster
+	// For example, map[string]RolloutStatus{"cluster1":UpgradeSucceed, "cluster2":Upgrading, "cluster3":ToUpgrade}
+	objs map[string]RolloutStatus
+}
+
+func NewRolloutHandler(rolloutStrategy RolloutStrategy, pdTpdTracker clusterv1beta1.PlacementDecisionClustersTracker, objs map[string]RolloutStatus) *RolloutHandler {
```
qiujian16: instead of map[string]RolloutStatus, will be a func(clusterName string) RolloutStatus betterr?

cluster/v1alpha1/helpers.go
```diff
@@ -0,0 +1,222 @@
...
+	return &RolloutHandler{
+		rolloutStrategy: rolloutStrategy,
+		pdTpdTracker:    pdTpdTracker,
+		objs:            objs,
+	}
+}
+
+func (r *RolloutHandler) GetRolloutCluster() []string {
```
qiujian16: it is not quite clear on what is the output, is this the list cluster that needs to be updated with current state? If so I think the the input should have
1. the strategy
2. the current state.

cluster/v1alpha1/helpers.go
```diff
@@ -0,0 +1,222 @@
...
+}
+
+func (r *RolloutHandler) progressivePerGroup(clusterGroupKeys []clusterv1beta1.GroupKey, clusterGroups map[clusterv1beta1.GroupKey]sets.Set[string], timeout string) []string {
+	result := []string{}
+
+	for _, key := range clusterGroupKeys {
+		if subclusters, ok := clusterGroups[key]; ok {
+			for _, cluster := range subclusters.UnsortedList() {
+				if status, exist := r.objs[cluster]; exist {
+					if canIAppendToResult(status, timeout) {
```
qiujian16: canIAppendToResult is this actually isUpgrading?

cluster/v1alpha1/helpers.go
haoqing0110: Yes, good suggestion. Fixed.

cluster/v1alpha1/helpers.go
haoqing0110: Update the output.

cluster/v1alpha1/helpers.go
haoqing0110: Not only is upgrading,  but also some failed but skipped clusters.

haoqing0110: May need a better function name.

cluster/v1alpha1/helpers.go
```diff
@@ -0,0 +1,305 @@
...
+	"k8s.io/utils/clock"
+	clusterv1beta1 "open-cluster-management.io/api/cluster/v1beta1"
+)
+
+type RolloutStatus int
+
+var RolloutClock = clock.Clock(clock.RealClock{})
+var maxTimeDuration = time.Duration(math.MaxInt64)
+
+const (
```
serngawy, line 25: Not all cases are upgrade, better names like
ToApply
Progressing 
Succeed
Failed

serngawy (commented):
in general sounds good and the class cluster/v1alpha1/helpers.go need unit test

cluster/v1alpha1/helpers.go
```diff
@@ -0,0 +1,311 @@
...
+	}
+
+	return &RolloutHandler{
+		rolloutStrategy:          rolloutStrategy,
+		pdTracker:                pdTracker,
+		clusterRolloutStatusFunc: clusterRolloutStatusFunc,
+	}
+}
+
+// Return the strategy actual take effect and a list of cluster that needs to update with current state
```
serngawy: The RolloutHandler already have the RolloutStrategy struct that already has its rolloutType. Why do we have GetRolloutAllClusters, GetRolloutProgressive and GetRolloutProgressivePerGroup Functions as public ? 
GetRolloutCluster Func should be enough ?

cluster/v1alpha1/helpers.go
```diff
@@ -0,0 +1,311 @@
...
+}
+
+// +k8s:deepcopy-gen=false
+type RolloutHandler struct {
+	// duck type rollout strategy
+	rolloutStrategy RolloutStrategy
+	// placement decision tracker
+	pdTracker *clusterv1beta1.PlacementDecisionClustersTracker
+	// Return the rollout status on each managed cluster
+	clusterRolloutStatusFunc ClusterRolloutStatusFunc
```
serngawy: does the clusterRolloutStatusFunc should be implemented in the Consumer controller ? unit test should be good place for an example.

cluster/v1alpha1/helpers.go
haoqing0110: Yes, agree, modify them as private func.

cluster/v1alpha1/helpers.go
haoqing0110: Yes, clusterRolloutStatusFunc should be implemented by the Consumer. Add more UT.

cluster/v1alpha1/helpers.go
haoqing0110: Will think about the name.

haoqing0110: The new name sounds good, have updated the status name.

haoqing0110:
/assign @qiujian16 @serngawy
 
Based on our offline discussion, the full code and UT are now added, plz help review and provide more comments.

cluster/v1alpha1/helpers.go
```diff
@@ -0,0 +1,316 @@
...
+
+func NewRolloutHandler(rolloutStrategy RolloutStrategy, pdTracker *clusterv1beta1.PlacementDecisionClustersTracker, clusterRolloutStatusFunc ClusterRolloutStatusFunc) *RolloutHandler {
+	if pdTracker == nil {
+		return nil
+	}
+
+	return &RolloutHandler{
+		rolloutStrategy:          rolloutStrategy,
+		pdTracker:                pdTracker,
+		clusterRolloutStatusFunc: clusterRolloutStatusFunc,
```
qiujian16: I think clusterRolloutStatusFunc and rolloutStrategy should be the input of GetRolloutCluster, you can init the handler at the beginning, and each time just give handler the strategy and current status.

qiujian16 (commented):
/approve

Need @serngawy and @mprahl to signoff

cluster/v1alpha1/helpers.go
```diff
@@ -0,0 +1,311 @@
...
+}
+
+func (r *RolloutHandler) getProgressiveClusters(rolloutStrategy RolloutStrategy, statusFunc ClusterRolloutStatusFunc) (*RolloutStrategy, map[string]ClusterRolloutStatus, error) {
+	strategy := RolloutStrategy{Type: Progressive}
+	strategy.Progressive = rolloutStrategy.Progressive.DeepCopy()
+	if strategy.Progressive == nil {
+		strategy.Progressive = &RolloutProgressive{}
+	}
+
+	// upgrade madatory decision groups first
```
serngawy: mandatory

cluster/v1alpha1/helpers.go
```diff
@@ -0,0 +1,311 @@
...
+	}
+
+	return &RolloutHandler{
+		pdTracker: pdTracker,
+	}
+}
+
+// The input is a duck type RolloutStrategy and a ClusterRolloutStatusFunc to return the rollout status on each managed cluster.
+// Return the strategy actual take effect and a list of cluster that needs to update with current state.
+func (r *RolloutHandler) GetRolloutCluster(rolloutStrategy RolloutStrategy, statusFunc ClusterRolloutStatusFunc) (*RolloutStrategy, map[string]ClusterRolloutStatus, error) {
```
serngawy: From Consumer API controller point of view; I think the ClusterRolloutStatus should contain the cluster groupIndex/Name ? How we can express that the Mandatory/DecisionGroups clusters are done or progressing ? Otherwise the logic of connecting the cluster to the decisionGroup need to be maintained in the controller as well. What do you think ?

cluster/v1alpha1/helpers.go
```diff
@@ -0,0 +1,311 @@
...
+}
+
+func (r *RolloutHandler) getProgressivePerGroupClusters(rolloutStrategy RolloutStrategy, statusFunc ClusterRolloutStatusFunc) (*RolloutStrategy, map[string]ClusterRolloutStatus, error) {
+	strategy := RolloutStrategy{Type: ProgressivePerGroup}
+	strategy.ProgressivePerGroup = rolloutStrategy.ProgressivePerGroup.DeepCopy()
+	if strategy.ProgressivePerGroup == nil {
+		strategy.ProgressivePerGroup = &RolloutProgressivePerGroup{}
+	}
+
+	// upgrade madatory decision groups first
```
serngawy: mandatory

cluster/v1alpha1/helpers.go
```diff
@@ -0,0 +1,311 @@
...
+
+func needToUpdate(status ClusterRolloutStatus, timeout time.Duration) bool {
+	switch status.Status {
+	case ToApply:
+		return true
+	case Progressing:
+		return true
+	case Succeed:
+		return false
+	case Failed:
```
serngawy: I believe we need to distinguish between timeout and failed. It can be failed before reaching timeout AND as well it can be timed out but not failed  (as the status will continue to reconcile it will reach failed/success later).

cluster/v1alpha1/helpers.go
```diff
@@ -0,0 +1,311 @@
...
+	}
+
+	return result
+}
+
+func needToUpdate(status ClusterRolloutStatus, timeout time.Duration) bool {
+	switch status.Status {
+	case ToApply:
+		return true
+	case Progressing:
```
serngawy: We should check for timeout within progressing state, otherwise it will progress forever ?

serngawy (changes requested):

cluster/v1alpha1/helpers.go
haoqing0110: Have updated the code.

cluster/v1alpha1/helpers.go
haoqing0110: It makes sense I think. Can have GroupKey added, only for the clusters need to update eg,
```
type ClusterRolloutStatus struct {
        // cluster group name and group key
        GroupKey GroupKey
	// rollout status
	Status RolloutStatus
	// the last transition time of rollout status
	LastTransitionTime *time.Time
}

func (r *RolloutHandler) GetRolloutCluster(rolloutStrategy RolloutStrategy, statusFunc ClusterRolloutStatusFunc) (*RolloutStrategy, map[string]ClusterRolloutStatus, error) {...
```
Or maybe a whole picture, eg,
```
type ClustersRolloutStatus map[string]ClusterRolloutStatus

type RolloutStatus struct {
	MandatoryGroupsSucceed bool
       // progressive with mandatory groups or progressive per group will return this field
        DecisionGroupsStatus map[GroupKey]ClustersRolloutStatus 
        // all or progressive non mandatory group clusters will return in this field
	ClustersStatus ClustersRolloutStatus 
        // clusters to rollout, all the rollout strategy will return this field
	ClustersToRollout ClustersRolloutStatus
}

func (r *RolloutHandler) GetRolloutCluster(rolloutStrategy RolloutStrategy, statusFunc ClusterRolloutStatusFunc) (*RolloutStrategy, RolloutStatus, error) {...

```
can discuss more in this part.

cluster/v1alpha1/helpers.go
haoqing0110: Sounds good. Failed or progressing is the status provided by consumer contoller's ClusterRolloutStatusFunc, and should return timeout in the result of ClusterRolloutStatus by the RolloutHandler.

cluster/v1alpha1/helpers.go
haoqing0110: Add timeout check for progressing.

cluster/v1alpha1/helpers.go
haoqing0110: Have updated the func, when the cluster staus is progressing or failed and reach timeout, will return status TimeOut.

cluster/v1alpha1/helpers.go
haoqing0110: Per the offline discussion, the returned ClusterRolloutStatus will contain the group name and group index. 

Also, since the timeout clusters are intended to skip and continue to rollout others, the returned value would be like below, separate clusters to rollout and clusters that are timeout. 
```
type RolloutResult struct {
	// clusters to rollout
	ClustersToRollout map[string]ClusterRolloutStatus
	// clusters that are timeout
	ClustersTimeOut map[string]ClusterRolloutStatus
}
```

cluster/v1alpha1/helpers.go
```diff
@@ -0,0 +1,388 @@
...
+	Progressing
+	// resource desired status is applied and last applied status is succeed
+	Succeed
+	// resource desired status is applied and last applied status is failed
+	Failed
+	// when the rollout status is progressing or failed and the status remains for longer than the timeout
+	// then the status will be set to timeout
+	TimeOut
+	// skip rollout on this cluster
+	Skip RolloutStatus = iota
```
serngawy: Not sure, what is the case for skip ?

haoqing0110: Add this because in the addon upgrade, when the user defines 2 placements, there might be clusters that overlap. Since each cluster should only belong to one placement, it should be skipped when going through another placement.

serngawy (commented):
/LGTM for me

cluster/v1alpha1/helpers.go
```diff
@@ -0,0 +1,388 @@
...
+	// then the status will be set to timeout
+	TimeOut
+	// skip rollout on this cluster
+	Skip
+)
+
+// Return the rollout status on each managed cluster
+type ClusterRolloutStatusFunc func(clusterName string) ClusterRolloutStatus
+
+type ClusterRolloutStatus struct {
```
mprahl, line 45: Was there any consideration for a history of state transitions of the rollout?

cluster/v1alpha1/helpers.go
```diff
@@ -0,0 +1,388 @@
...
+	Status RolloutStatus
+	// the last transition time of rollout status
+	// this is used to calculate timeout for progressing and failed status
+	// optional field
+	LastTransitionTime *time.Time
+}
+
+type RolloutResult struct {
+	// clusters to rollout
+	ClustersToRollout map[string]ClusterRolloutStatus
```
mprahl, line 60: It seems redundant that `ClusterRolloutStatus` has a cluster name field when it's used in the context of a map where the keys are cluster names.

My opinion is to remove `ClusterRolloutStatus.Name` or make `ClustersToRollout` a slice of `[]ClusterRolloutStatus`.

The same feedback applies for `ClustersTimeOut`.

cluster/v1alpha1/helpers.go
```diff
@@ -0,0 +1,388 @@
...
+
+// +k8s:deepcopy-gen=false
+type RolloutHandler struct {
+	// placement decision tracker
+	pdTracker *clusterv1beta1.PlacementDecisionClustersTracker
+}
+
+func NewRolloutHandler(pdTracker *clusterv1beta1.PlacementDecisionClustersTracker) *RolloutHandler {
+	if pdTracker == nil {
+		return nil
```
mprahl: I think it'd be more appropriate to return an error here instead.

cluster/v1beta1/helpers.go
```diff
@@ -117,18 +119,29 @@ type PlacementDecisionGetter interface {
...
+	existingScheduledClusters      sets.Set[string]
+	existingScheduledClusterGroups map[GroupKey]sets.Set[string]
+	clustesGroupsIndexToName       map[int32]string
+	clustesGroupsNameToIndex       map[string][]int32
+	lock                           sync.RWMutex
 }
 
-func NewPlacementDecisionClustersTracker(placement *Placement, pdl PlacementDecisionGetter, existingScheduledClusters sets.Set[string]) *PlacementDecisionClustersTracker {
+// +k8s:deepcopy-gen=false
+type GroupKey struct {
```
mprahl, line 132: Would it make sense to track the placement decision name and `resourceVersion` to know when it may have changed in the middle of the rollout or after the fact?

This could be used as a "shortcut" during a controller's initial list query of applicable PlacementDecision objects.

cluster/v1alpha1/helpers.go
```diff
@@ -0,0 +1,388 @@
...
+
+func (r *RolloutHandler) getRolloutAllClusters(rolloutStrategy RolloutStrategy, statusFunc ClusterRolloutStatusFunc) (*RolloutStrategy, RolloutResult, error) {
+	strategy := RolloutStrategy{Type: All}
+	strategy.All = rolloutStrategy.All.DeepCopy()
+	if strategy.All == nil {
+		strategy.All = &RolloutAll{}
+	}
+
+	// parse timeout
+	failureTimeout, err := parseTimeout(strategy.All.Timeout.Timeout)
```
mprahl, line 109: What does a timeout do for the all strategy? In the policy example, presumably all clusters would be assigned the policy and if they are connected to the hub, they would copy the policy down and evaluate it.

cluster/v1alpha1/helpers.go
```diff
@@ -0,0 +1,388 @@
...
+		strategy.Progressive = &RolloutProgressive{}
+	}
+
+	// upgrade mandatory decision groups first
+	groupKeys := decisionGroupsToGroupKeys(strategy.Progressive.MandatoryDecisionGroups.MandatoryDecisionGroups)
+	clusterGroupKeys, clusterGroups := r.pdTracker.ExistingClusterGroups(groupKeys)
+
+	rolloutresult := progressivePerGroup(clusterGroupKeys, clusterGroups, maxTimeDuration, statusFunc)
+	if len(rolloutresult.ClustersToRollout) > 0 {
+		return &strategy, rolloutresult, nil
```
mprahl: So this section only returns the clusters in the mandatory decision groups? If so, I was not expecting this behavior from the `GetRolloutCluster` description, which calls this method.

cluster/v1alpha1/helpers.go
```diff
@@ -0,0 +1,388 @@
...
+
+	return &RolloutHandler{
+		pdTracker: pdTracker,
+	}
+}
+
+// The input is a duck type RolloutStrategy and a ClusterRolloutStatusFunc to return the rollout status on each managed cluster.
+// Return the strategy actual take effect and a list of cluster that needs to update with current state.
+func (r *RolloutHandler) GetRolloutCluster(rolloutStrategy RolloutStrategy, statusFunc ClusterRolloutStatusFunc) (*RolloutStrategy, RolloutResult, error) {
+	switch rolloutStrategy.Type {
```
mprahl, line 88: What's the behavior if the rollout strategy changes in the middle of a rollout?

cluster/v1alpha1/helpers.go
```diff
@@ -0,0 +1,388 @@
...
+		strategy.ProgressivePerGroup = &RolloutProgressivePerGroup{}
+	}
+
+	// upgrade mandatory decision groups first
+	groupKeys := decisionGroupsToGroupKeys(strategy.ProgressivePerGroup.MandatoryDecisionGroups.MandatoryDecisionGroups)
+	clusterGroupKeys, clusterGroups := r.pdTracker.ExistingClusterGroups(groupKeys)
+
+	rolloutresult := progressivePerGroup(clusterGroupKeys, clusterGroups, maxTimeDuration, statusFunc)
+	if len(rolloutresult.ClustersToRollout) > 0 {
+		return &strategy, rolloutresult, nil
```
mprahl: Same comment as above. I was not expecting this behavior based on the description of the calling function.

cluster/v1alpha1/helpers.go
```diff
@@ -0,0 +1,388 @@
...
+			// go through group by group
+			for _, cluster := range subclusters.UnsortedList() {
+				status := statusFunc(cluster)
+				status.GroupKey = key
+
+				newstatus, needtorollout := needToRollout(status, timeout)
+				if needtorollout {
+					rolloutclusters[cluster] = status
+				}
+				if newstatus == TimeOut {
```
mprahl: Why is `newstatus` only used if it's a timeout?

cluster/v1alpha1/helpers.go
```diff
@@ -0,0 +1,388 @@
...
+	}
+
+	return &RolloutHandler{
+		pdTracker: pdTracker,
+	}
+}
+
+// The input is a duck type RolloutStrategy and a ClusterRolloutStatusFunc to return the rollout status on each managed cluster.
+// Return the strategy actual take effect and a list of cluster that needs to update with current state.
+func (r *RolloutHandler) GetRolloutCluster(rolloutStrategy RolloutStrategy, statusFunc ClusterRolloutStatusFunc) (*RolloutStrategy, RolloutResult, error) {
```
mprahl, line 87: What's the expected behavior for the caller to know when a cluster or cluster group has timed out?

It seems to me that somehow returning a context with a deadline for every potential timeout would allow the caller to poll the channel of the context to know if a timeout occurred and then verify if the rollout finished before that timeout.

cluster/v1alpha1/helpers.go
```diff
@@ -0,0 +1,388 @@
...
+
+	if length == 0 {
+		return RolloutResult{
+			ClustersToRollout: rolloutclusters,
+			ClustersTimeOut:   timeoutclusters,
+		}
+	}
+
+	// sort the clusters
+	sort.Strings(clusters)
```
mprahl, line 206: So the logic is to always progress in alphabetical order?

cluster/v1alpha1/helpers.go
```diff
@@ -0,0 +1,388 @@
...
+	}
+
+	if length <= 0 || length > total {
+		length = total
+	}
+
+	return length, nil
+}
+
+func parseTimeout(timeoutStr string) (time.Duration, error) {
```
mprahl, line 348: Why not use `time.ParseDuration`?

cluster/v1alpha1/helpers.go
haoqing0110: The current idea is to use the snapshot of cluster status to do rollout. What's the use case like to use history?

cluster/v1beta1/helpers.go
haoqing0110: Let me think about it..

cluster/v1alpha1/helpers.go
haoqing0110: For the all strategy, all the clusters would apply the new resource at once, some might keep in progressing or failed status, and if there's a timeout, the controller won't keep waiting for it to reach a successful state. 
cc @serngawy , correct me if not accurate.

cluster/v1alpha1/helpers.go
haoqing0110: If there's mandatory decision groups defined and not all clusters in mandatory groups upgrade successfully, `rolloutresult.ClustersToRollout` will contain the clusters that need to rollout and will return here. If not, it will continue to go through other groups.

cluster/v1alpha1/helpers.go
haoqing0110: I'm not sure if I understand the question here. Is expectation is to add more description for `GetRolloutCluster`, to mention that if mandatory groups defined, will return clusters to rollout in mandatory groups, and if all the mandatory groups rollout successfully, will return the rest of the clusters need to rollout ?

cluster/v1alpha1/helpers.go
haoqing0110: When the input status is progressing or failed, and the status lasts longer than timeout, want to rewrite the status to timeout and return.

cluster/v1alpha1/helpers.go
haoqing0110: @serngawy mentions that some policies might keep in progressing status on some clusters for a long time, don't want them to block upgrading others, and also need to know which are timeout, right?

cluster/v1alpha1/helpers.go
haoqing0110: yes, to ensure each time returns the same clusters that need to rollout.

cluster/v1alpha1/helpers.go
haoqing0110: The unit are different, so not use it: 
- For `time.ParseDuration` valid time units are "ns", "us" (or "µs"), "ms", "s", "m", "h". 
- For our timeout, it follows pattern `"^(([0-9])+[h|m|s])|None$"`

cluster/v1alpha1/helpers.go
haoqing0110: When the rollout strategy changes, it should trigger another reconcile to call GetRolloutCluster() I think. The returned "clusters to rollout" based on the current status snapshot.

cluster/v1alpha1/helpers.go
haoqing0110: Fixed.

cluster/v1alpha1/helpers.go
haoqing0110: Fixed.

cluster/v1beta1/helpers.go
haoqing0110: I'm not sure if need this, since if the user defines rollout strategy and has placement defined, the controller should watch the placement and decision. When the decision changes, should trigger a reconcile to call GetRolloutCluster().

cluster/v1alpha1/helpers.go
serngawy: yes, during progressive rollout if a policies(or manifestwork) continues to be in nonCompliant state we need Timeout to not stop applying the policy for other clusters.

cluster/v1alpha1/helpers.go
serngawy: yes, for some consumer controller (eg; TALM) even with ALL rollout it needs to set a state as Successful or Failed after Timeout (It will not continue to reconcile all the times, there is a maintenance window time constrain)

cluster/v1alpha1/helpers.go
serngawy: one thing may not clear here, if a cluster removed/added from the placement (even without changing the  rolloutStrategy) the consumer controller should track this behavior from placementDecisionTracker [here](https://github.com/open-cluster-management-io/api/pull/254/files#diff-a55f54ab06035f29ac70ef3b6c049983eef5c90f296007e3a6b6d6ca171f73b0R149)  
Is that correct ?

haoqing0110: Not sure where "here" points to (the URL doesn't work), the consumer controller needs to run pdTracker.Get() to refresh the placement decisions and will also get the added and deleted clusters.

serngawy: yap, the URL pointing to the pdTracker.Get() function.

cluster/v1alpha1/helpers.go
mprahl: It's not needed but it could be useful for the future.

cluster/v1alpha1/helpers.go
mprahl: Say for example the mandatory decision group completes and the controller has moved on to the next group (we'll call it group2). Then a new cluster is added to the mandatory decision group, `GetRolloutCluster` would return just that single cluster that was added and the controller would ignore group2 until that new cluster finishes?

If that's the case, what happens to the partial rollout in `group2` if that new cluster in the mandatory decision group does not succeed?

cluster/v1alpha1/helpers.go
mprahl: Yes, I think that would be helpful.

cluster/v1alpha1/helpers.go
mprahl: I think `newstatus` always returns the correct status of either the same status or timeout. I think it may be better to just always use `newstatus`.

cluster/v1alpha1/helpers.go
mprahl: What I'm getting at is how is the controller to know when a timeout has occurred? I'd like this to be event driven as opposed to continuously polling `GetRolloutCluster`. My idea was a context with a timeout could be returned so the controller can be notified when the timeout has occurred.

cluster/v1alpha1/helpers.go
mprahl: Yes, but other than `None`, isn't your string a subset of the duration format? It seems to me that you could just call `parseDuration` when it's not `None`.

cluster/v1alpha1/helpers.go
haoqing0110: > GetRolloutCluster would return just that single cluster that was added and the controller would ignore group2 until that new cluster finishes?

Yes, I think so.

> If that's the case, what happens to the partial rollout in group2 if that new cluster in the mandatory decision group does not succeed?

If group2 is partially rollout and the new cluster fails, the group2 rollout progress won't continue (clusters already upgraded won't rollback) until the new cluster succeeds.

cluster/v1alpha1/helpers.go
haoqing0110: Have modified the code.

cluster/v1alpha1/helpers.go
haoqing0110: Need to think about it...

cluster/v1alpha1/helpers.go
haoqing0110: Makes sense. Have modified the code.

cluster/v1alpha1/helpers.go
haoqing0110: Try to return the timeout time in the ClusterRolloutStatus , so that the consumer controller can use this time to requeue the related resource like `syncCtx.Queue().AddAfter(key, duration)`
 
```
type ClusterRolloutStatus struct {
	// cluster group key
	// optional field
	GroupKey clusterv1beta1.GroupKey
	// rollout status
	// required field
	Status RolloutStatus
	// the last transition time of rollout status
	// this is used to calculate timeout for progressing and failed status
	// optional field
	LastTransitionTime *metav1.Time
	// the timeout time when status is progressing or failed
	// optional field
	TimeOutTime *metav1.Time
}
```

cluster/v1alpha1/helpers.go
```diff
@@ -99,171 +98,179 @@ func (r *RolloutHandler) GetRolloutCluster(rolloutStrategy RolloutStrategy, stat
...
 		status := statusFunc(cluster)
-		if groupKey, exist := clusterToGroupKey[cluster]; exist {
+		if groupKey, exists := clusterToGroupKey[cluster]; exists {
 			status.GroupKey = groupKey
 		}
 
-		newstatus, needtorollout := needToRollout(status, timeout)
-		status.Status = newstatus.Status
-		status.TimeOutTime = newstatus.TimeOutTime
+		newStatus, needToRollout := needToRollout(status, timeout)
```
mprahl: It's confusing that `needToRollout` is the same name as the `needToRollout` function.

haoqing0110: Update to `newStatus, needToRollout := determineRolloutStatusAndContinue(status, timeout)`

mprahl (approved):
This looks good to me! We can always iterate on this later once the policy integration design is done and implementation has started.

mprahl:
Sorry @haoqing0110 , I didn't think my approval would count as a `lgtm` to cause it to merge the PR.

haoqing0110:
@mprahl looks good, github is so smart now.

cluster/v1alpha1/helpers.go
```diff
@@ -0,0 +1,379 @@
...
+// RolloutStatus represents the status of a rollout operation.
+type RolloutStatus int
+
+const (
+	// ToApply indicates that the resource's desired status has not been applied yet.
+	ToApply RolloutStatus = iota
+	// Progressing indicates that the resource's desired status is applied and last applied status is not updated.
+	Progressing
+	// Succeed indicates that the resource's desired status is applied and last applied status is successful.
+	Succeed
```
dhaiducek, line 31: Nit: I think this would fit better as "Succeeded"

cluster/v1beta1/helpers_test.go
```diff
@@ -388,201 +388,320 @@ func TestGetClusterSetsOfCluster(t *testing.T) {
 	}
 }
 
-type placementDecisionGetter struct {
-	client client.Client
+type FakePlacementDecisionGetter struct {
+	FakeDecisions []*PlacementDecision
```
dhaiducek, line 392: What's the purpose for this being an array of pointers? It seems it'd be much more straightforward for the users to not have the pointer there, but maybe I'm missing an implementation detail.

dhaiducek (commented):
Really late the the party here, sorry. I have a couple comments/questions as I'm starting to poke at the library. There may be more as I dive deeper.

cluster/v1beta1/helpers_test.go
haoqing0110: This is the testing code to provide a fake placementDecisionGetter to easily initiate some decisions. 
In the real code, the PlacementDecisionGetter would be like this https://github.com/open-cluster-management-io/ocm/blob/main/pkg/work/helper/helpers.go#L475

haoqing0110: @dhaiducek ^^^^

dhaiducek: Thanks for the link. I wasn't aware of the cluster lister. The purpose of the pointer is that it's grabbing it from the cache rather than the API server. I figured it might be something like that but hadn't found the code:
https://github.com/open-cluster-management-io/api/blob/2b2399b5f6e88e28a962aa82307ccc00bc492009/client/cluster/listers/cluster/v1beta1/placementdecision.go#L35-L38

haoqing0110: The placementDecisionGetter needs to be implemented by developer self, needs to implement List() interface. 

In the example of https://github.com/open-cluster-management-io/ocm/blob/main/pkg/work/helper/helpers.go#L475, the clusterlister is passed from https://github.com/open-cluster-management-io/ocm/blob/main/pkg/work/hub/controllers/manifestworkreplicasetcontroller/manifestworkreplicaset_controller.go#L120 `placementInformer.Lister()`.

cluster/v1alpha1/helpers.go
dhaiducek: I've filed a PR in case there's an interest in updating this: https://github.com/open-cluster-management-io/api/pull/267

PR #255, opened 2023-07-20
✨ Add mwrSet rolloutStrategy

serngawy:


work/v1alpha1/types_manifestworkreplicaset.go
```diff
@@ -86,6 +90,23 @@ type LocalPlacementReference struct {
 	// +kubebuilder:validation:Required
 	// +kubebuilder:validation:MinLength=1
 	Name string `json:"name"`
+
+	// +optional
```
haoqing0110, line 94: Should here add `// +kubebuilder:default={type: All}` ?

work/v1alpha1/types_manifestworkreplicaset.go
```diff
@@ -86,6 +90,23 @@ type LocalPlacementReference struct {
...
+	// PlacementRef Name
+	Name string `json:"name"`
+
+	// availableClusters shows number of clusters having manifestWorks available regards total number of clusters
+	// ex; 28/100 (28 out of 100)
+	AvailableClusters string `json:"availableClusters"`
+
+	// availableDecisionGroups shows number of decisionGroups having manifestWorks available regards total number of decisionGroups
+	// ex; 2/4 (2 out of 4)
+	AvailableDecisionGroups string `json:"availableDecisionGroups"`
```
haoqing0110, line 106: Can have a discussion here, whether need a common way to show the rollout progress. Today, addon shows this message in condition Progressing.
```
status:
  installProgression:
  - name: aws-placement
    namespace: default
    configReferences
...
    conditions:
    - lastTransitionTime: "2022-11-02T05:48:12Z"
      message: 200/400 upgrading...
      reason: Upgrading
      status: "True"
      type: Progressing
```

serngawy: Good to have common condition but I guess it hard to happen. Like in ManifestWorkReplicaSet there is no **progressing** condition neither **upgrading** reason.  As well as the Manifest has only apply/processing/available state.

work/v1alpha1/types_manifestworkreplicaset.go
serngawy: its already there in the Rollout API definition [here](https://github.com/open-cluster-management-io/api/blob/main/cluster/v1alpha1/types_rolloutstrategy.go#L27)

haoqing0110: Adding the default value on the top of RolloutStrategy is to ensure even no RolloutStrategy field defined in yaml, 
there's will be RolloutStrategy and "type: All" generated.

serngawy: okay, done.

work/v1alpha1/types_manifestworkreplicaset.go
```diff
@@ -86,6 +90,24 @@ type LocalPlacementReference struct {
...
+type PlacementSummary struct {
+	// PlacementRef Name
+	Name string `json:"name"`
+
+	// availableClusters shows number of clusters having manifestWorks available regards total number of clusters
+	// ex; 28/100 (28 out of 100)
+	AvailableClusters string `json:"availableClusters"`
+
+	// availableDecisionGroups shows number of decisionGroups having manifestWorks available regards total number of decisionGroups
+	// ex; 2/4 (2 out of 4)
```
qiujian16, line 105: is x/x a common format? how to validate it and do you expect some other controller to recognize it?

work/v1alpha1/types_manifestworkreplicaset.go
qiujian16: I think we might need to introduce progressing/upgrading in ManifestWorkReplicaSet for rollout strategy. We should be able to check whether a manifestwork is applied or not already (spec is changed but not applied yet). And in the long term, we should be able to define progress/degraded condition for manifestwork.

serngawy: The [ManifestWorkReplicaSetSummary](https://github.com/open-cluster-management-io/api/blob/main/work/v1alpha1/types_manifestworkreplicaset.go#L93) showing the applied/progrssing/available/degraded state for all clusters manifestworks. The PlacementSummary is more to shows the placement rollout state. Not sure, Do we need to add manifestWorkSummary for each placementSummary  ?

work/v1alpha1/types_manifestworkreplicaset.go
serngawy: it is more like a descriptive message and yes, we can validate it.

work/v1alpha1/types_manifestworkreplicaset.go
serngawy: done, added the ManifestWorkReplicaSetSummary to the placementSummary so every placement will has its own summary info AND added a condition for RollOut.

work/v1alpha1/types_manifestworkreplicaset.go
qiujian16: can we have a more clear description of AvailableDecisionGroups? if not all manifestwork in this decision is available, is this an AvailableDecisionGroups?

serngawy: No, it count the groups that have all clusters manifestwork in available state. I updated the comment

PR #256, opened 2023-07-26
Add image registries annotation key

zhujian7:


zhujian7:
cc @qiujian16 @zhiweiyin318

cluster/v1/types.go
```diff
@@ -234,3 +234,18 @@ const (
 	// ClusterNameLabelKey is the key of a label to set ManagedCluster name.
 	ClusterNameLabelKey = "open-cluster-management.io/cluster-name"
 )
+
+const (
+	// ClusterImageRegistriesAnnotationKey is an annotation key on ManagedCluster to configure image override for addons
```
qiujian16, line 239: I think we should also  mention addonDeploymentConfig will override this.

zhujian7:
/unhold
cc @qiujian16 PTAL

cluster/v1/types.go
zhujian7: Fixed

PR #257, opened 2023-07-31
add add-on proxy support

elgnay:


addon/v1alpha1/types_addondeploymentconfig.go
```diff
@@ -45,6 +45,11 @@ type AddOnDeploymentConfigSpec struct {
 	//
 	// +optional
 	Registries []ImageMirror `json:"registries,omitempty"`
+
+	// ProxyConfig holds proxy settings for add-on agent on the managed cluster.
+	// Either nil or empty means no proxy settings is available.
+	// +optional
+	ProxyConfig *ProxyConfig `json:"proxyConfig,omitempty"`
```
qiujian16: why nil if it is the same as empty?

elgnay: It will be treated as no proxy configured for the add-on agent if any of the following happens
1). The user does not specify `proxyConfig`;
```
Spec: {}
```
2). The user specify an empty `ProxyConfig`;
```
Spec:
  ProxyConfig: {}
```

qiujian16: But even if with no pointer, and user set Spec: {}, the API returned will be just the same as:
```
Spec:
  ProxyConfig: {}
```
It seems to me that setting it as pointer is not necessary.

elgnay: That makes sense. I've updated the code as suggested.

PR #259, opened 2023-08-10
:sparkles: add cloudevents clients

skeeey:


skeeey:
refer to https://github.com/open-cluster-management-io/enhancements/pull/97

client/cloudevents/protocol/fake/client.go
```diff
@@ -0,0 +1,48 @@
+package fake
+
+import (
+	"context"
+	"fmt"
+
+	cloudevents "github.com/cloudevents/sdk-go/v2"
+
+	"github.com/cloudevents/sdk-go/v2/event"
```
morvencao: `cloudevents.Event` is alias of `event.Event`: https://github.com/cloudevents/sdk-go/blob/main/v2/alias.go#L29
Can we just use `cloudevents.Event`?

client/cloudevents/protocol/fake/client.go
```diff
@@ -0,0 +1,48 @@
...
+}
+
+func (c *CloudEventClient) StartReceiver(ctx context.Context, fn interface{}) error {
+	receiver, ok := fn.(func(ctx context.Context, evt cloudevents.Event))
+	if !ok {
+		return fmt.Errorf("unsupported receiver function %T", fn)
+	}
+
+	for _, evt := range c.receivedEvents {
+		receiver(context.TODO(), evt)
```
morvencao: Is this intent to use new context, instead of `ctx`?

client/cloudevents/agentclient.go
```diff
@@ -0,0 +1,287 @@
...
+	evt.SetTime(time.Now())
+	evt.SetSource(c.agentID)
+	evt.SetExtension(types.ExtensionClusterName, c.clusterName)
+
+	objs, err := c.lister.List(types.ListOptions{Source: types.SourceAll})
+	if err != nil {
+		return err
+	}
+
+	resources := &payload.ResourceVersionList{Versions: []payload.ResourceVersion{}}
```
morvencao: Nit: can we pre-allocate the  ResourceVersion slice with length of objects?
```golang
resources := &payload.ResourceVersionList{Versions: make([]payload.ResourceVersion, 0, len(objs))}
```

this can optimize memory usage, when there are two many objects, to avoid slice resize.
this also can apply to status hashes construction.

client/cloudevents/interface.go
```diff
@@ -0,0 +1,64 @@
...
+	// Each event should have the following extensions: `resourceid`, `resourceversion` and `clustername`.
+	// The source set the `deletiontimestamp` extension to indicate one resource object is deleting from a source.
+	// The agent set the `originalsource` extension to indicate one resource belonged to which source.
+	Encode(source string, eventType types.CloudEventType, obj T) (*cloudevents.Event, error)
+
+	// Decode a cloudevents event to a resource object.
+	Decode(event *cloudevents.Event) (T, error)
+}
+
+type StatusHashGetter[T ResourceObject] interface {
```
qiujian16: could this be a func instead of a struct?

client/cloudevents/agentclient.go
```diff
@@ -0,0 +1,287 @@
...
+	agentID          string
+	clusterName      string
+}
+
+func NewCloudEventAgentClient[T ResourceObject](
+	contextGetter protocol.CloudEventsContextGetter,
+	subClient, pubClient cloudevents.Client,
+	lister Lister[T],
+	statusHashGetter StatusHashGetter[T],
+	agentID, clusterName string,
```
qiujian16: needs explanation on the input var. e.g. what does agentId, clusterName means. It seems to me agentId is actually source. We should not assume the client can only be used by agent. Also what if client is use by the consumer on server, it should not set clusterName when init the client.

client/cloudevents/agentclient.go
```diff
@@ -0,0 +1,287 @@
...
+		clusterName:      clusterName,
+	}
+}
+
+func (c *CloudEventAgentClient[T]) WithCodec(gvr types.GroupVersionResource, codec Codec[T]) {
+	c.codecs[gvr] = codec
+}
+
+// Resync the resources spec of the current agent by sending a spec resync request to sources
+func (c *CloudEventAgentClient[T]) Resync(ctx context.Context, eventGVR types.GroupVersionResource) error {
```
qiujian16: why resync spec is public but resync status is private.

client/cloudevents/agentclient.go
```diff
@@ -0,0 +1,287 @@
...
+	}
+
+	evt := cloudevents.NewEvent()
+	evt.SetID(uuid.New().String())
+	evt.SetType(eventType.String())
+	evt.SetTime(time.Now())
+	evt.SetSource(c.agentID)
+	evt.SetExtension(types.ExtensionClusterName, c.clusterName)
+
+	objs, err := c.lister.List(types.ListOptions{Source: types.SourceAll})
```
qiujian16: who will sent this request? I am not quite clear on why here is sourceall but we do not set cluster either.

client/cloudevents/sourceclient.go
```diff
@@ -0,0 +1,315 @@
...
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/klog/v2"
+
+	"open-cluster-management.io/api/client/cloudevents/payload"
+	"open-cluster-management.io/api/client/cloudevents/protocol"
+	"open-cluster-management.io/api/client/cloudevents/types"
+)
+
+type CloudEventSourceClient[T ResourceObject] struct {
```
qiujian16: comment please. I think what you mean is this client is used by server but not agent？

client/cloudevents/sourceclient.go
```diff
@@ -0,0 +1,315 @@
...
+	}
+
+	evt := cloudevents.NewEvent()
+	evt.SetID(uuid.New().String())
+	evt.SetType(eventType.String())
+	evt.SetTime(time.Now())
+	evt.SetSource(c.sourceID)
+
+	// list objects from all clusters with the current source
+	objs, err := c.lister.List(types.ListOptions{ClusterName: types.ClusterAll})
```
qiujian16: why not set source id here?

client/cloudevents/sourceclient.go
```diff
@@ -0,0 +1,315 @@
...
+	contextGetter    protocol.CloudEventsContextGetter
+	subClient        cloudevents.Client
+	pubClient        cloudevents.Client
+	lister           Lister[T]
+	codecs           map[types.GroupVersionResource]Codec[T]
+	statusHashGetter StatusHashGetter[T]
+	sourceID         string
+}
+
+func NewCloudEventSourceClient[T ResourceObject](
```
qiujian16, line 39: comments on all input var

client/cloudevents/sourceclient.go
```diff
@@ -0,0 +1,315 @@
...
+	}
+
+	if lastStatusHash == currentStatusHash {
+		return evt, nil
+	}
+
+	return types.StatusModified, nil
+}
+
+func (c *CloudEventSourceClient[T]) getObj(resourceID string, objs []T) (obj T, exists bool) {
```
qiujian16: seems not need to be in the struct

client/cloudevents/sourceclient.go
```diff
@@ -0,0 +1,315 @@
...
+
+// Resync the resources status of the current source by sending a status resync request to agents
+func (c *CloudEventSourceClient[T]) Resync(ctx context.Context, eventGVR types.GroupVersionResource) error {
+	eventType := types.CloudEventType{
+		GroupVersionResource: eventGVR,
+		SubResource:          types.SubResourceStatus,
+		Action:               types.ResyncRequestAction,
+	}
+
+	evt := cloudevents.NewEvent()
```
qiujian16: i think we should have a wrapper of NewEvent with generated uuid and time, and type/source as input.

client/cloudevents/types/types.go
```diff
@@ -0,0 +1,97 @@
+package types
+
+import (
+	"fmt"
+	"strings"
+)
+
+const (
```
qiujian16, line 12: please add doc for all const.

client/cloudevents/types/types.go
```diff
@@ -0,0 +1,97 @@
...
+	// Defaults to all clusters.
+	ClusterName string
+
+	// Agent use the source ID to restrict the list of returned objects by their source ID.
+	// Defaults to all sources.
+	Source string
+}
+
+// GroupVersionResource identifies the type of the resource included in the cloud event payload.
+type GroupVersionResource struct {
```
qiujian16: we cannot use schema.GroupVersionResource directly?

skeeey: using the schema.GroupVersionResource

client/cloudevents/types/types.go
skeeey: added

client/cloudevents/sourceclient.go
skeeey: add a event builder for this

client/cloudevents/sourceclient.go
skeeey: add the source id here if we have a central store on the server, we need the source id

client/cloudevents/sourceclient.go
skeeey: yes, it is, the source and agent have different new func now

client/cloudevents/agentclient.go
skeeey: the agent will sent this, the agent need to resync its resource spec from all sources when it starts

client/cloudevents/agentclient.go
skeeey: resync status is used to handle the resync status request, it's a internal logical handler, the outside client doesn't call it, rename the func to `respondResyncStatusRequest` to avoid confusion

client/cloudevents/agentclient.go
skeeey: this only for agent, description is added

client/cloudevents/interface.go
skeeey: using a func instead

client/cloudevents/agentclient.go
skeeey: done

client/cloudevents/protocol/fake/client.go
skeeey: using `cloudevents.Event` instead

client/cloudevents/protocol/fake/client.go
skeeey: fixed, it should not use a new context

client/cloudevents/sourceclient.go
skeeey: added

client/cloudevents/sourceclient.go
skeeey: removed from the struct

morvencao (commented):
Looks good to me, thanks.

client/cloudevents/agentclient.go
```diff
@@ -0,0 +1,303 @@
...
+// NewCloudEventAgentClient returns a client for an agent to resync/send/receive its resources with cloud events.
+//
+// An agent is a component that handles the deployment of requested resources on the managed cluster and status report
+// to the source.
+//
+// The following arguments are required to create a client
+//   - cloudEventContextGetter gets a cloudevents context when sending a event, different protocol should have its own
+//     context.
+//   - cloudEventsSubClient is a cloudevents client, it is used to receive the event from a broker.
+//   - cloudEventsPubClient is a cloudevents client, it is used to send the event from a broker.
```
qiujian16:
```suggestion
//   - cloudEventsPubClient is a cloudevents client, it is used to send the event to a broker.
```

client/cloudevents/sourceclient.go
qiujian16: still need a comment on what CloudEventSourceClient indicates.

client/cloudevents/sourceclient.go
```diff
@@ -0,0 +1,320 @@
...
+	}
+	if result := c.pubClient.Send(cloudeventsCtx, evt); cloudevents.IsUndelivered(result) {
+		return fmt.Errorf("failed to send: %v", result)
+	}
+
+	klog.V(4).Infof("Sent resync request:\n%s", evt)
+	return nil
+}
+
+// Publish a resource spec from a source to a spec.
```
qiujian16:
```suggestion
// Publish a resource spec from a source to an agent.
```

qiujian16: how to know which agent to send?

client/cloudevents/sourceclient.go
```diff
@@ -0,0 +1,320 @@
...
+	}
+
+	klog.V(4).Infof("Sent event:\n%s", evt)
+	return nil
+}
+
+// Subscribe the events that are from the agent spec resync request or agent resource status request.
+// For spec resync request, source publish the current resources spec back as response.
+// For resource status request, source receives resource status and handles the status with resource handlers.
+func (c *CloudEventSourceClient[T]) Subscribe(ctx context.Context, handlers ...ResourceHandler[T]) error {
```
qiujian16, line 144: do we need to set the sourceid in the context or it will be don implicitly by the client?

client/cloudevents/sourceclient.go
qiujian16: there seems to be something missing, e.g how to setup the sub/pub client for agent and source client, will it be same context or they need to use the different context, who takes the responsibility to set the context?

client/cloudevents/sourceclient.go
skeeey: we don't, the sourceid will be get from the event

client/cloudevents/sourceclient.go
skeeey: the event includes the cluster name extension, when we send the event, the CloudEventsClient will get the cluster name to determine the event will be sent to which cluster, e.g. for mqtt client, it use the cluster name to create the topic

client/cloudevents/sourceclient.go
skeeey: I try to introduce a `CloudEventsClient` interface to avoid using the cloudevents clients and context directly, please take a look

client/cloudevents/sourceclient.go
qiujian16: a client for a certain source should only subscribe to a certain source, isn't it?

qiujian16 (commented):
I think should also add a README in the client/cloudevents to describe how the client can be used.

client/cloudevents/sourceclient.go
skeeey: yes, for a source, it only subscribe its own resource, for mqtt, it limits by the topic, e.g. `/sources/maestro/clusters/+/status`

client/cloudevents/protocol/protocol.go
```diff
@@ -0,0 +1,22 @@
...
+
+	cloudevents "github.com/cloudevents/sdk-go/v2"
+	"open-cluster-management.io/api/client/cloudevents/protocol/mqtt"
+)
+
+// CloudEventsClient is used to send/receive cloudevents. Each event protocol has its own implementation.
+//
+// Available implementations:
+//   - MQTT
+type CloudEventsClient interface {
```
qiujian16: I am thinking the interface should be like

```
type CloudEventsClient interface {
    Sender() cloudevents.Client
    Subscriber() cloudevents.Client
    WithContext(ctx context.Context, opts) context.Context
}
```

and the mqtt.options returns the interface with

options.Source()
options.Agent()

yanmxa: nit:
  ```go
  // wording
  Sender -> Receiver
  Publisher -> Subscriber
  
  // dynamic parameters?
  WithContext(ctx context.Context, opts ...interface{}) context.Context
  ```

skeeey: Use `CloudEventsSourceOptions` and `CloudEventsAgentOptions` instead of this interface when building a cloudeventsclient

qiujian16 (commented):
/approve
/lgtm
/hold

I think it is a good start. I am still thinking if statusHashGetter should be in the interface of codec, but we can decide this later. Would @morvencao and @yanmxa to review also? if there is no additional comment, feel free to unhold

skeeey:
> I think should also add a README in the client/cloudevents to describe how the client can be used.

I will open a new pr to add the README after we introduced the work part

yanmxa:
Thanks! 
Looks good to me.

PR #260, opened 2023-08-16
🌱 addon consume common rollout api

haoqing0110:


addon/v1alpha1/types_clustermanagementaddon.go
```diff
@@ -187,89 +187,9 @@ type PlacementStrategy struct {
 	Configs []AddOnConfig `json:"configs,omitempty"`
 	// The rollout strategy to apply addon configurations change.
 	// The rollout strategy only watches the addon configurations defined in ClusterManagementAddOn.
-	// +kubebuilder:default={type: UpdateAll}
+	// +kubebuilder:default={type: All}
```
qiujian16, line 190: this default value change might be breaking for the existing API

haoqing0110: Yes, the existing cma will still hold value UpdateAll. 
This will impact the cma that claims managed by addon-manager, and addon-manager-controller will report error `err: incorrect rollout strategy type UpdateAll` . 

Since the addon-manager is an experimental feature in the last release, might only need to doc this breaking change?

PR #265, opened 2023-08-24
:seedling: add api upgrade flow

haoqing0110:
## Related issue(s)

Fixes #

### DIFF
--- a/docs/development.md
+++ b/docs/development.md
@@ -69,12 +69,14 @@ Once you decided to introduce a new version API, you will also need to consider 
 Below is the suggested API migration flow, the example code is to migrate `ManagedClusterSet` API from v1beta1 to v1beta2.
 
 ### Relase N (add new version)
-- Add new version v1beta2 API to api repo. For example:
+- Update the CRD to include the new version v1beta2. 
+
+  Set served version as both v1beta1 and v1beta2, keep storage version as v1beta1. For example:
   - https://github.com/open-cluster-management-io/api/pull/174
 
-- Convert between old and new versions.
+- Pick a conversion strategy.
 
-  At this stage, both v1beta1 and v1beta2 are in served versions and storage version is v1beta1. If v1beta2 schema not change or can be converted to v1beta1, the CRD storage version can remain unchanged. If the v1beta2 schema changes and requires custom logic, you need to add conversion webhook to transform CRs between v1beta1 and v1beta2. For example: 
+  At this stage, both v1beta1 and v1beta2 are in served versions and storage versoin is v1beta1. The custom resource objects must sometimes be converted between the version they are stored at and the version they are served at. If there are no schema changes, the default None conversion strategy may be used and only the apiVersion field will be modified when serving different versions. If the v1beta2 schema changes and requires custom logic, you need to add conversion webhook to transform CRs between v1beta1 and v1beta2. For example: 
   - https://github.com/open-cluster-management-io/registration/pull/272
   - https://github.com/open-cluster-management-io/registration-operator/pull/279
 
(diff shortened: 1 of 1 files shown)


haoqing0110:
cc @zhujian7 we might work together to refine the API migration flow.

haoqing0110:
cc @qiujian16 @elgnay

haoqing0110:
/assign @deads2k 

Hello @deads2k , we have a summary of the API upgrade flow based on our experience in previous releases, hope you can help review and provide some comments, thanks a lot!

docs/development.md
```diff
@@ -60,4 +60,29 @@ To generate `zz_generated.deepcopy.go` & `zz_generated.swagger_doc_generated.go`
...
+When introducing a new version API, you need to deprecate the old version API and then remove it. 
+Below is the suggested API migration flow, the example code is to migrate `ManagedClusterSet` API from v1beta1 to v1beta2.
+
+### Relase N (add new version and depracated old version)
+- Add new version v1beta2 API and mark v1beta1 as deprecated. For example: 
+  - https://github.com/open-cluster-management-io/api/pull/174
+- Add conversion webhook to transform CRs between v1beta1 and v1beta2. (optional) For example: 
+  - https://github.com/open-cluster-management-io/registration/pull/272
+  - https://github.com/open-cluster-management-io/registration-operator/pull/279
+- Change internal consumers(ui/foundation/submarinar-addon/placement) to use clusterset api v1beta2.
```
qiujian16: these cannot be treated as internal consumers. They are actually external consumers.

docs/development.md
```diff
@@ -60,4 +60,29 @@ To generate `zz_generated.deepcopy.go` & `zz_generated.swagger_doc_generated.go`
...
+- Other clusterset consumers (external consumers) can upgrade to clusterset api v1beta2
```
qiujian16: I think consumers can start migrate, but I think the migration should start at N+1

docs/development.md
```diff
@@ -60,4 +60,29 @@ To generate `zz_generated.deepcopy.go` & `zz_generated.swagger_doc_generated.go`
...
+
+### Relase N+1 (migrate old version to new version)
+- Migrate stored version of clusterset api to v1beta2 and remove v1beta1 from stored version. For example: 
+  - https://github.com/open-cluster-management-io/api/pull/202
+  - https://github.com/open-cluster-management-io/registration-operator/pull/315
+  - https://github.com/open-cluster-management-io/registration/pull/297
+- Other clusterset consumers (external consumers) must upgrade to clusterset api v1beta2.
+
+### Relase N+2 (remove old version)
```
qiujian16: We cannot say we MUST remove old version in N+2, it depends on the maturity of the API and consumers.

docs/development.md
```diff
@@ -60,4 +60,29 @@ To generate `zz_generated.deepcopy.go` & `zz_generated.swagger_doc_generated.go`
...
-Before you commit you changes, please run `make verify` locally to make sure you have generated all required files.
\ No newline at end of file
+Before you commit you changes, please run `make verify` locally to make sure you have generated all required files.
+
+## API upgrade flow
+When introducing a new version API, you need to deprecate the old version API and then remove it. 
+Below is the suggested API migration flow, the example code is to migrate `ManagedClusterSet` API from v1beta1 to v1beta2.
+
+### Relase N (add new version and depracated old version)
+- Add new version v1beta2 API and mark v1beta1 as deprecated. For example:
```
qiujian16: My understanding is we mark it deprecated when we change the stored version?

haoqing0110: Sounds makes sense. 

In our previous releases, we tended to add new API and mark the old version deprecated first, letting the user know it's deprecated for at least 1 release and then start migration in the next release. 

Now we have some experience in migration, it's okay to mark it deprecated when changing the stored version in release N+1.

docs/development.md
haoqing0110: have updated the doc

docs/development.md
haoqing0110: have updated the doc

docs/development.md
haoqing0110: have updated the doc

docs/development.md
haoqing0110: have updated the doc, also with some tips on how to set the served and storage version.

qiujian16:
also need to check here https://github.com/open-cluster-management-io/community/blob/main/sig-architecture/api_changes.md on what is updated.

haoqing0110:
https://github.com/open-cluster-management-io/community/blob/main/sig-architecture/api_changes.md is a useful addition to this flow, append some URLs to the related part.

docs/development.md
```diff
@@ -60,4 +60,44 @@ To generate `zz_generated.deepcopy.go` & `zz_generated.swagger_doc_generated.go`
...
+  - https://github.com/open-cluster-management-io/registration/pull/272
+  - https://github.com/open-cluster-management-io/registration-operator/pull/279
+
+- The consumers (eg, ui/foundation/submarinar-addon/placement) CAN start to migrate to clusterset api v1beta2.
+
+- See also [How to add a new version](https://github.com/open-cluster-management-io/community/blob/main/sig-architecture/api_changes.md#how-to-add-a-new-version)
+
+### Relase N+1 (deprecated old version and start migration)
+- Mark v1beta1 as deprecated, served versions are v1beta1 and v1beta2, set storage: true on v1beta2.
+
```
xuezhaojun, line 88: Do we consider to sperate: [set storage: true on v1beta2] and [migrate stored version of clusterset api to v1beta2] into 2 releases?

Release N+1: 
- Mark v1beta1 as deprecated, served versions are v1beta1 and v1beta2, set storage: true on v1beta2.

Release N+2:
- Migrate stored version of clusterset api to v1beta2 and remove v1beta1 from CRD `status.storedVersions`.

---

The motivation behind this is: if we do 2 things all in Release N+1, then we need to make sure storage set to v1beta2 happens first, StorageVersionMigration get created second.

If any StorageVersionMigration is created before the storage: true move to v1beta2, the migration will end up with Succeeded condition but still using v1beta1 in the etcd.

Doing these 2 things in 2 releases can decouple the order dependency between 2 steps.

xuezhaojun: @qiujian16  Please take another review, should we put this flow into best-practice?

xuezhaojun: The risk we concerns here is this flow hasn't been verified by history. We never process in this way before, it's just a design.

haoqing0110:
@qiujian16 @elgnay @xuezhaojun plz help do a final review.

docs/development.md
```diff
@@ -60,4 +60,60 @@ To generate `zz_generated.deepcopy.go` & `zz_generated.swagger_doc_generated.go`
...
+- Ensure the new CRD is applied first, then create a `StorageVersionMigration` resource to migrate stored version to v1beta2, and then remove v1beta1 from CRD `status.storedVersions`. 
+
+  **The migration sequence is important:**
+  - Ensure the new CRD (set storage: true on v1beta2) is applied before you do the migration. If the order is reversed, the migration will not take effect.
+  - `StorageVersionMigration` is running like a job. If you need to run the migration again for other version in the future, keep the same resource name and just change the spec won't trigger a new migration. So strongly suggest you append version info into the `StorageVersionMigration` resource name, for example `managedclustersets.v1beta1-to-v1beta2.cluster.open-cluster-management.io`.
+  - Ensure all the stored version has migrated to v1beta2 before you remove v1beta1 from `status.storedVersion`.
+  - Suggest you do the migration and change the stored version in the same release.
+
+  In registration-operator it has a migration controller and crd status controller to handle the progress, code example: 
+  - https://github.com/open-cluster-management-io/registration-operator/pull/315
```
xuezhaojun, line 103: @haoqing0110 

When add the StorageVersionMigration, make the version filed equals to the version you want to migrate from. In terms of this example, the version you want to migration from is `v1beta1`, so the StorageVersionMigration should be:
```
apiVersion: migration.k8s.io/v1alpha1
kind: StorageVersionMigration
metadata:
  name: managedclustersetbindings.cluster.open-cluster-management.io
spec:
  resource:
    group: cluster.open-cluster-management.io
    resource: managedclustersets
    version: v1beta1 # The version must be v1beta1
```

haoqing0110, line 117: @xuezhaojun thanks, have updated this example in the doc.

docs/development.md
```diff
@@ -60,4 +60,74 @@ To generate `zz_generated.deepcopy.go` & `zz_generated.swagger_doc_generated.go`
...
+  - https://github.com/open-cluster-management-io/registration-operator/pull/332
+  - https://github.com/open-cluster-management-io/registration/pull/297
+
+- The consumers (eg, ui/foundation/submarinar-addon/placement) SHOULD migrate to clusterset api v1beta2 before it's removed.
+
+- See also [Migrate stored objects to the new version](https://github.com/open-cluster-management-io/community/blob/main/sig-architecture/api_changes.md#migrate-stored-objects-to-the-new-version)
+
+### Relase N+x (remove old version)
+
+when the new version API becomes mature, and ready to remove old version (for example, the consumers have adopted the new version API), remove the old version related code:
```
qiujian16: we need to clarify what READY means. What needs to be checked to ensure it is ready to remove the old version.

qiujian16:
/approve

@xuezhaojun leave to you for final review

PR #269, opened 2023-08-29
:sparkles: add payload for manifestwork

skeeey:
## Summary
add cloudevents payload for manifestwork


client/cloudevents/payload/manifestwork/mainfiest.go
```diff
@@ -0,0 +1,41 @@
...
+)
+
+var ManifestEventDataType = types.CloudEventsDataType{
+	Group:    "io.open-cluster-management.works",
+	Version:  "v1alpha1",
+	Resource: "manifests",
+}
+
+// SingleManifest represents the data in a cloudevent, it contains a single manifest.
+type SingleManifest struct {
```
morvencao: nit: do we need 'single', 'Manifest' is more simpler and straightforward?

client/cloudevents/types/types.go
```diff
@@ -19,6 +19,14 @@ const (
 	SourceAll = ""
 )
 
+const (
+	// CloudEventsDataTypeAnnotationKey is the key of the cloudevents data type annotation.
+	CloudEventsDataTypeAnnotationKey = "cloudevents.open-cluster-management.io/datatype"
```
morvencao: what's this annotation for?

client/cloudevents/payload/manifestwork/mainfiest.go
skeeey: Use Manifest as the struct name

client/cloudevents/types/types.go
skeeey: this will be used by work-agent to distinguish a manifestwork's event data type, move to work package

cloudevents/work/payload/mainfiest.go
```diff
@@ -0,0 +1,54 @@
...
+	// 4. Degraded represents the current state of manifest of a SingleManifest does not match the desired state for a
+	//    certain period.
+	// 5. Deleted represents the manifests of a SingleManifest is deleted from a managed cluster.
+	Conditions []metav1.Condition `json:"conditions"`
+
+	// Status represents the conditions of this manifest on a managed cluster.
+	Status *workv1.ManifestCondition `json:"status,omitempty"`
+}
+
+type ManifestConfigOption struct {
```
qiujian16, line 46: why not use ManifestConfigOption in work api? because of the resourcemeta field?

cloudevents/work/lister.go
```diff
@@ -0,0 +1,28 @@
...
+
+import (
+	"k8s.io/client-go/tools/cache"
+
+	"open-cluster-management.io/api/cloudevents/generic/types"
+	workv1 "open-cluster-management.io/api/work/v1"
+)
+
+// ManifestWorkLister list the ManifestWorks from a ManifestWorkInformer's local cache.
+type ManifestWorkLister struct {
```
qiujian16, line 12: how does this lister get initiated?

cloudevents/work/agent/handler/resourcehandler.go
```diff
@@ -0,0 +1,31 @@
...
+	"open-cluster-management.io/api/cloudevents/generic/types"
+	workv1 "open-cluster-management.io/api/work/v1"
+)
+
+// NewManifestWorkAgentHandler returns a ResourceHandler for a ManifestWork on managed cluster.
+func NewManifestWorkAgentHandler(ctx context.Context, workClient workv1client.ManifestWorkInterface) generic.ResourceHandler[*workv1.ManifestWork] {
+	return func(action types.ResourceAction, work *workv1.ManifestWork) error {
+		switch action {
+		case types.Added:
+			_, err := workClient.Create(ctx, work, metav1.CreateOptions{})
```
qiujian16: this is not clear to me, why the resource handler finally creates a manifestwork using a client? I thought you are going to put it in the cache.Store.

qiujian16 (commented):
shouldn't we have a implementation of workv1client.ManifestWorkInterface

cloudevents/work/payload/mainfiest.go
skeeey: yes, it is, for a single manifest, I think we alreay have a certain type resource, so we may not set the resourcemeta outside

cloudevents/work/lister.go
skeeey: when we build the work clients for cloud events, we use inform store to initiate it

cloudevents/work/agent/handler/resourcehandler.go
skeeey: yes, it will be put in the store finally, we will have a new client, it implements the `ManifestWorkInterface`, the client will put the manifestwork to the sore, the logic will be same with here, so I used the client to handle this

skeeey:
> shouldn't we have a implementation of workv1client.ManifestWorkInterface

~Yes, we need, I will open a new pr to implement the interface after this pr~

will add it

cloudevents/work/agent/handler/resourcehandler.go
qiujian16: it is confusing I think, if it is a certain type client, we should use that as the input instead of the interface.

cloudevents/work/lister.go
qiujian16: if it is from informer, why we not use lister directly but using cache.store?

skeeey: yeah, a lister can be used here, I will use lister instead of store

cloudevents/work/agent/handler/resourcehandler.go
skeeey: hrm... I do not introduce the new client in this pr yet, I will add the implementation, and use a certain type client

cloudevents/work/agent/client/manifestwork.go
```diff
@@ -0,0 +1,195 @@
...
+		return nil, err
+	}
+	return updatedWork, nil
+}
+
+func (c *ManifestWorksAgentClient) Delete(ctx context.Context, name string, opts metav1.DeleteOptions) error {
+	klog.V(4).Infof("deleting manifestwork %s", name)
+	work, err := c.lister.Get(name)
+	if err != nil {
+		return err
```
qiujian16: should ignore isnotfound?

cloudevents/work/agent/client/manifestwork.go
```diff
@@ -0,0 +1,195 @@
...
+	}
+
+	c.watcher.Receive(watch.Event{Type: watch.Modified, Object: work})
```
qiujian16: what is the intension of this? and I do not think agent should have the permission to delete?

cloudevents/work/agent/client/manifestwork.go
```diff
@@ -0,0 +1,195 @@
...
+	}
+
+	return &workv1.ManifestWorkList{}, nil
+}
+
+func (mw *ManifestWorksAgentClient) Watch(ctx context.Context, opts metav1.ListOptions) (watch.Interface, error) {
+	return mw.watcher, nil
+}
+
+func (mw *ManifestWorksAgentClient) Patch(ctx context.Context, name string, pt kubetypes.PatchType, data []byte, opts metav1.PatchOptions, subresources ...string) (result *workv1.ManifestWork, err error) {
```
qiujian16, line 139: we have starting to use patch a lot in ocm repo.

cloudevents/work/agent/client/manifestwork.go
```diff
@@ -0,0 +1,196 @@
...
+}
+
+func (c *ManifestWorksAgentClient) Get(ctx context.Context, name string, opts metav1.GetOptions) (*workv1.ManifestWork, error) {
+	klog.V(4).Infof("getting manifestwork %s", name)
+	return c.lister.Get(name)
+}
+
+func (c *ManifestWorksAgentClient) List(ctx context.Context, opts metav1.ListOptions) (*workv1.ManifestWorkList, error) {
+	klog.V(4).Infof("sync manifestworks")
+	// send resync request to fetch manifestworks from source
```
qiujian16: this is tricky, we only use list to resync?

cloudevents/work/agent/client/manifestwork.go
```diff
@@ -0,0 +1,196 @@
...
+const ManifestsDeleted = "Deleted"
+
+const (
+	UpdateRequestAction = "update_request"
+	DeleteRequestAction = "delete_request"
+)
+
+var manifestWorkGR = schema.GroupResource{Group: workv1.GroupName, Resource: "manifestworks"}
+
+// ManifestWorkWatcher implements the ManifestWorkInterface. It receives manifestworks spec from source by
```
qiujian16:
```suggestion
// ManifestWorksAgentClient implements the ManifestWorkInterface. It receives manifestworks spec from source by
```

cloudevents/work/agent/client/manifestwork.go
skeeey: this only happens when the agent starts, if we want to resync in other time, I think we need call cloudevents resync function, or we don't do resync here? we provide a way to do resync when we build the clients

cloudevents/work/agent/client/manifestwork.go
skeeey: good catch, we don't need the delete function for agent, we just need mark the manifest has the deletiontimestamp, this should be finished by update function

cloudevents/work/agent/client/manifestwork.go
skeeey: ~add todo~

need to implement this function

skeeey:
now ManifestWorkResourceHandler send kube events to handle the works in local and ManifestWorksAgentClient sends  works status to source

PR #271, opened 2023-08-31
:sparkles: implement patch function in ManifestWorkAgentClient

skeeey:
## Summary
implement path function in ManifestWorkAgentClient

### DIFF
--- a/cloudevents/work/agent/client/manifestwork.go
+++ b/cloudevents/work/agent/client/manifestwork.go
@@ -4,8 +4,10 @@ import (
 	"context"
 	"fmt"
 
+	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	kubetypes "k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/watch"
 	"k8s.io/klog/v2"
@@ -15,6 +17,7 @@ import (
 	"open-cluster-management.io/api/cloudevents/generic"
 	"open-cluster-management.io/api/cloudevents/generic/types"
 	"open-cluster-management.io/api/cloudevents/work/agent/codec"
+	"open-cluster-management.io/api/cloudevents/work/utils"
 	"open-cluster-management.io/api/cloudevents/work/watcher"
 	workv1 "open-cluster-management.io/api/work/v1"
 )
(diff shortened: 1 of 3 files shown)


cloudevents/work/agent/client/manifestwork.go
```diff
@@ -26,117 +27,123 @@ const (
...
 	lister            workv1lister.ManifestWorkNamespaceLister
 }
 
-var _ workv1client.ManifestWorkInterface = &ManifestWorksAgentClient{}
+var _ workv1client.ManifestWorkInterface = &ManifestWorkAgentClient{}
 
-func (c *ManifestWorksAgentClient) Create(ctx context.Context, manifestWork *workv1.ManifestWork, opts metav1.CreateOptions) (*workv1.ManifestWork, error) {
-	klog.Fatal("Create function for ManifestWorksAgentClient is unsupported")
+func (c *ManifestWorkAgentClient) Create(ctx context.Context, manifestWork *workv1.ManifestWork, opts metav1.CreateOptions) (*workv1.ManifestWork, error) {
+	klog.Fatal("Create function for ManifestWorkAgentClient is unsupported")
```
qiujian16: fatal seems not appropriate, should we just return not supported error here/.

cloudevents/work/utils/utils.go
```diff
@@ -0,0 +1,53 @@
...
+		if err != nil {
+			return nil, err
+		}
+
+	case types.MergePatchType:
+		patchedData, err = jsonpatch.MergePatch(workData, patchData)
+		if err != nil {
+			return nil, err
+		}
+	case types.StrategicMergePatchType:
```
qiujian16: StrategicMergePatchType is not supported in CRD.

cloudevents/work/agent/client/manifestwork.go
```diff
@@ -26,117 +27,123 @@ const (
...
 
-	updatedWork := manifestWork.DeepCopy()
+	eventDataType, err := types.ParseCloudEventsDataType(patchedWork.Annotations[codec.CloudEventsDataTypeAnnotationKey])
+	if err != nil {
+		return nil, err
+	}
 
 	eventType := types.CloudEventsType{
 		CloudEventsDataType: *eventDataType,
 		SubResource:         types.SubResourceStatus,
```
qiujian16, line 104: should we set SubResource after checking the sub resource?

cloudevents/work/agent/client/manifestwork.go
```diff
@@ -26,117 +27,123 @@ const (
...
+		c.watcher.Receive(watch.Event{Type: watch.Deleted, Object: newWork})
+		return newWork, nil
+	}
 
-func (c *ManifestWorksAgentClient) List(ctx context.Context, opts metav1.ListOptions) (*workv1.ManifestWorkList, error) {
-	klog.V(4).Infof("sync manifestworks")
-	// send resync request to fetch manifestworks from source when the ManifestWorkInformer status
-	if err := c.cloudEventsClient.Resync(ctx); err != nil {
-		return nil, err
+	if IsStatusUpdate(subresources) {
```
qiujian16: if it is not status subresource we should return error

cloudevents/work/agent/client/manifestwork.go
skeeey: return MethodNotSupported error

cloudevents/work/agent/client/manifestwork.go
skeeey: we don't need, we just send the work status to source here (deletion is also a status update for a source)

cloudevents/work/agent/client/manifestwork.go
skeeey: return the error at the end

cloudevents/work/utils/utils.go
skeeey: do we need handle the crd here? we patch a work here, if a work has a crd, we will stop here? or the caller should not use this type when they patch their works that contain crds?

skeeey: and it seems to be difficult for us to distinguish pacth a work or patch the crds in works

qiujian16: no I mean resource served by crd, in this case ManifestWork does not support StrategicMergePatchType

skeeey: OK, remvoe the StrategicMergePatchType support

cloudevents/work/agent/client/manifestwork.go
qiujian16: i mean it could be either patch the finalizer (patch spec) or patch status. We can not patch other sub resources.

skeeey: return the error if the patch is not for status or finalizers and allow update finalizers(add/remvoe)

PR #272, opened 2023-08-31
:seedling: Add `Ignore` rollout status

dhaiducek:
## Summary

An `Ignore` rollout status in my reading of it felt different than a `Skip` status, where `Ignore` would mean that it wouldn't factor into a rollout but would receive updates while `Skip` would mean it didn't receive updates, but I'm open to discussion on it.

I also have an optional commit here to add `RolloutType` to make the enum a bit more concrete.

## Related issue(s)

- #254

### DIFF
--- a/cluster/v1alpha1/helpers.go
+++ b/cluster/v1alpha1/helpers.go
@@ -33,6 +33,8 @@ const (
 	// TimeOut indicates that the rollout status is progressing or failed and the status remains
 	// for longer than the timeout, resulting in a timeout status.
 	TimeOut
+	// Ignore indicates that the resource's desired status is applied but the last applied status is irrelevant.
+	Ignore
 	// Skip indicates that the rollout should be skipped on this cluster.
 	Skip
 )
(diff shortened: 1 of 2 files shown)


haoqing0110:
For the skip status, when adding it, expect it to be a status return from ClusterRolloutStatusFunc by API consumers, indicating that the consumers do not want to do a rollout on this cluster. 
For the ignore status, not sure what does "receive updates" mean? in the PR seems ingore returns the same value as skip.

dhaiducek:
/hold

> For the skip status, when adding it, expect it to be a status return from ClusterRolloutStatusFunc by API consumers, indicating that the consumers do not want to do a rollout on this cluster. For the ignore status, not sure what does "receive updates" mean? in the PR seems ingore returns the same value as skip.

Thanks for the feedback. We want users to be able to specify that a particular cluster should get updated in the rollout but it doesn't matter to them whether the status is returned successful. It seems adding `Ignore` is confusing, so we're going to consider this further. For this it seems like we should just use `Succeeded` if it succeeds but add a `FailedNonBlocking` status instead of `Ignore` to say that the library is aware it failed during the rollout but is continuing on.

PR #273, opened 2023-09-01
:sparkles: a client builder for manifestworks

skeeey:
## Summary
a client builder for manifestworks

## Related issue(s)

Fixes #

### DIFF
--- a/cloudevents/work/clientbuilder.go
+++ b/cloudevents/work/clientbuilder.go
@@ -1,0 +1,159 @@
+package work
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"k8s.io/client-go/rest"
+	"k8s.io/klog/v2"
+
+	workclientset "open-cluster-management.io/api/client/work/clientset/versioned"
+	workv1client "open-cluster-management.io/api/client/work/clientset/versioned/typed/work/v1"
+	workinformers "open-cluster-management.io/api/client/work/informers/externalversions"
+	workv1informers "open-cluster-management.io/api/client/work/informers/externalversions/work/v1"
+	"open-cluster-management.io/api/cloudevents/generic"
+	"open-cluster-management.io/api/cloudevents/generic/options/mqtt"
+	agentclient "open-cluster-management.io/api/cloudevents/work/agent/client"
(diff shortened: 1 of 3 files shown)


cloudevents/work/clientbuilder.go
```diff
@@ -0,0 +1,157 @@
...
+		return nil, err
+	}
+
+	manifestWorkClient := agentclient.NewManifestWorkAgentClient(cloudEventsClient, watcher)
+	workClient := &internal.WorkV1ClientWrapper{ManifestWorkClient: manifestWorkClient}
+	workClientSet := &internal.WorkClientSetWrapper{WorkV1ClientWrapper: workClient}
+	factory := workinformers.NewSharedInformerFactoryWithOptions(workClientSet, b.informerResyncTime, b.informerOptions...)
+	informers := factory.Work().V1().ManifestWorks()
+	manifestWorkLister := informers.Lister()
+	namespacedLister := manifestWorkLister.ManifestWorks(b.clusterName)
```
qiujian16, line 139: I am worried that we have multiple thread write cache.Store at the same time. Is cache.Store thread safe? What happens if a thread is updating the spec and another is updating the status or finallizer?

skeeey: the informer cache is a [thread safe store](https://github.com/kubernetes/client-go/blob/v0.26.1/tools/cache/store.go#L273), when we update the spec/status/finallizer, all of these actions will be handled by the [reflector](https://github.com/kubernetes/client-go/blob/v0.26.1/tools/cache/reflector.go#L530) , the reflector will call its store to [update](https://github.com/kubernetes/client-go/blob/v0.26.1/tools/cache/thread_safe_store.go#L221) them, so I think it should be thread safe call

qiujian16: I think we should add some note here that client and informer is sharing the store in this implementation. We may need to revisit the implementation in the future. Ideally, the store should be only written from the server.

skeeey: OK, added a todo here

qiujian16:
/approve

leave to @morvencao to tag

morvencao (commented):
💯 
/lgtm

PR #276, opened 2023-09-11
🌱 Update rollout lib

serngawy:
## Related issue(s)

Fixes #

### DIFF
--- a/cluster/v1alpha1/helpers.go
+++ b/cluster/v1alpha1/helpers.go
@@ -37,11 +37,10 @@ const (
 	Skip
 )
 
-// ClusterRolloutStatusFunc defines a function to return the rollout status for a managed cluster.
-type ClusterRolloutStatusFunc func(clusterName string) ClusterRolloutStatus
-
 // ClusterRolloutStatus holds the rollout status information for a cluster.
 type ClusterRolloutStatus struct {
+	// cluster name
+	ClusterName string
 	// GroupKey represents the cluster group key (optional field).
 	GroupKey clusterv1beta1.GroupKey
 	// Status is the required field indicating the rollout status.
@@ -53,50 +52,57 @@ type ClusterRolloutStatus struct {
 	TimeOutTime *metav1.Time
 }
 
-// RolloutResult contains the clusters to be rolled out and the clusters that have timed out.
+// RolloutResult contains list of clusters that are timeOut, removed and required to rollOut
 type RolloutResult struct {
-	// ClustersToRollout is a map where the key is the cluster name and the value is the ClusterRolloutStatus.
-	ClustersToRollout map[string]ClusterRolloutStatus
-	// ClustersTimeOut is a map where the key is the cluster name and the value is the ClusterRolloutStatus.
-	ClustersTimeOut map[string]ClusterRolloutStatus
+	// ClustersToRollout is a slice of ClusterRolloutStatus that will be rolled out.
+	ClustersToRollout []ClusterRolloutStatus
+	// ClustersTimeOut is a slice of ClusterRolloutStatus that are timeout.
+	ClustersTimeOut []ClusterRolloutStatus
+	// ClustersRemoved is a slice of ClusterRolloutStatus that are removed.
+	ClustersRemoved []ClusterRolloutStatus
 }
 
+// ClusterRolloutStatusFunc defines a function that return the rollout status for a given workload.
+type ClusterRolloutStatusFunc[T any] func(clusterName string, workload T) (ClusterRolloutStatus, error)
+
+// The RolloutHandler required workload type (interface/struct) to be assigned to the generic type.
+// The custom implementation of the ClusterRolloutStatusFunc is required to use the RolloutHandler.
 // +k8s:deepcopy-gen=false
-type RolloutHandler struct {
+type RolloutHandler[T any] struct {
 	// placement decision tracker
-	pdTracker *clusterv1beta1.PlacementDecisionClustersTracker
+	pdTracker  *clusterv1beta1.PlacementDecisionClustersTracker
+	statusFunc ClusterRolloutStatusFunc[T]
 }
 
-func NewRolloutHandler(pdTracker *clusterv1beta1.PlacementDecisionClustersTracker) (*RolloutHandler, error) {
+// NewRolloutHandler creates a new RolloutHandler with the give workload type.
+func NewRolloutHandler[T any](pdTracker *clusterv1beta1.PlacementDecisionClustersTracker, statusFunc ClusterRolloutStatusFunc[T]) (*RolloutHandler[T], error) {
 	if pdTracker == nil {
 		return nil, fmt.Errorf("invalid placement decision tracker %v", pdTracker)
 	}
 
-	return &RolloutHandler{pdTracker: pdTracker}, nil
+	return &RolloutHandler[T]{pdTracker: pdTracker, statusFunc: statusFunc}, nil
 }
 
-// The input is a duck type RolloutStrategy and a ClusterRolloutStatusFunc to return the rollout status on each managed cluster.
-// Return the strategy actual take effect and a list of clusters that need to rollout and that are timeout.
+// The input are a RolloutStrategy and existingClusterRolloutStatus list.
+// The existing ClusterRolloutStatus list should be created using the ClusterRolloutStatusFunc to determine the current workload rollout status.
+// The existing ClusterRolloutStatus list should contain all the current workloads rollout status such as ToApply, Progressing, Succeeded,
+// Failed, TimeOut and Skip in order to determine the added, removed, timeout clusters and next clusters to rollout.
 //
-// ClustersToRollout: If mandatory decision groups are defined in strategy, will return the clusters to rollout in mandatory decision groups first.
-// When all the mandatory decision groups rollout successfully, will return the rest of the clusters that need to rollout.
-//
-// ClustersTimeOut: If the cluster status is Progressing or Failed, and the status lasts longer than timeout defined in strategy,
-// will list them RolloutResult.ClustersTimeOut with status TimeOut.
-func (r *RolloutHandler) GetRolloutCluster(rolloutStrategy RolloutStrategy, statusFunc ClusterRolloutStatusFunc) (*RolloutStrategy, RolloutResult, error) {
+// Return the actual RolloutStrategy that take effect and a RolloutResult contain list of ClusterToRollout, ClustersTimeout and ClusterRemoved.
+func (r *RolloutHandler[T]) GetRolloutCluster(rolloutStrategy RolloutStrategy, existingClusterStatus []ClusterRolloutStatus) (*RolloutStrategy, RolloutResult, error) {
 	switch rolloutStrategy.Type {
 	case All:
-		return r.getRolloutAllClusters(rolloutStrategy, statusFunc)
+		return r.getRolloutAllClusters(rolloutStrategy, existingClusterStatus)
 	case Progressive:
-		return r.getProgressiveClusters(rolloutStrategy, statusFunc)
+		return r.getProgressiveClusters(rolloutStrategy, existingClusterStatus)
 	case ProgressivePerGroup:
-		return r.getProgressivePerGroupClusters(rolloutStrategy, statusFunc)
+		return r.getProgressivePerGroupClusters(rolloutStrategy, existingClusterStatus)
 	default:
 		return nil, RolloutResult{}, fmt.Errorf("incorrect rollout strategy type %v", rolloutStrategy.Type)
 	}
 }
 
-func (r *RolloutHandler) getRolloutAllClusters(rolloutStrategy RolloutStrategy, statusFunc ClusterRolloutStatusFunc) (*RolloutStrategy, RolloutResult, error) {
+func (r *RolloutHandler[T]) getRolloutAllClusters(rolloutStrategy RolloutStrategy, existingClusterStatus []ClusterRolloutStatus) (*RolloutStrategy, RolloutResult, error) {
 	// Prepare the rollout strategy
 	strategy := RolloutStrategy{Type: All}
 	strategy.All = rolloutStrategy.All.DeepCopy()
@@ -110,30 +116,23 @@ func (r *RolloutHandler) getRolloutAllClusters(rolloutStrategy RolloutStrategy, 
 		return &strategy, RolloutResult{}, err
 	}
 
-	// Get all clusters and perform progressive rollout
-	totalClusterGroups := r.pdTracker.ExistingClusterGroupsBesides()
-	totalClusters := totalClusterGroups.GetClusters().UnsortedList()
-	rolloutResult := progressivePerCluster(totalClusterGroups, len(totalClusters), failureTimeout, statusFunc)
+	allClusterGroups := r.pdTracker.ExistingClusterGroupsBesides()
+	allClusters := allClusterGroups.GetClusters().UnsortedList()
+
+	// Check for removed Clusters
+	currentClusterStatus, removedClusterStatus := r.getRemovedClusters(allClusterGroups, existingClusterStatus)
+	rolloutResult := progressivePerCluster(allClusterGroups, len(allClusters), failureTimeout, currentClusterStatus)
+	rolloutResult.ClustersRemoved = removedClusterStatus
 
 	return &strategy, rolloutResult, nil
 }
 
-func (r *RolloutHandler) getProgressiveClusters(rolloutStrategy RolloutStrategy, statusFunc ClusterRolloutStatusFunc) (*RolloutStrategy, RolloutResult, error) {
+func (r *RolloutHandler[T]) getProgressiveClusters(rolloutStrategy RolloutStrategy, existingClusterStatus []ClusterRolloutStatus) (*RolloutStrategy, RolloutResult, error) {
 	// Prepare the rollout strategy
 	strategy := RolloutStrategy{Type: Progressive}
 	strategy.Progressive = rolloutStrategy.Progressive.DeepCopy()
 	if strategy.Progressive == nil {
 		strategy.Progressive = &RolloutProgressive{}
-	}
-
-	// Upgrade mandatory decision groups first
-	groupKeys := decisionGroupsToGroupKeys(strategy.Progressive.MandatoryDecisionGroups.MandatoryDecisionGroups)
-	clusterGroups := r.pdTracker.ExistingClusterGroups(groupKeys...)
-
-	// Perform progressive rollout for mandatory decision groups
-	rolloutResult := progressivePerGroup(clusterGroups, maxTimeDuration, statusFunc)
-	if len(rolloutResult.ClustersToRollout) > 0 {
-		return &strategy, rolloutResult, nil
 	}
 
 	// Parse timeout for non-mandatory decision groups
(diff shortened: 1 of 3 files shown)


serngawy:
Adding
@dhaiducek

cluster/v1alpha1/helpers.go
```diff
@@ -200,28 +215,45 @@ func progressivePerCluster(clusterGroupsMap clusterv1beta1.ClusterGroupsMap, len
...
 		status.Status = newStatus.Status
 		status.TimeOutTime = newStatus.TimeOutTime
 
+		if status.Status == TimeOut {
+			timeoutClusters = append(timeoutClusters, status)
+			continue
+		}
 		if needToRollout {
-			rolloutClusters[cluster] = status
+			rolloutClusters = append(rolloutClusters, status)
```
haoqing0110: Need to check the rolloutClusters length I think, should not put all the needToRollout existing clusters into the result.

cluster/v1alpha1/helpers.go
```diff
@@ -236,32 +268,54 @@ func progressivePerCluster(clusterGroupsMap clusterv1beta1.ClusterGroupsMap, len
...
-func progressivePerGroup(clusterGroupsMap clusterv1beta1.ClusterGroupsMap, timeout time.Duration, statusFunc ClusterRolloutStatusFunc) RolloutResult {
-	rolloutClusters := map[string]ClusterRolloutStatus{}
-	timeoutClusters := map[string]ClusterRolloutStatus{}
+func progressivePerGroup(clusterGroupsMap clusterv1beta1.ClusterGroupsMap, timeout time.Duration, existingClusterStatus []ClusterRolloutStatus) RolloutResult {
+	rolloutClusters := []ClusterRolloutStatus{}
+	timeoutClusters := []ClusterRolloutStatus{}
+	existingClusters := make(map[string]bool)
 
-	clusterGroupKeys := clusterGroupsMap.GetOrderedGroupKeys()
+	for _, status := range existingClusterStatus {
```
haoqing0110, line 294: This function is to rollout by group, why go through the existingClusterStatus and put it to result? what will the returned rolloutClusters be like?

cluster/v1alpha1/helpers.go
serngawy: This iterate on the existing clusterRolloutStatus, Assuming it shouldn't never reach the length. okay will add it for safety.

cluster/v1alpha1/helpers.go
serngawy: it return the clusters that are still progressing, failing (before the timeout) and new clusters add to the group after the rollout starts

cluster/v1alpha1/helpers.go
haoqing0110: For addon, there's the possibility that 3 managedclusteraddons are created(existing) and status is "ToApply", if don't check the length, all of the 3 will be returned.

cluster/v1alpha1/helpers.go
haoqing0110: Still have some concerns about going through the existingClusterStatus first and then going through the rest of the clusters, for both progressivePerCluster and progressivePerGroup. 

For example, the env has 100 clusters with status Succeed, then rollout starts, what's the expected cluster status returned by existingClusterStatus? For addon existingClusterStatus will return ToApply when rollout starts, in this case, all the clusters will be returned by GetRolloutCluster(), which doesn't make sense.

cluster/v1alpha1/helpers.go
serngawy: I'm assuming the existingClusterStatus hold only clusterStatus; progressing, succeeded, failed and timeout cause the clusterStatus ToApply means the cluster workload does not exist yet. So as addon controller you should not add clusterStatus ToApply to the existingClusterStatus cause the addon (workload) does not exist. What do you think ? 
I will add check to ignore the ToApply clusterStatus at the  existingClusterStatus loop.

cluster/v1alpha1/helpers.go
serngawy: I understand your point, same as above let me know your thoughts.

cluster/v1alpha1/helpers.go
serngawy: I added the check for the ToApply status, so if the existingClusterStatus has all clusters in ToApply status, the RolloutResult will contain only the required maxConcurrency or the required decisionGroup. Added unitTest as well for this case.

cluster/v1alpha1/helpers.go
```diff
@@ -55,24 +54,30 @@ type ClusterRolloutStatus struct {
...
-	ClustersTimeOut map[string]ClusterRolloutStatus
+	// ClustersToRollout is a slice of ClusterRolloutStatus that will be rolled out.
+	ClustersToRollout []ClusterRolloutStatus
+	// ClustersTimeOut is a slice of ClusterRolloutStatus that are timeout.
+	ClustersTimeOut []ClusterRolloutStatus
+	// ClustersRemoved is a slice of ClusterRolloutStatus that are removed.
+	ClustersRemoved []ClusterRolloutStatus
 }
 
+// ClusterRolloutStatusFunc defines a function to return the rollout status for a managed cluster.
```
dhaiducek:
```suggestion
// ClusterRolloutStatusFunc defines a function to return the rollout status for a managed cluster for a given workload.
```

cluster/v1alpha1/helpers.go
haoqing0110: I think the ToApply means the desired status not applied yet, no matter it's a fresh install (no current workload) or upgrade case (current workload does not match the desired workload) .

cluster/v1alpha1/helpers.go
haoqing0110: Another gap after this change is that, in addon, one cma will have multiple placement, there's placement overlap. For example:
The config-v1 existingClusterStatus has cluster1 and config-v2 existingClusterStatus has cluster2, cluster3. For config-v1, cluster2 is not in existing cluster and it should also not be treated as added cluster and return in rollout result. 
```
  installStrategy:
    type: Placements
    placements:
    - name: placement1 // placement1 select cluster1 and cluster2
      namespace: default
      configs:
      - group: addon.open-cluster-management.io
        resource: addonhubconfigs
        name: config-v1
      rolloutStrategy:
        type: Progressive
    - name: placement2 // placement2 select cluster2 and cluster3
      namespace: default
      configs:
      - group: addon.open-cluster-management.io
        resource: addonhubconfigs
        name: config-v2
      rolloutStrategy:
        type: Progressive
```
I'm thinking should we just rollout the clusters returned by ClusterRolloutStatusFunc and put added/deleted cluster in another field.

cluster/v1alpha1/helpers.go
```diff
@@ -55,24 +54,30 @@ type ClusterRolloutStatus struct {
...
-	// ClustersToRollout is a map where the key is the cluster name and the value is the ClusterRolloutStatus.
-	ClustersToRollout map[string]ClusterRolloutStatus
-	// ClustersTimeOut is a map where the key is the cluster name and the value is the ClusterRolloutStatus.
-	ClustersTimeOut map[string]ClusterRolloutStatus
+	// ClustersToRollout is a slice of ClusterRolloutStatus that will be rolled out.
+	ClustersToRollout []ClusterRolloutStatus
+	// ClustersTimeOut is a slice of ClusterRolloutStatus that are timeout.
+	ClustersTimeOut []ClusterRolloutStatus
+	// ClustersRemoved is a slice of ClusterRolloutStatus that are removed.
+	ClustersRemoved []ClusterRolloutStatus
```
dhaiducek, line 62: I'm assuming this is an array now so that the mapping to each cluster's name is clearer to the end user? I kind of liked the `map[string]ClusterRolloutStatus` for uniqueness and simplicity.

cluster/v1alpha1/helpers.go
```diff
@@ -110,90 +116,166 @@ func (r *RolloutHandler) getRolloutAllClusters(rolloutStrategy RolloutStrategy,
...
+		if len(rolloutResult.ClustersToRollout) > 0 || len(rolloutResult.ClustersTimeOut) > 0 {
+			rolloutResult.ClustersRemoved = removedClusterStatus
+			return &strategy, rolloutResult, nil
+		}
+	}
+
+	// Calculate the length for progressive rollOut
+	// If the MaxConcurrency not defined, total clusters length is considered as maxConcurrency.
+	clusterGroups = r.pdTracker.ExistingClusterGroupsBesides(groupKeys...)
+	length, err := calculateLength(strategy.Progressive.MaxConcurrency, len(clusterGroups.GetClusters()))
```
dhaiducek: Aside: This is outside of this PR's intent, but could we consider renaming `calculateLength()` to something more descriptive like `calculateRolloutSize()`?

cluster/v1alpha1/helpers.go
```diff
@@ -110,90 +116,166 @@ func (r *RolloutHandler) getRolloutAllClusters(rolloutStrategy RolloutStrategy,
...
+	removedClusterStatus := []ClusterRolloutStatus{}
 
-	if length == 0 {
+	clusters := clusterGroupsMap.GetClusters().UnsortedList()
+	for _, clusterStatus := range existingClusterStatus {
+		exist := false
+		for _, cluster := range clusters {
+			if clusterStatus.ClusterName == cluster {
+				exist = true
+				currentClusterStatus = append(currentClusterStatus, clusterStatus)
```
dhaiducek, line 228: Should there be a `break` here so that the loop doesn't continue?

cluster/v1alpha1/helpers.go
```diff
@@ -110,90 +116,166 @@ func (r *RolloutHandler) getRolloutAllClusters(rolloutStrategy RolloutStrategy,
...
+	timeoutClusters := []ClusterRolloutStatus{}
+	existingClusters := make(map[string]bool)
+
+	for _, status := range existingClusterStatus {
+		if status.ClusterName == "" {
+			continue
+		}
+
+		existingClusters[status.ClusterName] = true
+		if status.Status == Succeeded || status.Status == TimeOut {
```
dhaiducek: Optional: This could be a `switch` statement.

cluster/v1alpha1/helpers.go
```diff
@@ -110,90 +116,166 @@ func (r *RolloutHandler) getRolloutAllClusters(rolloutStrategy RolloutStrategy,
...
+				ClustersToRollout: rolloutClusters,
+				ClustersTimeOut:   timeoutClusters,
+			}
+		}
+	}
+
+	if len(rolloutClusters) >= length {
 		return RolloutResult{
 			ClustersToRollout: rolloutClusters,
 			ClustersTimeOut:   timeoutClusters,
```
dhaiducek: Isn't this check redundant since it's already in the loop above?

dhaiducek (commented):
Thanks for the update! I've left some minor comments.

(Regarding the current discussion about removed clusters: In general, I like having the removed clusters available to the controller so that they can be handled right away, but I'm still processing the existing discussion about it.)

cluster/v1alpha1/helpers.go
serngawy: I understand its easier to iterate with a map but with the changes that happen we don't really need the clusterName to be in a map. Plus its better to move the clusterName to the ClusterRolloutStatus struct to be identifiable object by itself.

cluster/v1alpha1/helpers.go
serngawy: done

cluster/v1alpha1/helpers.go
serngawy: done

cluster/v1alpha1/helpers.go
serngawy: yes, done

cluster/v1alpha1/helpers.go
serngawy: okay, done.

cluster/v1alpha1/helpers.go
serngawy: yes, removed.

cluster/v1alpha1/helpers.go
serngawy: The removed (deleted) clusters are already in another field [ClustersRemoved](https://github.com/open-cluster-management-io/api/pull/276/files#diff-5e4fb09ae0fb45e5fd5a1d44b77b23aa09be3e9bfaf09191a46cb4eb38a058caR62) . 
Regards the added clusters, not sure if I understand the example above correctly. Config-v1 is associated with placement1 and config-v2 is associated with placement2; those are 2 different configs with 2 different placements.  However; as cluster-2 is associated with config-v1 and config-v2, cluster-2 must apply config-v2 as it has higher version. Is that correct ?  
If yes, I guess this is more addon logic that is need to be handled inside the addon controller. what do you think ?

haoqing0110: Addon has returned a SKIP status for cluster2 when rollout config-v1 to indicate not to rollout it. Not sure if PR can handle the case when cluster2 is not in existingClusterStatus and user want to SKIP rollout it.

cluster/v1alpha1/helpers.go
```diff
@@ -110,119 +116,182 @@ func (r *RolloutHandler) getRolloutAllClusters(rolloutStrategy RolloutStrategy,
...
+	}
+	return currentClusterStatus, removedClusterStatus
+}
+
+func progressivePerCluster(clusterGroupsMap clusterv1beta1.ClusterGroupsMap, length int, timeout time.Duration, existingClusterStatus []ClusterRolloutStatus) RolloutResult {
+	rolloutClusters := []ClusterRolloutStatus{}
+	timeoutClusters := []ClusterRolloutStatus{}
+	existingClusters := make(map[string]bool)
+
+	for _, status := range existingClusterStatus {
```
haoqing0110, line 244: Should here sort the existingClusterStatus by name by alphabetical order to ensure every time returns the same result?

haoqing0110: Could help explain a bit more for the existingClusterStatus, for example:
1. For a fresh install, should existingClusterStatus include any clusters? 
2. For a rollout case, if 100 clusters have old workload with status success, should they be put into existingClusterStatus with status Succeed or should they not be put into existingClusterStatus?

cluster/v1alpha1/helpers.go
```diff
@@ -55,24 +54,30 @@ type ClusterRolloutStatus struct {
...
+	// ClustersRemoved is a slice of ClusterRolloutStatus that are removed.
+	ClustersRemoved []ClusterRolloutStatus
 }
 
+// ClusterRolloutStatusFunc defines a function to return the rollout status for a given workload.
+type ClusterRolloutStatusFunc[T any] func(clusterName string, workload T) (ClusterRolloutStatus, error)
+
 // +k8s:deepcopy-gen=false
-type RolloutHandler struct {
+type RolloutHandler[T any] struct {
```
qiujian16, line 71: should it be runtime.Object rather than any?

cluster/v1alpha1/helpers.go
```diff
@@ -83,20 +88,21 @@ func NewRolloutHandler(pdTracker *clusterv1beta1.PlacementDecisionClustersTracke
 //
 // ClustersTimeOut: If the cluster status is Progressing or Failed, and the status lasts longer than timeout defined in strategy,
 // will list them RolloutResult.ClustersTimeOut with status TimeOut.
-func (r *RolloutHandler) GetRolloutCluster(rolloutStrategy RolloutStrategy, statusFunc ClusterRolloutStatusFunc) (*RolloutStrategy, RolloutResult, error) {
+// func (r *RolloutHandler) GetRolloutCluster(rolloutStrategy RolloutStrategy, statusFunc ClusterRolloutStatusFunc) (*RolloutStrategy, RolloutResult, error) {
```
qiujian16: remove this line

cluster/v1alpha1/helpers.go
```diff
@@ -110,119 +116,182 @@ func (r *RolloutHandler) getRolloutAllClusters(rolloutStrategy RolloutStrategy,
...
-			ClustersTimeOut:   timeoutClusters,
+		if !exist {
+			removedClusterStatus = append(removedClusterStatus, clusterStatus)
+		}
+	}
+	return currentClusterStatus, removedClusterStatus
+}
+
+func progressivePerCluster(clusterGroupsMap clusterv1beta1.ClusterGroupsMap, length int, timeout time.Duration, existingClusterStatus []ClusterRolloutStatus) RolloutResult {
+	rolloutClusters := []ClusterRolloutStatus{}
```
qiujian16: use var rolloutClusters, timeoutClusters []ClusterRolloutStatus

cluster/v1alpha1/helpers.go
```diff
@@ -236,32 +305,57 @@ func progressivePerCluster(clusterGroupsMap clusterv1beta1.ClusterGroupsMap, len
...
+			if status.Status == TimeOut {
+				timeoutClusters = append(timeoutClusters, status)
+				continue
+			}
 
+			if needToRollout {
+				rolloutClusters = append(rolloutClusters, status)
+			}
+		}
+	}
```
haoqing0110, line 337: Should here return after going through existingClusterStatus?

cluster/v1alpha1/helpers.go
```diff
@@ -236,32 +305,57 @@ func progressivePerCluster(clusterGroupsMap clusterv1beta1.ClusterGroupsMap, len
...
+			continue
+		}
+
+		existingClusters[status.ClusterName] = true
+		switch status.Status {
+		case ToApply:
+			// Set as false to consider the cluster in the decisionGroups iteration.
+			existingClusters[status.ClusterName] = false
+		case Failed, Progressing:
+			newStatus, needToRollout := determineRolloutStatusAndContinue(status, timeout)
```
haoqing0110: Can here combine the switch case and determineRolloutStatusAndContinue's switch case into one function?

cluster/v1alpha1/helpers.go
serngawy: yes, I added the check for the ToApply state as well.

cluster/v1alpha1/helpers.go
serngawy: so the idea for existingClusterStatus; if the the existing cluster has Skip status it will not be considered in the rolloutResult->ClustersToRollout. The addon set the cluster2 ClusterRolloutStatus to skip (for configv1 and placement1) then the rolloutResult->ClustersToRollout will not have it. I added comments [here](https://github.com/open-cluster-management-io/api/pull/276/files#diff-5e4fb09ae0fb45e5fd5a1d44b77b23aa09be3e9bfaf09191a46cb4eb38a058caR88) as well

haoqing0110: Yes, addon might need to put SKIP cluster in existingClusterStatus so that won't put it in ClustersToRollout. Sounds good.

cluster/v1alpha1/helpers.go
serngawy: runtime.Object will required casting for the used Type and We want the RolloutHandler to be initiated with a certain type (eg; manifestwork, policy, addon, ...etc) not a runtime.object

cluster/v1alpha1/helpers.go
serngawy: okay, done

cluster/v1alpha1/helpers.go
serngawy: okay, done.

cluster/v1alpha1/helpers.go
qiujian16: but any is a more loose constraint, isn't it?

serngawy: Well, yes "any" is kind of loose, but we are safe. The main idea behind make a generic Type "any" is to avoid doing casting (or force certain type) at the clusterRolloutStatusFunc Function implementation similar to the unit test example [here](https://github.com/serngawy/api/blob/rolloutLib/cluster/v1alpha1/helpers_test.go#L47) AND the helper lib is safe we don't call the clusterRolloutStatusFunc any where, its just a func definition that will be used to create the existing ClustersRollOutStatus at the consumer API side. What do you think ? Is that make sense ?

xuezhaojun: Run `make update` will fail on deepcopy functions generating step:
```
Generating deepcopy funcs
W1017 14:10:32.195161   62703 parse.go:863] Making unsupported type entry "T" for: &types.TypeParam{check:(*types.Checker)(nil), id:0x1, obj:(*types.TypeName)(0xc009651900), index:0, bound:(*types.Interface)(0xc00009a960)}
F1017 14:10:32.221298   62703 deepcopy.go:890] Hit an unsupported type open-cluster-management.io/api/cluster/v1alpha1.ClusterRolloutStatusFunc[T] for open-cluster-management.io/api/cluster/v1alpha1.ClusterRolloutStatusFunc[T], from open-cluster-management.io/api/cluster/v1alpha1.RolloutHandler[T]
```

Seems deepcopy lib doesn't support generic type yet. https://github.com/kubernetes/gengo/issues/225

Either we revert the normal struct or we disable package-level generation and explicitly add the comment to each struct we need do deepcopy function generating: https://github.com/open-cluster-management-io/api/pull/288

cc @haoqing0110

dhaiducek: @xuezhaojun I noticed that also and ran `controller-gen` to generate the functions instead--would that work instead here (at least until `gengo` supports generics...or even migrate the whole thing to use `controller-gen`)?
```shell
./_output/tools/bin/controller-gen object:headerFile="hack/empty.txt" paths="./cluster/v1alpha1"
```

xuezhaojun: Hi, @dhaiducek , I test `controller-gen` with latest api main branch, it fails and generates unexpected code for generic type:
```
./_output/tools/bin/controller-gen object:headerFile="hack/empty.txt" paths="./cluster/v1alpha1"
open-cluster-management.io/api/cluster/v1alpha1:-: invalid type: func(clusterName string, workload T) (open-cluster-management.io/api/cluster/v1alpha1.ClusterRolloutStatus, error)
Error: not all generators ran successfully
run `controller-gen object:headerFile=hack/empty.txt paths=./cluster/v1alpha1 -w` to see all available markers, or `controller-gen object:headerFile=hack/empty.txt paths=./cluster/v1alpha1 -h` for usage
```

xuezhaojun: As I mentioned above, we can also change to the explicit way to claim which struct needs gen deepcopy funcs: https://github.com/open-cluster-management-io/api/pull/288

dhaiducek: Actually it seems the `gengo` tooling may not be respecting the `deepcopy-gen=false` tag? In any case, I don't have a great deal of familiarity with deepcopy, but here's a commit that might work as an alternative to selectively generating the APIs:
- https://github.com/dhaiducek/api/commit/3734d0e19b78b8feb430640486103e520a15a927

serngawy: Thanks @xuezhaojun I'm able to reproduce it as well. having the same issue as @dhaiducek sounds like deepCopy generator does not respect the `deepcopy-gen=false` tag ! 
Based on the deepCopy code doc [here](https://github.com/kubernetes/code-generator/blob/master/cmd/deepcopy-gen/main.go#L39) setting the deepcopy-gen=false should work.

xuezhaojun: Thanks for providing the alternative! I will follow with it.

PR #279, opened 2023-09-12
🐛 support update&updateStatus for manifestwork cloudevents client.

morvencao:
## Summary

This PR introduces the addition of 'Update' and 'UpdateStatus' functionalities for the manifestwork cloudevents client. These methods are necessary as the work-client implementation will use them, but currently the manifestwork cloudevents client only includes 'Get' and 'Patch'.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -5,11 +5,13 @@ on:
     branches:
       - main
       - release-*
+      - br_support_update_for_manifestwork_cloudevents_client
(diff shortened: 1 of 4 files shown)


skeeey:
@morvencao the newest work-agent in ocm only use patch way to update mw or it's status, this is why we only implement the patch way for the agent mw client :)

morvencao:
closing this as the new work-agent only needs `patch`.

PR #280, opened 2023-09-15
:sparkles: retry connecting the cloudevents client

skeeey:
## Summary
Retry connecting the cloudevents client when the client has network error

### DIFF
--- a/cloudevents/generic/baseclient.go
+++ b/cloudevents/generic/baseclient.go
@@ -1,0 +1,209 @@
+package generic
+
+import (
+	"context"
+	"fmt"
+	"sync"
+	"time"
+
+	cloudevents "github.com/cloudevents/sdk-go/v2"
+
+	"k8s.io/apimachinery/pkg/util/runtime"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/util/flowcontrol"
+	"k8s.io/klog/v2"
+	"k8s.io/utils/clock"
+
+	"open-cluster-management.io/api/cloudevents/generic/options"
+)
+
+const (
+	restartReceiverSignal = iota
+	stopReceiverSignal
+)
+
+type receiveFn func(ctx context.Context, evt cloudevents.Event)
+
+type baseClient struct {
+	sync.RWMutex
+	cloudEventsOptions     options.CloudEventsOptions
+	cloudEventsClient      cloudevents.Client
+	cloudEventsRateLimiter flowcontrol.RateLimiter
+	receiverChan           chan int
+}
+
+func (c *baseClient) connect(ctx context.Context) error {
+	var err error
+	c.cloudEventsClient, err = c.cloudEventsOptions.Client(ctx)
+	if err != nil {
+		return err
+	}
+
+	// start a go routine to handle cloudevents client connection errors
+	go func() {
+		var err error
+
+		// the reconnect backoff will stop at [1,5) min interval. If we don't backoff for 10min, we reset the backoff.
+		connBackoffManager := wait.NewExponentialBackoffManager(5*time.Second, 1*time.Minute, 10*time.Minute, 5.0, 1.0, &clock.RealClock{})
+		cloudEventsClient := c.cloudEventsClient
+
+		for {
+			if cloudEventsClient == nil {
+				klog.V(4).Infof("reconnecting the cloudevents client")
+				cloudEventsClient, err = c.cloudEventsOptions.Client(ctx)
+				// TODO enhance the cloudevents SKD to avoid wrapping the error type to distinguish the net connection
+				// errors
+				if err != nil {
+					// failed to reconnect, try agin
+					runtime.HandleError(fmt.Errorf("the cloudevents client reconnect failed, %v", err))
+					<-connBackoffManager.Backoff().C()
+					continue
+				}
+
+				// the cloudevents network connection is back, refresh the current cloudevents client and send the
+				// receiver restart signal
+				klog.V(4).Infof("the cloudevents client is reconnected")
+				c.resetClient(cloudEventsClient)
+				c.sendReceiverSignal(restartReceiverSignal)
+			}
+
+			select {
+			case <-ctx.Done():
+				return
+			case err, ok := <-c.cloudEventsOptions.ErrorChan():
+				if !ok {
+					// error channel is closed, do nothing
+					return
+				}
+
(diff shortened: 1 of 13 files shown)


cloudevents/generic/baseclient.go
```diff
@@ -0,0 +1,140 @@
...
+
+			if retrying {
+				klog.V(4).Infof("reconnecting the cloudevents client")
+				cloudEventsClient, err := c.cloudEventsOptions.Client(ctx)
+				if err != nil {
+					// failed to reconnect, retry agin
+					// TODO enhance the cloudevents SKD to avoid wrapping the error type, then we can only handle the
+					// net connection errors here
+					klog.Errorf("failed to reconnect, %v", err)
+					<-connBackoffManager.Backoff().C()
```
qiujian16: if it goes into this branch, will it hang on the select in the next loop?

cloudevents/generic/baseclient.go
```diff
@@ -0,0 +1,140 @@
...
+					continue
+				}
+
+				klog.Errorf("the cloudevents client is disconnected, %v", err)
+
+				// the cloudevents network connection is closed, stop the receiver and start retry
+				if receiverCancel != nil {
+					receiverCancel()
+				}
+				retrying = true
```
qiujian16: can we just set c.cloudEventsClient = nil here?

skeeey: the cloudEventsClient is also used by send method, if we set it to `nil` here, the send may meet the nil pointer, so I did not set it to nil, what's your suggestion?

cloudevents/generic/baseclient.go
skeeey: the select in the next loop need to wait the backoff, the backoff is a blocking action

cloudevents/generic/baseclient.go
qiujian16: even though, the send will fail right? I think there might need a rwlock in send otherwise there is always a change that the two threads race.

qiujian16: I mean after backoff it will go to select in the next loop while client is not instantiated, and it will not have a chance to receive message from c.cloudEventsErrChan

skeeey: yes, the send will be fail, we maybe need a lock here for the client

the c.cloudEventsErrChan will be only send by a running client, so if the client is retrying, the select will not be blocked the next retry

cloudevents/generic/baseclient.go
```diff
@@ -0,0 +1,157 @@
...
+		var err error
+
+		receiving := false
+		cloudEventsClient := c.cloudEventsClient
+		// the reconnect backoff will stop at [1,5) min interval. If we don't backoff for 10min, we reset the backoff.
+		connBackoffManager := wait.NewExponentialBackoffManager(5*time.Second, 1*time.Minute, 10*time.Minute, 5.0, 1.0, &clock.RealClock{})
+
+		for {
+			if cloudEventsClient == nil {
+				klog.V(4).Infof("reconnecting the cloudevents client")
```
qiujian16: one thing missing is: what if the client just send but not subscribe? Do we alway assume that the client will send and subscribe together?

skeeey: hm... I'm not sure, maybe we cannot assume this, in our case, it seems that for our case the send and sub are always together, but as a lib, we don't stop to separate the send and subscribe, user can send in one process, and sub in other process, maybe we need to have one retry mechanism for the client

skeeey: separate the reconnect part from the subscription, PTAL

cloudevents/generic/baseclient.go
```diff
@@ -0,0 +1,216 @@
...
+)
+
+const (
+	restartReceiverSignal = iota
+	stopReceiverSignal
+)
+
+type receiveFn func(ctx context.Context, evt cloudevents.Event)
+
+type baseClient struct {
```
qiujian16, line 27: should we have a separated Connect(ctx context.Context) in baseClient? so we do not start goroutine during new?

skeeey: use connect method instead of new method

qiujian16 (commented):
/approve

qiujian16:
Maybe @yanmxa or @morvencao could also take a look

cloudevents/generic/baseclient.go
```diff
@@ -0,0 +1,209 @@
...
+					return
+				}
+
+				switch signal {
+				case restartReceiverSignal:
+					klog.V(4).Infof("restart the cloudevents receiver")
+					// make sure the current client is the newest
+					c.RLock()
+					cloudEventsClient = c.cloudEventsClient
+					c.RUnlock()
```
morvencao, line 179: do we still need to lock it here? since it is locked at the beginning of the subscribe function

skeeey: yes, we need it, the beginning lock is for the subscribe function (it ensure there is only one subscription go routine running), this lock is in a independent go routine (the subscribe lock will be released after this go routine starts)

morvencao: emm, right, thanks for the explanation.

morvencao:
looks good to me, leave this to @yanmxa

cloudevents/generic/baseclient.go
```diff
@@ -0,0 +1,209 @@
...
+
+		// the reconnect backoff will stop at [1,5) min interval. If we don't backoff for 10min, we reset the backoff.
+		connBackoffManager := wait.NewExponentialBackoffManager(5*time.Second, 1*time.Minute, 10*time.Minute, 5.0, 1.0, &clock.RealClock{})
+		cloudEventsClient := c.cloudEventsClient
+
+		for {
+			if cloudEventsClient == nil {
+				klog.V(4).Infof("reconnecting the cloudevents client")
+				cloudEventsClient, err = c.cloudEventsOptions.Client(ctx)
+				// TODO enhance the cloudevents SKD to avoid wrapping the error type to distinguish the net connection
```
yanmxa, line 54: The reference: https://github.com/eclipse/paho.golang/issues/26 

I will try to add the https://github.com/eclipse/paho.golang/blob/master/autopaho/auto.go code to the mqtt protocol of cloudevents

yanmxa:
Reference: https://github.com/cloudevents/sdk-go/issues/940

/LGTM

PR #281, opened 2023-09-15
:sparkles: Replace `Timeout` with `RolloutConfig`

dhaiducek:
## Summary

Implement `RolloutConfig`, which provides more fine-tuned configuration for users than the current `Timeout`.

## Related issue(s)

- https://github.com/open-cluster-management-io/enhancements/pull/100


qiujian16:
@dhaiducek since we are about to release 0.12.0, would like to know the progress of this PR. Is it ready for review, do you plan to include it in 0.12.0 or it can be moved to next release?

dhaiducek:
@qiujian16 It was waiting on #276, but now that it's merged I'll continue work here. I'll work on it this week, but I think it's also fine if `v0.12.0` is released without these library updates so that this PR doesn't block the release.

dhaiducek:
Status update: I wrestled with the CI a bit because I'm on an M1 MacBook and some of the binaries aren't actually built for arm64 (those changes are in this PR). I've got current code working, migrating `Timeout` --> `ProgressDeadline`, but need to add logic (and tests) for the other fields that are in the new struct alongside it.

dhaiducek:
/cc @serngawy @haoqing0110 

Sorry for the long delay here--between wrestling with the CI and getting pulled away for other tasks, it's taken longer to update than I thought. There might be some more tests required here, but I think the logic is largely (if not entirely) there.

Some notes:
- I added logic to pass `Timeout` over to `ProgressDeadline` and mark `Timeout` as deprecated. I debated having `MaxFailures` set to `100%` so the behavior wouldn't change there, but I think it better to leave it and make the migration obvious.
- I made a decision to handle `MaxFailures` over the total number of clusters for `Progressive` and per group for `ProgressivePerGroup`, partially because `Progressive` clusters are sorted alphabetically, so having a cluster injected at the beginning or middle could complicate things.

serngawy:
Sounds good to me in general, I added more unit test to cover Rollout ProgressivePerGroup  case.

dhaiducek:
After discussion with @serngawy I've updated the logic so that `MandatoryDecisionGroups` tolerate no failures for any progressive rollout type.

cluster/v1alpha1/helpers.go
```diff
@@ -181,13 +205,21 @@ func (r *RolloutHandler[T]) getProgressivePerGroupClusters(rolloutStrategy Rollo
...
 
 	// Parse timeout for non-mandatory decision groups
-	failureTimeout, err := parseTimeout(strategy.ProgressivePerGroup.Timeout.Timeout)
+	failureTimeout, err := parseTimeout(strategy.ProgressivePerGroup.ProgressDeadline)
 	if err != nil {
 		return &strategy, RolloutResult{}, err
 	}
 
+	// Check format of MaxFailures--this value will be re-parsed and used in progressivePerGroup()
+	_, err = calculateRolloutSize(maxFailures, 100, 0)
```
haoqing0110: Can understand the intention here. Is that possible to split the calculateRolloutSize into a parse function and a calculate function, so that don't need to hardcode a 100 which looks strange.

haoqing0110:
LGTM, seems the `ProgressDeadline` is just a rename of `Timeout`, I'm fine to leave the `Timeout` and mark it as deprecated.

cluster/v1alpha1/helpers.go
dhaiducek: Ah, good point--I'll update it.

qiujian16:
/approve

Since manifestworkReplicaSet and ClusterManagerAddon install strategy  is in alpha stage. Let's remove timeout in the next release (ie 0.14.0)

dhaiducek:
> Since manifestworkReplicaSet and ClusterManagerAddon install strategy is in alpha stage. Let's remove timeout in the next release (ie 0.14.0)

Sounds good, thanks @qiujian16!

@serngawy @haoqing0110 I'd appreciate if you'd do another review here when you have a moment! 😄

serngawy:
> > Since manifestworkReplicaSet and ClusterManagerAddon install strategy is in alpha stage. Let's remove timeout in the next release (ie 0.14.0)
> 
> Sounds good, thanks @qiujian16!
> 
> @serngawy @haoqing0110 I'd appreciate if you'd do another review here when you have a moment! 😄

I gave it another review, sounds good to me @dhaiducek appreciate the good work :) 

ops, didn't consider my approve will merge it @haoqing0110

serngawy (approved):

PR #282, opened 2023-09-18
:seedling: combine the sender and receiver to one client

skeeey:
## Summary
combine the sender and receiver to one client


qiujian16 (commented):
/approve
@yanmxa pls take a look

PR #286, opened 2023-09-25
:seedling: add clusterDeleting condition and  cleanup-priority annotation

zhiweiyin318:
## Related issue(s)

Fixes #

### DIFF
--- a/cluster/v1/types.go
+++ b/cluster/v1/types.go
@@ -257,3 +257,26 @@ const (
 	// relating to the ManagedCluster is removed when the ManagedCluster is deleted.
 	ManagedClusterFinalizer = "cluster.open-cluster-management.io/api-resource-cleanup"
 )
+
+const (
+	// ManagedClusterConditionDeleting is a condition which means the cluster is in deletion process.
+	ManagedClusterConditionDeleting string = "Deleting"
+
+	// ConditionDeletingReasonResourceRemaining is a reason for the condition ManagedClusterIsDeleting, which means
+	// there are resources are remaining during deletion process.
+	ConditionDeletingReasonResourceRemaining string = "ResourceRemaining"
+
+	// ConditionDeletingReasonResourceError is a reason for the condition ManagedClusterIsDeleting, which means
+	// meet errors during the deletion process.
+	ConditionDeletingReasonResourceError string = "DeletingError"
+
+	// CleanupPriorityAnnotationKey is a annotation for the resources deployed in cluster ns which are waiting to
+	// be cleaned up after cluster is deleted.
+	// The value is an integer value [0,100], The larger the value, the later the order of deletion.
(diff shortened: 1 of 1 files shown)


utils/constants.go
```diff
@@ -0,0 +1,13 @@
...
+const (
+	// LabelDeletionPriority is a label for the resources deployed in cluster ns which are waiting to be cleaned up
+	// after cluster is deleted.
+	// The value is an integer value, The larger the value, the later the order of deletion.
+	// The deletion order is :
+	// 1. delete resources without this label firstly.
+	// 2. delete resources with invalid value of this label.
+	// 3. delete resources following the priority value. For example, there are 2 manifestWorks, one value is set 100
+	// and another is set 10, the manifestWorks with 10 will be deleted before the one with 100.
+	LabelDeletionPriority string = "open-cluster-management.io/deletion-priority"
```
qiujian16: seems to me it should be a field rather than label

zhiweiyin318: changed to annotation with the name cleanup-priority.

cluster/v1/types.go
```diff
@@ -257,3 +257,16 @@ const (
 	// relating to the ManagedCluster is removed when the ManagedCluster is deleted.
 	ManagedClusterFinalizer = "cluster.open-cluster-management.io/api-resource-cleanup"
 )
+
+const (
+	// ManagedClusterConditionDeleting is a condition which means the cluster is in deletion process.
+	ManagedClusterConditionDeleting string = "ManagedClusterIsDeleting"
```
qiujian16: Deleting is good enough, it is ManagedCluster resource so it is not needed to specify it in the type

utils/constants.go
```diff
@@ -0,0 +1,13 @@
...
+const (
+	// CleanupPriorityAnnotationKey is a annotation for the resources deployed in cluster ns which are waiting to
+	// be cleaned up after cluster is deleted.
+	// The value is an integer value [0,100], The larger the value, the later the order of deletion.
+	// The deletion order is :
+	// 1. delete resources without this annotation firstly.
+	// 2. delete resources with invalid value of this annotation (!= [0,100]).
+	// 3. delete resources following the priority value. For example, there are 2 manifestWorks, one value is set 100
+	// and another is set 10, the manifestWorks with 10 will be deleted before the one with 100.
+	CleanupPriorityAnnotationKey string = "open-cluster-management.io/cleanup-priority"
```
qiujian16: this should also be in the types, but not here.

cluster/v1/types.go
zhiweiyin318: done

utils/constants.go
zhiweiyin318: moved to types of clusters

PR #291, opened 2023-10-18
:bug: Use controller-runtime for deepcopy generation for cluster:v1alpha1

dhaiducek:
## Summary

GenGo doesn't seem to respect the `+k8s:deepcopy-gen:false` tag, so this is an alternative to #288, where `controller-gen` is used for `cluster:v1alpha1` deepcopy generation rather than selective deepcopy generation.

## Related issue(s)

Followup to (and reverts):
- #288


xuezhaojun:
@dhaiducek The bot may squash commits after merging the PR, and also, we may be able to replace deepcopy-gen with controller-gen entirely so that there is no exception in the makefile flow.

dhaiducek:
> @dhaiducek The bot may squash commits after merging the PR, and also, we may be able to replace deepcopy-gen with controller-gen entirely so that there is no exception in the makefile flow.

Good point--I'll create a separate PR for the other updates.

dhaiducek:
I've split out the architecture updates into a new PR:
- https://github.com/open-cluster-management-io/api/pull/292

dhaiducek:
> we may be able to replace deepcopy-gen with controller-gen entirely so that there is no exception in the makefile flow.

I'm working on implementing a `controller-gen` generation, but it also has an issue where it's generating everything without errors but then this generic `sets.Set` isn't getting properly instantiated with `string` as it's defined for some reason, so my code editor is complaining:
https://github.com/open-cluster-management-io/api/blob/d7736a1c8ebb36510ff85da914a17293fe3a504a/cluster/v1beta1/helpers.go#L217

dhaiducek:
For now, this PR might be the easiest path forward given `controller-gen` it seems might also need a different exception for that edge case? I've spent quite a bit of time on this without a clear resolution, so I'm going to pause on this work for now, but this PR it seems would work as-is.

Makefile
```diff
@@ -54,6 +54,7 @@ verify: check-env verify-scripts verify-codegen-crds verify-gocilint
 
 update-scripts:
 	hack/update-deepcopy.sh
+	$(CONTROLLER_GEN) object:headerFile="hack/empty.txt" paths="./cluster/v1alpha1"
```
qiujian16, line 60: we should add a comment here to indicate why using controller-gen

hack/update-deepcopy.sh
```diff
@@ -10,6 +10,6 @@ verify="${VERIFY:-}"
 GOFLAGS="" bash ${CODEGEN_PKG}/generate-groups.sh "deepcopy" \
   open-cluster-management.io/api/generated \
   open-cluster-management.io/api \
-  "cluster:v1 cluster:v1alpha1 cluster:v1beta1 cluster:v1beta2 work:v1alpha1 work:v1 operator:v1 addon:v1alpha1" \
+  "cluster:v1 cluster:v1beta1 cluster:v1beta2 work:v1alpha1 work:v1 operator:v1 addon:v1alpha1" \
```
qiujian16, line 16: ditto a comment, so we do not forget this tweak.

qiujian16 (commented):
@xuezhaojun  will this impact the `make verify`?

xuezhaojun:
> @xuezhaojun will this impact the `make verify`?

No, from the code, `verify` and `update` are using the same update-script under the hood. 

https://github.com/open-cluster-management-io/api/blob/d7736a1c8ebb36510ff85da914a17293fe3a504a/hack/verify-deepcopy.sh#L6

qiujian16 (commented):
/approve
/lgtm

PR #296, opened 2023-11-15
:sparkles: Add a condition of clock sync.

xuezhaojun:
## Summary

Add a condition to represent whether the clock of hub and agents are synced.

## Related issue(s)

Fixes #

### DIFF
--- a/cluster/v1/types.go
+++ b/cluster/v1/types.go
@@ -200,6 +200,8 @@ const (
 	// cluster is available, the kube-apiserver is healthy and the Klusterlet agent is
 	// running with the minimum deployment on this managed cluster
 	ManagedClusterConditionAvailable string = "ManagedClusterConditionAvailable"
+	// ManagedClusterConditionClockSynced means the clock between the hub and the agent is synced.
+	ManagedClusterConditionClockSynced string = "ManagedClusterConditionClockSynced"
 )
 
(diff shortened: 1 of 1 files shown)


cluster/v1/types.go
```diff
@@ -200,6 +200,8 @@ const (
 	// cluster is available, the kube-apiserver is healthy and the Klusterlet agent is
 	// running with the minimum deployment on this managed cluster
 	ManagedClusterConditionAvailable string = "ManagedClusterConditionAvailable"
+	// ManagedClusterConditionClockSynced means the clock between the hub and the agent is synced.
+	ManagedClusterConditionClockSynced string = "ManagedClusterConditionClockSynced"
```
qiujian16, line 204: could it be a reason under available condition?

xuezhaojun: The concern here is clock out-of-sync doesn't equal to Available = Unknown.

The trigger point of ClockSynced = False is that hub is `1` leaseDuration faster than the agent(the default gracePeriod is `5` * leaseDuration). At that point, the managed cluster is still available but we already want to get users attention and want to warn they to check on NTP status in their environment.

Making OutOfSync a reason of UnKnown is also a choice, but from user experience perspective, kind of feel there should be an independent condition to warn the user **before** things really go wrong and can't function.

PR #297, opened 2023-11-23
:sparkles: Add flag to allow use best effort resource for klusterlet

haowells:
## Summary
see PR in https://github.com/open-cluster-management-io/ocm/pull/317

## Related issue(s)

Fixes #


qiujian16:
would need to run `make update` to update deepcopy and client code

operator/v1/types_klusterlet.go
```diff
@@ -86,6 +86,10 @@ type KlusterletSpec struct {
 	// registration-agent and work-agent will use it to communicate with hub api server.
 	// +optional
 	HubApiServerHostAlias *HubApiServerHostAlias `json:"hubApiServerHostAlias,omitempty"`
+
+	// BestEffortResource allow use the best effort resource for operator/registration/work deployment
+	// +optional
+	BestEffortResource bool `json:"bestEffortResource,omitempty"`
```
qiujian16: It is better no to use boolean in API. You might consider a resources struct with a type in it. We can just set type to "BestEffort" and "Default" today, and add other option to be able to specify resource in the future.

haowells: @qiujian16  thanks. address your comments and PTAL again.

operator/v1/types_klusterlet.go
```diff
@@ -86,6 +86,22 @@ type KlusterletSpec struct {
...
+type ResourceQosClass string
+
+const (
+	ResourceQosClassDefault    ResourceQosClass = "Default"
+	ResourceQosClassBestEffort ResourceQosClass = "BestEffort"
+)
+
+// ResourceType allow user override the default pod QoS classes
+type ResourceType struct {
+	Type ResourceQosClass `json:"type"`
```
qiujian16, line 108: // +kubebuilder:validation:Enum=Default;BestEffort
  // +kubebuilder:default:=Default

Should add some valildation

operator/v1/types_klusterlet.go
```diff
@@ -86,6 +86,22 @@ type KlusterletSpec struct {
 	// registration-agent and work-agent will use it to communicate with hub api server.
 	// +optional
 	HubApiServerHostAlias *HubApiServerHostAlias `json:"hubApiServerHostAlias,omitempty"`
+
+	// ResourceType specify QoS classes of klusterlet deployment
+	// +optional
+	ResourceType ResourceType `json:"resourceType,omitempty"`
```
qiujian16: Not sure ResourceType is a good name. How about `ResourceRequest` or `ResourceRequirement`?

haowells: ok. let's use ResourceRequirement

operator/v1/types_klusterlet.go
```diff
@@ -86,6 +86,24 @@ type KlusterletSpec struct {
...
+	// ResourceRequirement specify QoS classes of klusterlet deployment
+	// +optional
+	ResourceRequirement *ResourceRequirement `json:"resourceRequirement,omitempty"`
+}
+
+type ResourceQosClass string
+
+const (
+	ResourceQosClassDefault    ResourceQosClass = "Default"
+	ResourceQosClassBestEffort ResourceQosClass = "BestEffort"
```
qiujian16, line 101: some comments maybe to indicate what besteffort means

haowells: comments added

qiujian16:
need to signoff. And you would need to run make update again.

PR #302, opened 2023-12-13
✨ add grpc option for cloudevents client.

morvencao:
## Summary

This PR tries to add gRPC option for cloudevents client.

## Related issue(s)

Fixes #


cloudevents/generic/options/grpc/options.go
```diff
@@ -0,0 +1,130 @@
...
+	CAFile         string
+	ClientCertFile string
+	ClientKeyFile  string
+}
+
+func NewGRPCOptions() *GRPCOptions {
+	return &GRPCOptions{}
+}
+
+func (o *GRPCOptions) AddFlags(flags *pflag.FlagSet) {
```
qiujian16: I think this should also be in the config similar as mqtt. It will be better it we share the same config file.

cloudevents/generic/options/grpc/protobuf/v1/cloudevent.proto
```diff
@@ -0,0 +1,80 @@
+syntax = "proto3";
+
+package io.cloudevents.v1;
+
+option go_package = "open-cluster-management.io/api/cloudevents/generic/options/grpc/protobuf/v1";
```
qiujian16, line 9: what is needed if the protobuf message is changed? We need some readme to explain.

morvencao: updated the grpc options to use config similar to mqtt.

cloudevents/generic/options/grpc/options.go
morvencao: updated the grpc options to use config similar to mqtt.

cloudevents/generic/options/grpc/protobuf/v1/cloudevent.proto
morvencao: added a readme file and some comments to guide developer how to update go code after modifying the proto file.

cloudevents/generic/options/grpc/options.go
```diff
@@ -0,0 +1,166 @@
...
+	"google.golang.org/grpc/credentials"
+	"google.golang.org/grpc/credentials/insecure"
+	"gopkg.in/yaml.v2"
+
+	cloudevents "github.com/cloudevents/sdk-go/v2"
+
+	"open-cluster-management.io/api/cloudevents/generic/options/grpc/protocol"
+)
+
+const (
```
skeeey, line 21: move these topics to a common file?

cloudevents/generic/options/grpc/agentoptions.go
```diff
@@ -0,0 +1,76 @@
...
+			errorChan:   make(chan error),
+			clusterName: clusterName,
+		},
+		AgentID:     agentID,
+		ClusterName: clusterName,
+	}
+}
+
+func (o *grpcAgentOptions) WithContext(ctx context.Context, evtCtx cloudevents.EventContext) (context.Context, error) {
+	eventType, err := types.ParseCloudEventsType(evtCtx.GetType())
```
skeeey, line 35: do we need this for grpc?

cloudevents/generic/options/grpc/options.go
morvencao: Currently, the topic structure for gRPC aligns with MQTT, but when we have a third option like Kafka, that would need a distinct pattern for topics. In that case, a common file for topics becomes unfeasible, make sense?

cloudevents/generic/options/grpc/agentoptions.go
morvencao: We require this information to assign the topic for event publish, and the topic will be included in the gRPC publish request, see: https://github.com/open-cluster-management-io/api/blob/3012fe56a2f617cfa03e2f55f660c6f530e437d4/cloudevents/generic/options/grpc/protobuf/v1/cloudevent.proto#L68-L73

cloudevents/generic/options/grpc/options.go
skeeey: yeah, but I think the ideal situation is to have unified topics :),  we can keep the topics in each option firstly

cloudevents/generic/options/grpc/protobuf/v1/README.md
```diff
@@ -0,0 +1,30 @@
+# CloudEvent gRPC Protobuf Definitions
```
skeeey, line 1: would you also link this in the https://github.com/open-cluster-management-io/api/blob/main/cloudevents/README.md, we have a client introduction in this doc, you can add a new option in the `CloudEventsOptions, it provides ...`

cloudevents/generic/options/grpc/options.go
morvencao: OK, let's organize topics within each option for now and and see if we can have a unified topics in the future.

cloudevents/generic/options/grpc/protobuf/v1/README.md
morvencao: Added grpc option to `cloudevents/README.md`

PR #307, opened 2024-01-02
:sparkles: Add nodeStatus.

xuezhaojun:
## Summary

Add node status section to represent ready, schedulable numbers of nodes on a managed cluster.

## Related issue(s)

Fixes #

### DIFF
--- a/cluster/v1/types.go
+++ b/cluster/v1/types.go
@@ -163,6 +163,25 @@ type ManagedClusterStatus struct {
 	// vendor or version specific and may not be included from all managed clusters.
 	// +optional
 	ClusterClaims []ManagedClusterClaim `json:"clusterClaims,omitempty"`
+
+	// NodeStatus represents the status of nodes on the managed cluster.
+	// +optional
+	NodeStatus NodeStatus `json:"nodeStatus,omitempty"`
+}
+
+// NodeStatus represents the status of nodes on the managed cluster.
+type NodeStatus struct {
+	// Ready represents the number of ready nodes on the managed cluster.
+	// +optional
+	Ready int32 `json:"ready,omitempty"`
+
+	// Schedulable represents the number of schedulable nodes on the managed cluster.
+	// +optional
+	Schedulable int32 `json:"schedulable,omitempty"`
+
+	// Total represents the total number of nodes on the managed cluster.
+	// +optional
+	Total int32 `json:"total,omitempty"`
 }
 
 // ManagedClusterVersion represents version information about the managed cluster.
--- a/cluster/v1/0000_00_clusters.open-cluster-management.io_managedclusters.crd.yaml
+++ b/cluster/v1/0000_00_clusters.open-cluster-management.io_managedclusters.crd.yaml
@@ -263,6 +263,26 @@ spec:
                   - type
                   type: object
                 type: array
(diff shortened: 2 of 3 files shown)


cluster/v1/types.go
```diff
@@ -163,6 +163,25 @@ type ManagedClusterStatus struct {
...
+	// +optional
+	Ready int32 `json:"ready,omitempty"`
+
+	// Schedulable represents the number of schedulable nodes on the managed cluster.
+	// +optional
+	Schedulable int32 `json:"schedulable,omitempty"`
+
+	// Total represents the total number of nodes on the managed cluster.
+	// +optional
+	Total int32 `json:"total,omitempty"`
```
qiujian16, line 184: Let's think more carefully on this field? will total/ready useful for scheduling? what is the relationship between ready and schedulable? How do we extend in the future. e.g. if want to schedule to node with certain labels?

xuezhaojun: The original thought of `ready` and `schedulable` is to follow the definetion in the kubernetes repo:

https://github.com/kubernetes/kubernetes/blob/138ac71fb12bdd4f24b4b1b10c054e07c55f4364/test/e2e/framework/node/resource.go#L446C1-L464C2

```go
// IsNodeSchedulable returns true if:
// 1) doesn't have "unschedulable" field set
// 2) it also returns true from IsNodeReady
func IsNodeSchedulable(node *v1.Node) bool {
	if node == nil {
		return false
	}
	return !node.Spec.Unschedulable && IsNodeReady(node)
}

// IsNodeReady returns true if:
// 1) it's Ready condition is set to true
// 2) doesn't have NetworkUnavailable condition set to true
func IsNodeReady(node *v1.Node) bool {
	nodeReady := IsConditionSetAsExpected(node, v1.NodeReady, true)
	networkReady := isConditionUnset(node, v1.NodeNetworkUnavailable) ||
		IsConditionSetAsExpectedSilent(node, v1.NodeNetworkUnavailable, false)
	return nodeReady && networkReady
}
```

From the above codes: **`schedulable` is a subset of `ready`.**

---

From my perspective, the `nodeStatus` section we add here is not aim to be used directly in schedule. It provides helper msg for user when they find workloads are not deployed as they expected. 

If we want to schedule based on node-level information, we should first convert it to managed-cluster level information. For example, if we need to schedule workloads to nodes with the `gpu:true` label:
* first we need to mark every managed cluster which has a `gpu:true` node. 
* then we schedule manifestworks to these `gpu:true` managed clusters by using placement.

qiujian16: But why total and ready is needed even if it is just to give info to user on why workload is not scheduled? only mark it as having gpu may not sufficient...also what if there are schedule nodes but all gpu node is unschedulable?  I do not have a clear answer.

xuezhaojun: How about adding a single field `Unschedulable` in the status or spec like node did: (`node.Spec.Unschedulable`)?

And this field serves for a single purpose -- when all nodes in this managed cluster are `unscheduable` the field turns to `true`.

We can keep collecting cases about scheduling based on node information and extend the scope later.

xuezhaojun:
The cases are not clear enough at the point of time. Close and wait for more informations provides by users.

PR #308, opened 2024-01-02
:seedling: add sequence id for status update event

skeeey:
## Related issue(s)

Fixes #


cloudevents/work/agent/codec/manifest.go
```diff
@@ -29,12 +30,14 @@ const (
 
 // ManifestCodec is a codec to encode/decode a ManifestWork/cloudevent with ManifestBundle for an agent.
 type ManifestCodec struct {
-	restMapper meta.RESTMapper
+	sequenceGenerator *snowflake.Node
```
qiujian16: this is only needed for status update, right？ I do not think a developer of consumer need to care about this.

cloudevents/generic/types/types.go
```diff
@@ -48,6 +48,10 @@ const (
 	// ExtensionResourceVersion is the cloud event extension key of the resource version.
 	ExtensionResourceVersion = "resourceversion"
 
+	// ExtensionSequenceID is the cloud event extension key of the event sequence ID.
+	// The event sequence id represents the order in which events occur on the agent/source.
```
qiujian16: not the strict order. I think you can only use it to order the event from one single agent. I do not think you need this for source.

cloudevents/work/agent/codec/manifest.go
skeeey: right, the agent generates the sequence id only when it creates the status update event.  

and creating the snowflake  generator when the agent init

cloudevents/generic/types/types.go
skeeey: yeah, the source does not need this, it already has resoruceversion, update the description and variable name to make this more clear

cloudevents/work/agent/codec/manifest.go
```diff
@@ -27,6 +28,16 @@ const (
 	CloudEventsOriginalSourceAnnotationKey = "cloudevents.open-cluster-management.io/originalsource"
 )
 
+var sequenceGenerator *snowflake.Node
+
+func init() {
+	// init the snowflake id generator, for each single agent, they send events independently, we set the snowflake
+	// node id to 1 for one single agent, that means, we can sort events sent by the same agent.
```
qiujian16:
```suggestion
	// init the snowflake id generator with node id 1 for each single agent. 
	// Each single agent has its own consumer id to be identified, and we 
	// can ensure the order of status update event from the same agent via
	// sequence id. The events from different agents are independent, hence
	// the ordering among them needs not to be guaranteed.
```

qiujian16 (commented):
update the comment, others look good

/approve

morvencao (commented):
/lgtm

PR #311, opened 2024-01-09
:sparkles: add QPS and burst for the agent components

zhiweiyin318:
## Related issue(s)

Fixes #


operator/v1/types_clustermanager.go
```diff
@@ -114,6 +119,11 @@ type WorkConfiguration struct {
 	//  	he can set featuregate/Foo=false before upgrading. Let's say the cluster-admin wants featuregate/Foo=false.
 	// +optional
 	FeatureGates []FeatureGate `json:"featureGates,omitempty"`
+
+	// FlagOptions represents a list of flag options for the component binary.
+	// The key is the option name, the value is the option value.
+	// The FlagOptions can override all configured options including the default options and FeatureGates, etc.
+	FlagOptions map[string]string `json:"flagOptions,omitempty"`
```
qiujian16: would be better if it is a list with key/value. What if this and featuregates are all set?

zhiweiyin318: changed to a list with name/value.
My thought is that the configured flagOptions can override all default configuration including the featuregates, ClientCertExpirationSeconds and ClusterAnnotations, and the featuregates, ClientCertExpirationSeconds and ClusterAnnotations could be deprecated in the future.

operator/v1/types_clustermanager.go
```diff
@@ -152,6 +170,13 @@ const (
 	FeatureGateModeTypeDisable FeatureGateModeType = "Disable"
 )
 
+type FlagOptions struct {
+	// Name is the flag option name of the component binary.
+	Name string `json:"name,omitempty"`
```
qiujian16: required, and need to do validation

operator/v1/types_clustermanager.go
```diff
@@ -152,6 +170,13 @@ const (
...
+	// Value is the value of the flag option.
+	Value string `json:"value,omitempty"`
```
qiujian16: required

operator/v1/types_clustermanager.go
```diff
@@ -126,6 +138,12 @@ type AddOnManagerConfiguration struct {
 	//  	he can set featuregate/Foo=false before upgrading. Let's say the cluster-admin wants featuregate/Foo=false.
 	// +optional
 	FeatureGates []FeatureGate `json:"featureGates,omitempty"`
+
+	// FlagOptions represents a list of flag options for the component binary.
+	// The key is the option name, the value is the option value.
+	// The FlagOptions can override all configured options including the default options and FeatureGates, etc.
+	// +optional
```
qiujian16: set this to listMap

operator/v1/types_klusterlet.go
```diff
@@ -157,6 +157,44 @@ type RegistrationConfiguration struct {
 	// ManagedCluster when creating only, other actors can update it afterwards.
 	// +optional
 	ClusterAnnotations map[string]string `json:"clusterAnnotations,omitempty"`
+
+	// KubeAPIQPS indicates the maximum QPS while talking with apiserver of hub cluster from the spoke cluster.
+	// If it is set empty, use the default value: 50
+	// +optional
+	// +kubebuilder:default:=50
+	// +kubebuilder:validation:Pattern=^(-?\d+)(.\d+)?$
+	KubeAPIQPS string `json:"qps,omitempty"`
```
qiujian16: the json field should be also kubeAPIQPS, why it is not int32 with max?

zhiweiyin318: changed to kubeAPIQPS. the QPS is float, but float is not suggested to be used in API, so use string here and added a validation.

qiujian16 (commented):
/approve
/lgtm

PR #316, opened 2024-01-18
✨ Configurable qos resources for deployments managed by clustermanager and klusterlet

dongbeiqing91:
## Summary

Support configurable qos resources for deployments managed by clustermanager and klusterlet because "Default" deployment resources w/ only hard coded "resources" but no "limits" is not good enough and "BestEffort" is usually not allowed in production.
## Related issue(s)

Fixes #


operator/v1/types_klusterlet.go
```diff
@@ -87,25 +90,30 @@ type KlusterletSpec struct {
...
 )
 
 // ResourceRequirement allow user override the default pod QoS classes
 type ResourceRequirement struct {
 	// +kubebuilder:validation:Enum=Default;BestEffort
 	// +kubebuilder:default:=Default
 	Type ResourceQosClass `json:"type"`
+	// Resources defines resource requests and limits when Type is ResourceQosClassConfigurable
+	// +optional
+	Resources *corev1.ResourceList `json:"resources,omitempty"`
```
qiujian16: You would need to add ResourceQosClassConfigurable in the // +kubebuilder:validation:Enum=Default;BestEffort for type.

I would suggest the type is ResourceRequirements and the field name should be ResourceRequirements too.

qiujian16: I do not think *corev1.ResourceList is the correct type. What you want is https://github.com/kubernetes/api/blob/master/core/v1/types.go#L2547?

operator/v1/types_clustermanager.go
```diff
@@ -70,6 +70,10 @@ type ClusterManagerSpec struct {
 	// AddOnManagerConfiguration contains the configuration of addon manager
 	// +optional
 	AddOnManagerConfiguration *AddOnManagerConfiguration `json:"addOnManagerConfiguration,omitempty"`
+
+	// ResourceRequirement specify QoS classes of deployments managed by clustermanager
+	// +optional
+	ResourceRequirement *ResourceRequirement `json:"resourceRequirement,omitempty"`
```
qiujian16, line 77: it should be noted that it applies to all containers since there are multiple containers running in the cluster manager.

zhiweiyin318:
the deepcopy go file is not updated, is there any error durning running the `make update`?

operator/v1/types_klusterlet.go
dongbeiqing91: Added ResourceQosClassConfigurable in the validation comment

> I would suggest the type is ResourceRequirements and the field name should be ResourceRequirements too

Do you mean changing 
```
ResourceRequirement *ResourceRequirement `json:"resourceRequirement,omitempty"`
```
to
```
ResourceRequirements *ResourceRequirements `json:"resourceRequirement,omitempty"`
```
But `json:"resourceRequirement,omitempty"` has to stay unchanged, right?

dongbeiqing91: Yes you are right. Fixed.

operator/v1/types_clustermanager.go
dongbeiqing91: comment added

dongbeiqing91:
> the deepcopy go file is not updated, is there any error durning running the `make update`?

It indeeded had some errors I didn't pay attention to. Let me check and fix it

operator/v1/types_klusterlet.go
qiujian16: ResourceRequirement *ResourceRequirement `json:"resourceRequirement,omitempty"` is ok

dongbeiqing91:
> the deepcopy go file is not updated, is there any error durning running the `make update`?

The project path was wrong. I have fixed it. deepcopy is correctly generated now.

operator/v1/types_klusterlet.go
```diff
@@ -87,25 +90,31 @@ type KlusterletSpec struct {
...
 type ResourceQosClass string
 
 const (
-	// Default use resource setting in the template file
+	// Default use resource setting in the template file (with requests but no limits in the resources)
 	ResourceQosClassDefault ResourceQosClass = "Default"
 	// If all containers in the pod don't set resource request and limits, the pod is treated as BestEffort.
 	ResourceQosClassBestEffort ResourceQosClass = "BestEffort"
+	// Configurable resources with requests and limits
+	ResourceQosClassConfigurable ResourceQosClass = "Configurable"
```
qiujian16: the type name should be ResourceRequirement

ResourceQosClassResourceRequirement ResourceQosClass = "ResourceRequirement"

dongbeiqing91: updated

operator/v1/types_klusterlet.go
```diff
@@ -87,25 +90,31 @@ type KlusterletSpec struct {
...
 
 // ResourceRequirement allow user override the default pod QoS classes
 type ResourceRequirement struct {
-	// +kubebuilder:validation:Enum=Default;BestEffort
+	// +kubebuilder:validation:Enum=Default;BestEffort;ResourceRequirement
 	// +kubebuilder:default:=Default
 	Type ResourceQosClass `json:"type"`
+	// Resources defines resource requests and limits when Type is ResourceQosClassResourceRequirement
+	// +optional
+	Resources *corev1.ResourceRequirements `json:"resources,omitempty"`
```
qiujian16:
```suggestion
	ResourceRequirement *corev1.ResourceRequirements `json:"resourceRequirement,omitempty"`
```

dongbeiqing91: if this change is made, the yaml will look like:
```
resourceRequirement:
  resourceRequirement:
    limits: 
       ...
```
duplicated `resourceRequirement`s look weild I suppose

dongbeiqing91: In kubernetes Container struct, it is
```
Resources ResourceRequirements `json:"resources,omitempty" protobuf:"bytes,8,opt,name=resources"`
```
https://github.com/kubernetes/api/blob/master/core/v1/types.go#L2682

zhiweiyin318:
the DCO Check requires a Signed-off-by trailer in the commit log.  so you need to add  `-s` when run `git commit`.

dongbeiqing91:
> the DCO Check requires a Signed-off-by trailer in the commit log. so you need to add `-s` when run `git commit`.

Done

operator/v1/types_klusterlet.go
qiujian16: it is intended to follow the union pattern https://github.com/kubernetes/enhancements/blob/master/keps/sig-api-machinery/1027-api-unions/README.md. So either the type should be Resources or the filed name should be ResourceRequirement

dongbeiqing91: updated

PR #320, opened 2024-02-13
🌱 Regenerate informers with WithTransform

skitt:
## Summary

Since 1.29, Kubernetes supports transforming informers, which is useful to reduce the amount of memory used by the underlying caches. This updates the code generator to the latest 0.29 release and re-generates the informers.

## Related issue(s)

Fixes #


vthapar:
Can we get review on this? This will be very helpful for optimization work we're doing on SubmarinerAddon.

qiujian16:
would you run go mod tidy and go mod vendor?

skitt:
> would you run go mod tidy and go mod vendor?

I’d already done that, the verify-deps failure happened because CI was set up for Go 1.20 and the build now requires 1.21.

PR #321, opened 2024-03-11
:sparkles: Add ClientCertAdditionalData & HubSensitive

xuezhaojun:
## Summary

`HubSensitive` by default is true, for [CustomSignedCSR like cluster-proxy](https://github.com/xuezhaojun/cluster-proxy/blob/b0cc356681b6ccb4e60f989b932b92854e1fed1e/pkg/proxyagent/agent/agent.go#L78-L79) , it will add an addtionalData of hub CA hash when storing certificates to a secert. So that hub change will trigger the renewal of CSR.

For addons that don't want CSR renewl when changing hub(for example, the client certificates is used to access a consist URL service), the addon developer can set the `HubSensitive` to false, and control the renewal of CSR by set their own `ClientCertAddtionalData`.

### DIFF
--- a/addon/v1alpha1/types_managedclusteraddon.go
+++ b/addon/v1alpha1/types_managedclusteraddon.go
@@ -68,6 +68,18 @@ type RegistrationConfig struct {
 	//
 	// +optional
 	Subject Subject `json:"subject,omitempty"`
+
+	// HubSensitive is a flag to indicate whether the registrationConfig is sensitive to the hub.
+	// If it is set to true, the renewal of the CSR will be triggered when the hub is changed.
+	// +optional
+	// +kubebuilder:default=true
+	HubSensitive bool `json:"hubSensitive,omitempty"`
+
+	// The signed CSR client certificates will be stored as a secret on the agent, clientCertAdditionalData
+	// is the additional data that will be stored alongside with the client certificate in that secret.
+	// Also, the change of the additional data will trigger the CSR renewal.
+	// +optional
+	ClientCertAdditionalData map[string]string `json:"clientCertAdditionalData,omitempty"`
 }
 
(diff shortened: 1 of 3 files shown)


xuezhaojun:
/assign @qiujian16 

Please take the time to review this pull request.

addon/v1alpha1/types_managedclusteraddon.go
```diff
@@ -68,6 +68,18 @@ type RegistrationConfig struct {
 	//
 	// +optional
 	Subject Subject `json:"subject,omitempty"`
+
+	// HubSensitive is a flag to indicate whether the registrationConfig is sensitive to the hub.
+	// If it is set to true, the renewal of the CSR will be triggered when the hub is changed.
+	// +optional
+	// +kubebuilder:default=true
+	HubSensitive bool `json:"hubSensitive,omitempty"`
```
qiujian16, line 76: should not use bool type

addon/v1alpha1/types_managedclusteraddon.go
```diff
@@ -68,6 +68,18 @@ type RegistrationConfig struct {
...
+
+	// The signed CSR client certificates will be stored as a secret on the agent, clientCertAdditionalData
+	// is the additional data that will be stored alongside with the client certificate in that secret.
+	// Also, the change of the additional data will trigger the CSR renewal.
+	// +optional
+	ClientCertAdditionalData map[string]string `json:"clientCertAdditionalData,omitempty"`
```
qiujian16, line 82: it is not recommanded to use map since it is unbondary and behave weird when merge.

xuezhaojun:
/close 

Decidei to use subject groups and OU.

PR #327, opened 2024-04-29
✨ add AppliedManifestWorkEvictionGracePeriod in klusterlet

elgnay:
## Related issue(s)

Fixes #

### DIFF
--- a/operator/v1/types_klusterlet.go
+++ b/operator/v1/types_klusterlet.go
@@ -232,6 +232,14 @@ type WorkAgentConfiguration struct {
 	// +optional
 	// +kubebuilder:default:=100
 	KubeAPIBurst int32 `json:"kubeAPIBurst,omitempty"`
+
+	// AppliedManifestWorkEvictionGracePeriod is the eviction grace period the work agent will wait before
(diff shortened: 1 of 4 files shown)


operator/v1/types_klusterlet.go
```diff
@@ -232,6 +232,14 @@ type WorkAgentConfiguration struct {
 	// +optional
 	// +kubebuilder:default:=100
 	KubeAPIBurst int32 `json:"kubeAPIBurst,omitempty"`
+
+	// AppliedManifestWorkEvictionGracePeriod is the eviction grace period the work agent will wait before
+	// evicting the AppliedManifestWorks, whose corresponding ManifestWorks are missing on the hub cluster, from
+	// the managed cluster. If not present, the default value of the work agent will be used.
+	// +optional
+	// +kubebuilder:validation:Type=string
```
qiujian16, line 240: what does nil mean?

elgnay: Nil means the appliedManifestWorkEvictionGracePeriod is not specified. The default value of the work agent will be used.

elgnay: If a struct is used instead of a pointer, an empty struct will result in a zero value of the appliedManifestWorkEvictionGracePeriod.

PR #329, opened 2024-05-20
:seedling: Change KubeConfigSecret to a struct type.

xuezhaojun:
## Summary

The KubeConfigSecret should not only be string but a struct in case we want to add addtional information such as `context` in the future.


operator/v1/types_klusterlet.go
```diff
@@ -210,6 +210,10 @@ type LocalSecretsConfig struct {
 	HubConnectionTimeoutSeconds int32 `json:"hubConnectionTimeoutSeconds,omitempty"`
 }
 
+type KubeConfigSecret struct {
+	Name string `json:"name"`
```
qiujian16, line 216: comment is needed, is this required field?

xuezhaojun: Update.

qiujian16 (commented):
/approve
/lgtm

PR #331, opened 2024-05-20
:sparkles: Add `MultipleHubs` in the feature gate list.

xuezhaojun:
## Summary

The `MultipleHubs` is in alpha stage, is used by both operator and registration-agent.

### DIFF
--- a/feature/feature.go
+++ b/feature/feature.go
@@ -74,6 +74,9 @@ const (
 
 	// ResourceCleanup will start gc controller to clean up resources in cluster ns after cluster is deleted.
 	ResourceCleanup featuregate.Feature = "ResourceCleanup"
+
+	// MultipleHubs allows user to configure multiple bootstrapkubeconfig connecting to different hubs via Klusterlet and let agent decide which one to use
+	MultipleHubs featuregate.Feature = "MultipleHubs"
 )
 
 // DefaultSpokeRegistrationFeatureGates consists of all known ocm-registration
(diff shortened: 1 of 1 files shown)


feature/feature.go
```diff
@@ -74,6 +74,9 @@ const (
 
 	// ResourceCleanup will start gc controller to clean up resources in cluster ns after cluster is deleted.
 	ResourceCleanup featuregate.Feature = "ResourceCleanup"
+
+	// SwitchHub allows user to configure multiple bootstrapkubeconfig via Klusterlet and let agent to decide which one to use
+	SwitchHub featuregate.Feature = "SwitchHub"
```
qiujian16: MultipleHubs might be a better name. Thought @elgnay ?

xuezhaojun: Agree, the `MultipleHubs` is better, fixed.

feature/feature.go
```diff
@@ -74,6 +74,9 @@ const (
 
 	// ResourceCleanup will start gc controller to clean up resources in cluster ns after cluster is deleted.
 	ResourceCleanup featuregate.Feature = "ResourceCleanup"
+
+	// MultipleHubs allows user to configure multiple bootstrapkubeconfig via Klusterlet and let agent to decide which one to use
```
qiujian16:
```suggestion
	// MultipleHubs allows user to configure multiple bootstrapkubeconfig connecting to different hubs via Klusterlet and let agent decide which one to use
```

xuezhaojun: Updated.

PR #332, opened 2024-05-23
✨ ReadOnly type in work

qiujian16:
## Related issue(s)

Fixes #


work/v1/types.go
```diff
@@ -189,6 +190,11 @@ const (
 	// If there is conflict, the related Applied condition of manifest will be in the status of False with the
 	// reason of ApplyConflict. This type allows another controller on the spoke to control certain field of the resource.
 	UpdateStrategyTypeServerSideApply UpdateStrategyType = "ServerSideApply"
+
+	// UpdateStrategyTypeReadOnly type means only check the existence of the resource based on the resource's metadata.
```
clyang82, line 194: May update others to use field name in the comment instead of the value. for example: `UpdateStrategyTypeServerSideApply` type means to update resource.

elgnay:
LGTM

PR #334, opened 2024-05-29
:sparkles: Add MultipleHubs in default operator feature gates.

xuezhaojun:
## Summary

This PR is related to the https://github.com/open-cluster-management-io/api/pull/331 we want to control this feature in both operator and registration part.


xuezhaojun:
/close

Operator(klusterlet) uses the same feature gate configuration field as registration-agent.

PR #339, opened 2024-08-23
:sparkles: Aws iam auth api change request

suvaanshkumar:
## Related issue(s)

Fixes #


qiujian16:
you would need to update the types.go at first, and then run make update to generate crd and client code. Ensure to run make verify after make update. Here https://github.com/open-cluster-management-io/api/blob/main/docs/development.md has some infos.

EricaJ6:
Thanks @qiujian16 for the suggestions. Please take a look at the updated changes with required changes.

operator/v1/types_klusterlet.go
```diff
@@ -172,6 +172,28 @@ type RegistrationConfiguration struct {
...
+
+type RegistrationDriver struct {
+	// Type of the authentication used by managedcluster to register as well as pull work from hub. Possible values are csr and awsirsa.
+	// +kubebuilder:default:=csr
+	// +kubebuilder:validation:Enum=csr;awsirsa
+	AuthType string `json:"authType"`
+
+	// Contain the details required for registering with eks hub cluster using IAM roles for service account.
+	// This is required only when the authType is awsirsa.
+	AwsIrsa AwsIrsa `json:"awsIrsa,omitempty"`
```
qiujian16: this should be the pointer

operator/v1/types_klusterlet.go
```diff
@@ -172,6 +172,28 @@ type RegistrationConfiguration struct {
...
+}
+
+type AwsIrsa struct {
+	// The arn of the EKS hub cluster. This will be required to pass information to hub, which hub will use to create IAM identities for this klusterlet.
+	// Example - arn:eks:us-west-2:12345678910:cluster/hub-cluster1.
+	// +required
+	EksHubClusterArn string `json:"eksHubClusterArn"`
```
qiujian16: maybe just hubClusterArn

operator/v1/types_klusterlet.go
```diff
@@ -172,6 +172,28 @@ type RegistrationConfiguration struct {
...
+	// +kubebuilder:validation:Enum=csr;awsirsa
+	AuthType string `json:"authType"`
+
+	// Contain the details required for registering with eks hub cluster using IAM roles for service account.
+	// This is required only when the authType is awsirsa.
+	AwsIrsa *AwsIrsa `json:"awsIrsa,omitempty"`
+}
+
+type AwsIrsa struct {
+	// The arn of the EKS hub cluster. This will be required to pass information to hub, which hub will use to create IAM identities for this klusterlet.
```
mikeshng: @qiujian16 do we still need to mention EKS here specifically after your field rename suggestion? Should we do something like: "The arn of the hub cluster (ie: an EKS cluster)."

operator/v1/types_klusterlet.go
```diff
@@ -172,6 +172,28 @@ type RegistrationConfiguration struct {
...
+	RegistrationDriver RegistrationDriver `json:"registrationDriver,omitempty"`
+}
+
+type RegistrationDriver struct {
+	// Type of the authentication used by managedcluster to register as well as pull work from hub. Possible values are csr and awsirsa.
+	// +kubebuilder:default:=csr
+	// +kubebuilder:validation:Enum=csr;awsirsa
+	AuthType string `json:"authType"`
+
+	// Contain the details required for registering with eks hub cluster using IAM roles for service account.
```
mikeshng:
```suggestion
	// Contain the details required for registering with hub cluster (ie: an EKS cluster) using AWS IAM roles for service account.
```
I think it's better to be consistent and use EKS in upper case in all the places.

operator/v1/types_klusterlet.go
EricaJ6: This arn is specific to the EKS cluster only. I can update the description to `The arn of the hub cluster (ie: an EKS cluster).`

qiujian16:
/approve

@mikeshng @elgnay would you take the review and tag?

elgnay:
LGTM

PR #344, opened 2024-11-04
🌱 Set `DefaultClusterSet` default value to true in feature gate

RokibulHasan7:
## Related issue(s)

Fixes #

### DIFF
--- a/feature/feature.go
+++ b/feature/feature.go
@@ -96,7 +96,7 @@ var DefaultSpokeRegistrationFeatureGates = map[featuregate.Feature]featuregate.F
 // feature keys for registration hub controller.  To add a new feature, define a key for it above and
 // add it here.
 var DefaultHubRegistrationFeatureGates = map[featuregate.Feature]featuregate.FeatureSpec{
-	DefaultClusterSet:          {Default: false, PreRelease: featuregate.Alpha},
+	DefaultClusterSet:          {Default: true, PreRelease: featuregate.Alpha},
 	V1beta1CSRAPICompatibility: {Default: false, PreRelease: featuregate.Alpha},
 	ManagedClusterAutoApproval: {Default: false, PreRelease: featuregate.Alpha},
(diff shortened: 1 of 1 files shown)


feature/feature.go
```diff
@@ -96,7 +96,7 @@ var DefaultSpokeRegistrationFeatureGates = map[featuregate.Feature]featuregate.F
 // feature keys for registration hub controller.  To add a new feature, define a key for it above and
 // add it here.
 var DefaultHubRegistrationFeatureGates = map[featuregate.Feature]featuregate.FeatureSpec{
-	DefaultClusterSet:          {Default: false, PreRelease: featuregate.Alpha},
+	DefaultClusterSet:          {Default: true, PreRelease: featuregate.Alpha},
```
qiujian16, line 99: commonly alpha feature should be false by default, we would need to promote it to beta feature.

RokibulHasan7: I see AddonManagement is also an alpha feature, and it's set to True. any plans to make it beta?

qiujian16: hrm, it is not desired to be still in alpha, it should also be promoted to beta.

RokibulHasan7: Should I promote both to beta?

qiujian16: Think about it twice. Any reason why you want to enable it by default, do you want to always create the default clusterset or global clusterset?

RokibulHasan7: We're enabling the DefaultClusterSet from [here](https://github.com/open-cluster-management-io/ocm/blob/7fd733082f5cf4de1af969b2edc6da1c07f14904/pkg/operator/helpers/helpers.go#L76) 
So, I think we can set Default true.

qiujian16: hrm that makes sense. then after this is enabled by false, i think we can remove the code in https://github.com/open-cluster-management-io/ocm/blob/7fd733082f5cf4de1af969b2edc6da1c07f14904/pkg/operator/helpers/helpers.go#L76

PR #345, opened 2024-11-12
✨ Add an IgnoreFields option to select whether to override resource by mw

qiujian16:
The option is to disable override resource on the spoke by manifestwork when the resource is changed by another actor on the spoke

## Related issue(s)

Fixes #


work/v1/types.go
```diff
@@ -209,6 +217,26 @@ type ServerSideApplyConfig struct {
...
+	// Fields in JSONPaths are all ignored when condition is met, otherwise no fields is ignored
+	// in the apply operation.
+	// +kubebuilder:default=OnSpokePresent
+	// +kubebuilder:validation:Enum=OnSpokePresent;OnSpokeChange
+	// +kubebuilder:validation:Required
+	// +required
+	Condition IgnoreFieldsCondition `json:"condition"`
+
+	// JSONPaths defines the list of json path in the resource to be ignored
+	JSONPaths []string `json:"jsonPaths"`
```
zhujian7, line 242: should this be a required field? if it is empty, what should we do?

qiujian16: I do not think list can be required, but we could set the minItem to 1

qiujian16: updated

zhujian7:
LGTM
/assign @elgnay @haoqing0110

PR #346, opened 2024-11-20
🐛 Make the work feedback rule jaonpath name unique

zhujian7:
## Related issue(s)

Fixes #

### DIFF
--- a/work/v1/types.go
+++ b/work/v1/types.go
@@ -94,6 +94,8 @@ type ManifestConfigOption struct {
 
 	// FeedbackRules defines what resource status field should be returned. If it is not set or empty,
 	// no feedback rules will be honored.
+	// +listType:=map
+	// +listMapKey:=type
 	// +optional
 	FeedbackRules []FeedbackRule `json:"feedbackRules,omitempty"`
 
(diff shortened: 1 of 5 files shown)


work/v1/types.go
```diff
@@ -94,6 +94,8 @@ type ManifestConfigOption struct {
 
 	// FeedbackRules defines what resource status field should be returned. If it is not set or empty,
 	// no feedback rules will be honored.
+	// +listType:=map
+	// +listMapKey:=type
```
zhujian7: Here, I am not sure if we should treat the `type` as the map key, if we do, If we do this, it may be incompatible, but if we do not, users can still configure two same name jsonpaths:
```
feedbackRules:
- type: JSONPaths
   jsonPaths:
   - name: "Replica"
      path: .spec.replicas
- type: JSONPaths
   jsonPaths:
   - name: "Replica"
      path: .status.replicas
```

@qiujian16 @zhiweiyin318 WDYT?

qiujian16: let's do name at first...I am not sure if anyone uses multiple jsonPaths today, not quite likely.

PR #349, opened 2024-11-26
:sparkles: Agent leader election config

haoqing0110:
## Summary

Previously submitted PR https://github.com/open-cluster-management-io/ocm/pull/695 to enable leader election of klusterlet-agent on single node managed cluster. It breaks some user e2e testing cases as they are using SNO for testing https://github.com/open-cluster-management-io/ocm/pull/708, enabling the SNO leader election might cause the agent to wait as bad as 137s to start working.

Create this PR to make the agent leader election time configurable by `klusterletconfig`, so that user can reduce the time in testing. 

## Related issue(s)

https://github.com/open-cluster-management-io/ocm/pull/695 
https://github.com/open-cluster-management-io/ocm/pull/708


operator/v1/types_klusterlet.go
```diff
@@ -163,6 +163,10 @@ type RegistrationConfiguration struct {
 	// +kubebuilder:default:=100
 	KubeAPIBurst int32 `json:"kubeAPIBurst,omitempty"`
 
+	// Leader election configuration
+	// +optional
+	LeaderElectionConfig `json:",inline"`
```
qiujian16, line 168: I do not think it needs to be inline.

operator/v1/types_klusterlet.go
```diff
@@ -278,6 +286,42 @@ type WorkAgentConfiguration struct {
 	AppliedManifestWorkEvictionGracePeriod *metav1.Duration `json:"appliedManifestWorkEvictionGracePeriod,omitempty"`
 }
 
+// Leader election configuration
+type LeaderElectionConfig struct {
+	// Disable leader election.
```
qiujian16, line 291: do we really need to expose every configurations here?

haoqing0110: Since the leader election will always be enabled in OCM single node case, and we are not 100% sure if it will impact the user's cases, it might be useful to provide a disable option in `Klusterlet`.

haoqing0110:
Close this as https://github.com/open-cluster-management-io/ocm/pull/727 will release the leadership when agent restarts and reduce the leader election time.

PR #350, opened 2024-12-06
✨ Allow custom signer signingCA to specify namespace in addOnTemplate

zhujian7:
## Related issue(s)
https://github.com/open-cluster-management-io/ocm/issues/688

Fixes #


mikeshng (approved):
/lgtm

Thanks Zhu Jian! Exactly what I needed!

qiujian16:
@zhujian7 I think addon manager will not by default has secret access permission for any ns? It has to be added on demand, right?

zhujian7:
> @zhujian7 I think addon manager will not by default has secret access permission for any ns? It has to be added on demand, right?

yes, the developer needs to ensure the addon-manger has the permission. I have added on the field comment

PR #354, opened 2024-12-28
✨ support add-on agent resource requirements settings

elgnay:
## Related issue(s)

Fixes #


addon/v1alpha1/types_addondeploymentconfig.go
```diff
@@ -58,6 +58,31 @@ type AddOnDeploymentConfigSpec struct {
...
 	// +kubebuilder:validation:Pattern=^[a-z0-9]([-a-z0-9]*[a-z0-9])?$
 	AgentInstallNamespace string `json:"agentInstallNamespace,omitempty"`
+
+	// Resources required by add-on agents.
+	// If a container matches multiple ContainerResourceRequirements, the last one in the array will
+	// take precedence.
+	// +optional
+	// +listType=map
+	// +listMapKey=containerID
+	AgentResources []ContainerResourceRequirements `json:"agentResources,omitempty"`
```
qiujian16: maybe just ResourceRequirements

addon/v1alpha1/types_addondeploymentconfig.go
```diff
@@ -58,6 +58,31 @@ type AddOnDeploymentConfigSpec struct {
...
+}
+
+// ContainerResourceRequirements defines resources required by one or a group of containers.
+type ContainerResourceRequirements struct {
+	// ContainerID is a unique identifier for an agent container. It consists of three parts: resource kind,
```
qiujian16: is this kind or resource type? e.g. should user use Deployment or deployments?

addon/v1alpha1/types_addondeploymentconfig.go
```diff
@@ -58,6 +58,31 @@ type AddOnDeploymentConfigSpec struct {
...
+	// resource name, and container name, separated by ':'. The format follows
+	// '{resource_kind}:{resource_name}:{container_name}'. Wildcards (*) can be used in any part to match
+	// multiple containers. For example, '*:*:*' matches all containers of the agent.
+	// +required
+	// +kubebuilder:validation:Required
+	// "+kubebuilder:validation:Pattern=`^.+:.+:.+$`
```
qiujian16: does this work with the begining of `"`?

addon/v1alpha1/types_addondeploymentconfig.go
elgnay: It's resource kind, such as Deployment. Of course, it could be resource type as well. Which one is better?

addon/v1alpha1/types_addondeploymentconfig.go
elgnay: The validation here is not strict. A value, like `":abc:def` is allowed.

addon/v1alpha1/types_addondeploymentconfig.go
elgnay: `ResourceRequirements` is fine with me.

addon/v1alpha1/types_addondeploymentconfig.go
qiujian16: it is better if it is group resource or group kind. Group might be needed since kind or resource names could be the same. Or alternatively, we only support certain type of resources.

addon/v1alpha1/types_addondeploymentconfig.go
elgnay: Updated

addon/v1alpha1/types_addondeploymentconfig.go
elgnay: Updated the validation pattern to limit the supported resource types.

qiujian16 (commented):
/approve
/lgtm

PR #355, opened 2025-01-03
:sparkles:Updated clustermanager API spec and generated files for spec change.

dtclxy64:
## Summary 
Updated clustermanager API spec and generated files for spec change.

## Related issue(s)
Fixes #


operator/v1/types_clustermanager.go
```diff
@@ -108,6 +108,19 @@ type RegistrationHubConfiguration struct {
 	//  	he can set featuregate/Foo=false before upgrading. Let's say the cluster-admin wants featuregate/Foo=false.
 	// +optional
 	FeatureGates []FeatureGate `json:"featureGates,omitempty"`
+
+	// This provides details to initialize Hub cluster
+	// +optional
+	RegistrationDriverHub RegistrationDriverHub `json:"registrationDriverHub,omitempty"`
```
qiujian16: we can call it authDriver or registrationDriver. Also I think this should be a list, we should be able to support multiple drivers at the same time on hub.

operator/v1/types_clustermanager.go
```diff
@@ -108,6 +108,19 @@ type RegistrationHubConfiguration struct {
 	//  	he can set featuregate/Foo=false before upgrading. Let's say the cluster-admin wants featuregate/Foo=false.
 	// +optional
 	FeatureGates []FeatureGate `json:"featureGates,omitempty"`
+
+	// This provides details to initialize Hub cluster
```
qiujian16: the comment needs to be updated, it is not clear what this drivers meant to do.

operator/v1/types_clustermanager.go
amrcoder: Updated with the suggestions

operator/v1/types_clustermanager.go
amrcoder: Modified the comment to give more clarity

operator/v1/types_clustermanager.go
qiujian16: could you make this as 

registrationDrivers []RegistrationDriver

amrcoder: We already have another struct with name RegistrationDriver in types_klusterlet.go. That is the reason it was initially named it as RegistrationDriverHub. So can we go with the second option that you had suggested which is authDriver?

qiujian16: it is ok that the struct name is RegistrationDriverHub, but let's make the field name to be registrationDrivers and make it as a slice.

amrcoder: Renamed to registrationDrivers and it was already changed to a slice in previous commit

operator/v1/types_clustermanager.go
```diff
@@ -108,6 +108,24 @@ type RegistrationHubConfiguration struct {
...
+	RegistrationDrivers []RegistrationDriverHub `json:"registrationDrivers,omitempty"`
+}
+
+type RegistrationDriverHub struct {
+
+	// Type of the authentication used by hub to initialize the Hub cluster. Possible values are csr and awsirsa.
+	// +required
+	// +kubebuilder:default:=csr
+	// +kubebuilder:validation:Enum=csr;awsirsa
+	AuthType string `json:"authType,omitempty"`
```
qiujian16, line 126: one minor comment,  we should add

// +listType=map
// +listMapKey=authType
here, so we will not set duplicate type

qiujian16:
/approve
Just one minor comment, feel free to merge it when it is resolved.

mikeshng:
/lgtm

I've verified that the following are added:

```
// +listType=map
// +listMapKey=authType
```

mikeshng (approved):

PR #347, opened 2024-11-25
:sparkles: Adding aws cluster-arn validation to klusterlet CRD

jaswalkiranavtar:
## Related issue(s)

Fixes # https://github.com/open-cluster-management-io/ocm/issues/514

PR #356, opened 2025-01-15
:sparkles: Adding hubClusterArn validation in spec

jaswalkiranavtar:
## Summary
Adding validation in spec to avoid adding programmatic validation in clusteradm and ocm

## Related issue(s)

Fixes #https://github.com/open-cluster-management-io/ocm/issues/514


mikeshng (commented):
/lgtm

This is similar to https://github.com/open-cluster-management-io/api/pull/347 which has been merged. So it looks fine to me.

qiujian16 (commented):
/approve
/lgtm

PR #358, opened 2025-02-24
:sparkles: Adding a new field to clustermanager API to pass list of aws tags to be added to aws resources

jaswalkiranavtar:
## Summary
Tags are required on all resources for cost attribution and identification purpose.

## Related issue(s)

Fixes #https://github.com/open-cluster-management-io/ocm/issues/514


operator/v1/types_clustermanager.go
```diff
@@ -125,7 +125,7 @@ type RegistrationDriverHub struct {
 	// +kubebuilder:validation:Enum=csr;awsirsa
 	AuthType string `json:"authType,omitempty"`
 
-	// This represents the hub cluster ARN
+	// This represents the hub cluster ARN. Applicable to only awsirsa authentication type.
```
mikeshng: Nit: Could you group the `Applicable to only awsirsa authentication type` fields together. 

To me the order should be: AuthType, AutoApprovedIdentities, HubClusterArn, Tags

operator/v1/types_clustermanager.go
```diff
@@ -135,6 +135,12 @@ type RegistrationDriverHub struct {
 	// For awsirsa authentication type, AutoApprovedIdentities represent a list of approved arn patterns
 	// +optional
 	AutoApprovedIdentities []string `json:"autoApprovedIdentities,omitempty"`
+
+	// List of tags to be added to AWS resources created by hub while processing awsirsa registration request.
+	// Applicable to only awsirsa authentication type.
+	// Example - "product:v1:tenant:app-name=My-App"
```
mikeshng, line 155: I am just wondering is `app-name=My-App` close to the real usage of this field? If it is, please ignore this comment. Is better to have example like primary/secondary hub? Just guessing.

mikeshng (commented):
Left some nit comments. 

I am ok with this but maybe consider the option of just use the cluster manager CR's labels/annotations (maybe follow some predefined regex/key). Let see what Qiu Jian thinks.

jaswalkiranavtar:
> Left some nit comments.
> 
> I am ok with this but maybe consider the option of just use the cluster manager CR's labels/annotations (maybe follow some predefined regex/key). Let see what Qiu Jian thinks.

We thought about it. In our case the keys for k8s labels and aws tags and complete different to the extent that regex cannot be used to replace them. Otherwise it was easier for us to just copy the labels from registration controller and use them as tags.

operator/v1/types_clustermanager.go
jaswalkiranavtar: Yes appname is the real usage, we have tags like appname, createdby, deptname etc

We are planning to reuse same roles for both primary and secondary, which is why we are not adding primary/secondary for now.

qiujian16:
hrm I was thinking the API should look like:

```
drivers:
- type: csr
   csr:
       approveIdentities:
       - user1
- type: awsisra
   awsisra:
        approvedIdentities:
        - arn:.....
        tag:
        - .....
```

It means in RegistrationDriverHub we had a wrapped configuration struct for different drivers, which easily separated the configurations for different drivers.

qiujian16:
https://github.com/open-cluster-management-io/api/pull/359 as an example

operator/v1/types_clustermanager.go
mikeshng: I can't resolve this thread. Please feel free to resolve it if you are able to. Thanks.

operator/v1/types_clustermanager.go
mikeshng: I can't resolve this thread. Please feel free to resolve it if you are able to. Thanks.

operator/v1/types_clustermanager.go
```diff
@@ -125,16 +125,36 @@ type RegistrationDriverHub struct {
 	// +kubebuilder:validation:Enum=csr;awsirsa
 	AuthType string `json:"authType,omitempty"`
 
+	// CSR represents the configuration for csr driver.
+	// +optional
+	CSR *CSRDriverConfig `json:"csr,omitempty"`
+
+	// AwsIrsa represents the configuration for awsisra driver.
+	// +optional
+	AwsIrsa *AwsIrsaConfig `json:"awsisra,omitempty"`
```
mikeshng, line 134:
```suggestion
	AwsIrsa *AwsIrsaDriverConfig `json:"awsisra,omitempty"`
```

To match the pattern of above `CSRDriverConfig`. I think both should either include "Driver" or both should omit it. Same comment for the other occurrences.

mikeshng (approved):
/lgtm

PR #361, opened 2025-03-17
✨ Added support for sigs.k8s.io about api

gnana997:
## Summary
Added support for Cluster Properties through sigs.k8s.io/about-api.
  - To use the feature in ocm to add support for cluster properties
 
## Related issue(s)
   https://github.com/open-cluster-management-io/ocm/issues/838

### DIFF
--- a/feature/feature.go
+++ b/feature/feature.go
@@ -23,6 +23,10 @@ const (
 	// itself to avoid impact to users.
 	ClusterClaim featuregate.Feature = "ClusterClaim"
 
+	// ClusterProperty is a feature gate on hub controller and spoke-agent. When it is enabled on the
+	// spoke agent, it will use the claim controller to manage the managed cluster property
+	ClusterProperty featuregate.Feature = "ClusterProperty"
+
 	// AddonManagement is a feature gate on hub controller and spoke-agent. When it is enabled on the
 	//spoke agent, it will start a new controllers to manage the managed cluster addons
 	// registration and maintains the status of managed cluster addons through watching their leases.
@@ -90,6 +94,7 @@ const (
 // add it here.
 var DefaultSpokeRegistrationFeatureGates = map[featuregate.Feature]featuregate.FeatureSpec{
 	ClusterClaim:               {Default: true, PreRelease: featuregate.Beta},
+	ClusterProperty:            {Default: false, PreRelease: featuregate.Alpha},
 	AddonManagement:            {Default: true, PreRelease: featuregate.Beta},
 	V1beta1CSRAPICompatibility: {Default: false, PreRelease: featuregate.Alpha},
 	MultipleHubs:               {Default: false, PreRelease: featuregate.Alpha},
--- a/feature/feature.go
+++ b/feature/feature.go
@@ -23,6 +23,10 @@ const (
 	// itself to avoid impact to users.
 	ClusterClaim featuregate.Feature = "ClusterClaim"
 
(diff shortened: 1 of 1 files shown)


gnana997:
Hi @qiujian16 , I have added new CRD and support for clusterProperties including the feature gate to add support for clusterProperties in ocm package.

Can you please let me know if I miss anything?

Thanks!

cluster/v1/types.go
```diff
@@ -163,6 +163,14 @@ type ManagedClusterStatus struct {
...
 	// +optional
 	ClusterClaims []ManagedClusterClaim `json:"clusterClaims,omitempty"`
+
+	// ClusterProperties represents cluster properties that a managed cluster claims,
+	// for example a unique cluster identifier (id.k8s.io) and kubernetes version
+	// (kubeversion.open-cluster-management.io). They are written from the managed
+	// cluster. The set of properties is not uniform across a fleet, some properties can be
+	// vendor or version specific and may not be included from all managed clusters.
+	// +optional
+	ClusterProperties []ManagedClusterProperty `json:"clusterProperties,omitempty"`
```
qiujian16: I do not think we need to add this field, we could still use the clusterClaims, with an updated comment.

cluster/v1alpha1/types.go
```diff
@@ -59,3 +59,42 @@ var ReservedClusterClaimNames = [...]string{
 	// product name, like OpenShift, Anthos, EKS and GKE
 	"product.open-cluster-management.io",
 }
+
+// New ClusterProperty definitions added below...
```
qiujian16: also no need to add this in our API, we should use the existing about API.

qiujian16 (commented):
We only need to add feature gate in api repo. Not necessary to define the API here.

cluster/v1/types.go
gnana997: Hi @qiujian16 , Got it. I have updated the code.
Thanks!

cluster/v1alpha1/types.go
gnana997: Hi @qiujian16 , Got it. I have updated the code.
Thanks!

feature/feature.go
```diff
@@ -90,6 +94,7 @@ const (
 // add it here.
 var DefaultSpokeRegistrationFeatureGates = map[featuregate.Feature]featuregate.FeatureSpec{
 	ClusterClaim:               {Default: true, PreRelease: featuregate.Beta},
+	ClusterProperty:            {Default: true, PreRelease: featuregate.Beta},
```
qiujian16: let's start with alpha and disabled by default

gnana997: Got it. I have updated and set it to false.
Thanks!

feature/feature.go
```diff
@@ -90,6 +94,7 @@ const (
 // add it here.
 var DefaultSpokeRegistrationFeatureGates = map[featuregate.Feature]featuregate.FeatureSpec{
 	ClusterClaim:               {Default: true, PreRelease: featuregate.Beta},
+	ClusterProperty:            {Default: false, PreRelease: featuregate.Beta},
```
qiujian16: featuregate.Alpha

gnana997:
Hi @qiujian16 , Can you please release the new version? I will use it in the ocm package to support the cluster property informer based on the feature gate.
Thanks!

qiujian16:
commonly we do not release with a single change. You can update the go mod by specifying the commit on the main branch
something like
go get open-cluster-management.io/api@{commit id}

PR #362, opened 2025-03-20
✨ Workload conditions

bhperry:
## Summary

Adding the necessary type changes to ManifestWork to support the [workload conditions and completion](https://github.com/open-cluster-management-io/enhancements/tree/main/enhancements/sig-architecture/138-workload-completion) feature.

## Related issue(s)
https://github.com/open-cluster-management-io/enhancements/issues/138


work/v1/types.go
```diff
@@ -101,8 +110,61 @@ type ManifestConfigOption struct {
 	// if it is not set.
 	// +optional
 	UpdateStrategy *UpdateStrategy `json:"updateStrategy,omitempty"`
+
+	// ConditionRules defines how to set manifestwork conditions for a specific manifest.
+	// +optional
```
qiujian16, line 117: you might want to add a marker `listMapKey: condition` and set strategy to merge.

work/v1/types.go
```diff
@@ -101,8 +110,61 @@ type ManifestConfigOption struct {
...
+	Condition string `json:"condition"`
+
+	// Type defines how a manifest should be evaluated for a condition.
+	// It can be CEL, or WellKnownCompletions.
+	// If the type is CEL, user should specify the celExpressions field
+	// If the type is WellKnownCompletions, certain common types in k8s.io/api will be considered
+	// completed as defined by hardcoded rules.
+	// +kubebuilder:validation:Required
+	// +required
+	Type ConditionRuleType `json:"type"`
```
qiujian16, line 138: need to set an kubebuilder:validation:Enum marker

work/v1/types.go
```diff
@@ -101,8 +110,61 @@ type ManifestConfigOption struct {
...
+
+	// CelExpressions defines the CEL expressions to be evaluated for the condition.
+	// Final result is the logical AND of all expressions.
+	// +optional
```
qiujian16, line 142: I am wondering whether with type WellKnownCompletions, we do not need to set any other field in this struct.
If so, we might have a nested struct and put all field not needed for WellKnownCompletions to it, e.g.

```go
type ConditionRule struct {
    Condition string `json:"condition"`
    Type ConditionRuleType `json:"type"`
   // this could be a pointer, so when type is WellKnownCompletions, this is nil
    CEL *CELRule `json:"cel"`
}

type CELRule struct {
    Expressions []string `json:"expressions"`
    Message string `json:"message"`
    ....
}
```

bhperry: My original idea was that the standard message and condition for WellKnownCompletions could be overridden by the user, such that they could use the well known rules but set them to a custom condition/message.

But I do see the benefit to simplifying and using nesting for the CELRule. I am open to switching to this approach if that is what you prefer.

qiujian16: I prefer a simplified way, and if there is use case later that we need message override for WellKnownCompletions, we could add another field solely for WellKnownCompletions type to set them.

qiujian16: And it might be more consistent if the type is WellKnownCondition? so for completed condition, user should set
```
condition: Completed
type: WellKownCondition
```
It would be easier for us to extend later for other condition.

bhperry: > And it might be more consistent if the type is WellKnownCondition? so for completed condition, user should set
> 
> ```
> condition: Completed
> type: WellKownCondition
> ```
> 
> It would be easier for us to extend later for other condition.

This is interesting. My thinking with `WellKnownCompletions` was that since Completed is a specialized case, users would want to be able to set that separately from other future well known conditions.

Are you suggesting that a rule like this would select just a single rule that matches the given condition
```
- condition: Completed
  type: WellKnownCondition
```

and then a rule like this would select all well knowns?
```
- type: WellKnownCondition
```

qiujian16: I am thinking condition is the required field, you could have something like
```
- condition: Completed
   type: WellKnownCondition
- condition: Degraded
   type: WellKnownCondition
```
for a certain resource.

bhperry: Ok, that makes sense to me. It does make WellKnownCondition a little more difficult to use than WellKnownStatus since you need to know what conditions are well known, but I do like that it gives individual control over each one.

qiujian16: hrm I see what you mean. I think alternatively, we could have 
```
conditionRule:
  type: WellKnownConditions
```
and
```
conditionRule:
  type: CEL
  cel:
   - condition: Available
      expressions:
      - ...
      messsage: "some message"
  - condition: Completed
     expressions:
      - ...
      messsage: "some message"
```
this looks ok to me also.

bhperry: I think we do need to allow a mix of `WellKnownConditions` and custom CEL.

Also in this format we would be unable to use `listMapKey: condition` to merge multiple rules since they could each express multiple conditions. Although there's already that issue with WellKnownConditions rules being overridden by custom CEL, so the handling of multiple copies of the same condition may need to be custom anyway.

qiujian16: yeah, that make sense

bhperry: How do you feel about the idea I mentioned previously? Seems like it is a nice middle ground of making things easy by default but also allowing individual customization
```
- condition: Completed
  type: WellKnownCondition  # Single well known condition
```

```
- type: WellKnownCondition  # All well known conditions
```

qiujian16: I think making condition as a required field and mergeKey will make setting type only impossible. And I think we have to make condition as the mergekey for better merge strategy and ensure user does not input duplicated conditions

qiujian16: question is what happens when user set
```
- type: WellKnownCondition 
- type: CEL
   condition: Completed
   expressions: ....
```
or
```
- type: CEL
   condition: Completed
    expressions: ... 
- type: CEL
   condition: Completed
   expressions: ....
```

bhperry: Options:
1. Reject in webhook validation
2. The last occurrence of a condition overrides all previous
3. The full rule for the condition becomes the logical AND of all expressions across the rules

qiujian16: might work if add
```
// +listType=map
// +listMapKey=type
// +listMapKey=condition
```
such that there is no duplicate setting. but need take a try,

work/v1/types.go
bhperry: This is already set on the typedef https://github.com/open-cluster-management-io/api/pull/362/files#diff-22c62942813270e1c12182a840cacc1358496a602668f058ef6aff4d0febbceaR156-R157

There is a mix of styles for enums. Some set on type, some on usage. I based this off of `type FeedBackType string` so followed that style.

qiujian16:
/approve
only is to check if https://github.com/open-cluster-management-io/api/pull/362#discussion_r2095829420 works

work/v1/types.go
qiujian16: @bhperry would you like to add // +listMapKey=type 
and we are good to merge this.

bhperry: I don't think type should be a map key. Condition is required now, so using just that as map key ensures each condition only has one rule. If we add type then OCM has to handle the case where a condition has a WellKnown and custom CEL rule.

qiujian16: got it, it is fine if condition is required.

PR #363, opened 2025-03-26
✨add ClusterClaimConfiguration to API

o-farag:
## Summary
Enhances API to allow support for custom ClusterClaim configurations. 
## Related issue(s)

Required for https://github.com/open-cluster-management-io/ocm/issues/897


qiujian16:
I think this should under registrationConfiguration field

cc @elgnay

elgnay:
The structure looks good to me. As @qiujian16 suggested, please move `ClusterClaimConfiguration` under `registrationConfiguration`.

o-farag:
@qiujian16 may I get this reviewed? I'd like to merge it for https://github.com/open-cluster-management-io/ocm/pull/933

qiujian16:
https://github.com/open-cluster-management-io/api/pull/363#issuecomment-2788110590

o-farag:
@qiujian16 sorry, I didn't see your comments for some reason. Updated!

qiujian16:
/approve

looks good to me but I will let @elgnay to final review
/assign @elgnay

operator/v1/types_klusterlet.go
```diff
@@ -176,6 +176,24 @@ type RegistrationConfiguration struct {
 	// This provides driver details required to register with hub
 	// +optional
 	RegistrationDriver RegistrationDriver `json:"registrationDriver,omitempty"`
+
+	// ClusterClaimConfiguration represents the configuration of ClusterClaim
+	// Effective only when the `ClusterClaim` feature gate is enabled.
+	// +optional
+	ClusterClaimConfiguration *ClusterClaimConfiguration `json:"clusterClaimConfiguration,omitempty"`
```
zhujian7, line 183:
```suggestion
	ClusterClaims *ClusterClaimConfiguration `json:"clusterClaims,omitempty"`
```
nit, not sure if it is better/simpler to remove the suffix configuration and make it as `clusterClaim` or `clusterClaims`

operator/v1/types_klusterlet.go
```diff
@@ -176,6 +176,24 @@ type RegistrationConfiguration struct {
...
+type ClusterClaimConfiguration struct {
+	// Maximum number of custom ClusterClaims allowed.
+	// +kubebuilder:validation:Required
+	// +kubebuilder:default:=20
+	// +required
+	MaxCustomClusterClaims int32 `json:"maxCustomClusterClaims"`
+
+	// Custom suffixes for reserved ClusterClaims.
+	// +optional
+	ReservedClusterClaimSuffixes []string `json:"reservedClusterClaimSuffixes,omitempty"`
```
zhujian7, line 196: I do not quite understand why a customer needs to configure the reserved cluster claims?

o-farag: > I do not quite understand why a customer needs to configure the reserved cluster claims?

@zhujian7 the [proposed enhancement](https://github.com/open-cluster-management-io/ocm/issues/897) outlines this addition, I'm not sure if a specific customer requested it

operator/v1/types_klusterlet.go
o-farag: @elgnay what do you think?

operator/v1/types_klusterlet.go
elgnay: It is guaranteed that reserved ClusterClaims will be reported in the status of the ManagedCluster. However, custom ClusterClaims are not guaranteed to appear due to the limitation on the number of custom ClusterClaims, even if the user sets a relatively high limit.

operator/v1/types_klusterlet.go
elgnay: I'm good with the name `ClusterClaims`. @qiujian16 Is there any naming convention or best practice we should follow when defining the API?

operator/v1/types_klusterlet.go
elgnay: Moreover, ClusterClaims matching the `ReservedClusterClaimSuffixes` should be treated as additional reserved claims. They must be merged with, not override, the built-in (hardcoded) reserved claims when reported to the hub. Please either include this information in the field comment or consider renaming the field to `AdditionalReservedClusterClaimSuffixes` for clarity.

operator/v1/types_klusterlet.go
qiujian16: I kinda think it is a configuration, clusterClaims might be confusing on allowing user to specify claims they would like to create.

operator/v1/types_klusterlet.go
qiujian16: We could start with a predefined suffix. I do not know if there are use cases that need is to be configurable, and I am not sure what it should be when user change the suffix from one to another.

operator/v1/types_klusterlet.go
o-farag: I agree with quijian16, since this configures cluster claims and aren't actual clusterclaims

PR #365, opened 2025-03-31
:sparkles:  add celSelector in placement

haoqing0110:
## Related issue(s)

Fixes
https://github.com/open-cluster-management-io/ocm/issues/843 
https://github.com/open-cluster-management-io/ocm/issues/921


cluster/v1alpha1/types_addonplacementscore.go
```diff
@@ -58,6 +59,10 @@ type AddOnPlacementScoreItem struct {
 	// +kubebuilder:validation:Maximum:=100
 	// +required
 	Value int32 `json:"value"`
+
+	// Quantity is the original value of the score.
+	// +optional
+	Quantity *resource.Quantity `json:"quantity"`
```
qiujian16: I am thinking to not adding quantity for now. I am still concerned that too frequent update on this and potential overuse.

haoqing0110: yes, potential overuse would be an issue, remove this from PR.

PR #379, opened 2025-06-12
✨ Networking config for default mode webhooks

bhperry:
## Summary

API changes to support customizing webhook ports and hostNetwork in Default installation mode

## Related issue(s)

https://github.com/open-cluster-management-io/ocm/pull/1035

## Summary by CodeRabbit

- **New Features**
  - Added per-mode webhook settings for ClusterManager with distinct Default and Hosted configurations, including default-mode support.
  - Configurable bind options for registration and work webhooks: bind port, health probe port, metrics port, and hostNetwork; defaults applied (bind port 9443, health 8000, metrics 8080).

- **Documentation**
  - Clarified Hosted webhook port is the external port (default 443).

- **Tests**
  - Updated integration tests to validate Hosted-mode webhook configuration and port behaviors.


operator/v1/types_clustermanager.go
```diff
@@ -243,6 +254,33 @@ type HostedClusterManagerConfiguration struct {
...
+	// Healthchecks may be disabled by setting a value of "0" or "".
+	// +optional
+	// +kubebuilder:default=":8000"
+	HealthProbeBindAddress string `json:"healthProbeBindAddress"`
+
+	// MetricsBindAddress represents the metrics address of a webhook-server. The default value is ":8080"
+	// Metrics may be disabled by setting a value of "0" or "".
+	// +optional
+	// +kubebuilder:default=":8080"
+	MetricsBindAddress string `json:"metricsBindAddress"`
```
qiujian16: this is and health check are commonly the same address and port with different paths.

operator/v1/types_clustermanager.go
```diff
@@ -243,6 +254,33 @@ type HostedClusterManagerConfiguration struct {
...
+
+	// HostNetwork enables running webhook pods with hostNetwork: true
+	// This may be required in some installations, such as EKS with Calico CNI,
+	// to allow the API Server to communicate with the webhook pods.
+	// +optional
+	HostNetwork bool `json:"hostNetwork,omitempty"`
```
qiujian16, line 332: we do not set a boolean field commonly. It could be a enum string which is more extensible.

operator/v1/types_clustermanager.go
bhperry: I don't think controller-runtime supports that

operator/v1/types_clustermanager.go
bhperry: Even when it's a field that translates directly to another boolean field?

What would you suggest as an enum? Could do something like `Networking: Container/Host` maybe?

qiujian16: I see, we will use this to directly set to webhook pod. That is ok in this case.

operator/v1/types_clustermanager.go
qiujian16: I see, but do we need different addresses? Could we update exiting webhookConfiguration struct by adding healthinessPort and metricsPort into it?

operator/v1/types_clustermanager.go
bhperry: Ok great. Thanks for letting me know about the preference for enum, I will keep that in mind in future

operator/v1/types_clustermanager.go
bhperry: I opted to use `*BindAddress` because:
1. This is the format expected by controller-runtime. Kubebuilder generates CLI args in this same format as well.
2. When `HostNetwork: true` is set, it may be desirable to bind healthcheck or metrics to localhost in order to prevent exposing those listeners externally

I did also consider simply adding new fields to WebhhokConfiguration, but they don't apply to Hosted options so I did not think it made sense to have extra fields in the public facing API that can only be used in some cases.

qiujian16: hrm, I think it might. Is to expose the address so another apiserver could access it. In some case, hostnetwork might be needed. Actually hosted mode is not quite used anyway, I just feel we do not need another configuration for the webhook.

bhperry: My understanding of hosted mode was that it's a remote webhook not directly controlled by OCM. And the webhook configuration is just an address/port to reach it. So while metrics/health/hostNet might be used in that remote location, they would not be controlled by this API config.

Is that not correct?

qiujian16: it is not quite correct, the address/port is still the setting on the webhook server to be exposed. The hosted mode means the controller/webhook server is running in cluster1, while hub cluster is cluster2. So the hub cluster needs to wire to the webhook server in cluster1 with an accessible address/port

bhperry: But the webhook server in cluster1 is not actually controlled by the settings in WebhookConfiguration right? It just tells cluster2 how to reach the webhook. Which would mean hostNetwork and monitoring/healthcheck settings wouldn't do anything in Hosted mode.

qiujian16: I think hosted mode can also use hostNetwork, and if hostNetwork is set, the ports should be set on the webhook server.

bhperry: Maybe I don't understand hosted mode. Is it watching clustermanager config on the hub cluster?

qiujian16: hosted mode is like the controllers/operators are not running on the same cluster that host CRDs. One scenarios is user creates a kube controlplane (dedicated kube-controlplane) without any node in it for ocm purpose only. In such case, the controllers/operator needs to run in another kube cluster. Since crd is hosted in the dedicated kube-controlplane, and the webhook is run on the other kube cluster, the endpoint of the webhook needs to be exposed so the dedicated kubecontroplane can access. There are some options: 
1. the webhook service is exposed as loadbalancer service or ingress, with which the webhookconfigurations in the cluster manager needs to be set with correct address/port.
2. it can also be exposed as hostNetwork, and with that the webhookconfigurations in the cluster manager also needs to be

I think webhookconfiguration can be used in both mode, it just indicate how the network is set and how the address/port is exposed.

bhperry: Ok I see it now. When I was looking at it before I thought hosted mode for webhooks meant completely outside of kubernetes. Thanks for clarifying.

I've updated the API to use a shared WebhookConfiguration struct

bhperry: The one downside of shared WebhookConfiguration is there are a few differences between validation and defaults for default/hosted mode. 

- Address required only in hosted mode (and I'm not sure it would do anything in default mode)
  -  This can be marked required with `XValidation` rules in hosted mode only
- Port defaults to 9443 in default mode, and 443 in hosted mode.
  - Unable to use kubebuilder defaults for this, would have to be handled in OCM

operator/v1/types_clustermanager.go
```diff
@@ -232,32 +232,61 @@ const (
...
 // HostedClusterManagerConfiguration represents customized configurations we need to set for clustermanager in the Hosted mode.
 type HostedClusterManagerConfiguration struct {
 	// RegistrationWebhookConfiguration represents the customized webhook-server configuration of registration.
 	// +optional
+	// +kubebuilder:validation:XValidation:rule="self.address != ''",message="Address is required for hosted webhook configuration"
 	RegistrationWebhookConfiguration WebhookConfiguration `json:"registrationWebhookConfiguration,omitempty"`
 
 	// WorkWebhookConfiguration represents the customized webhook-server configuration of work.
 	// +optional
+	// +kubebuilder:validation:XValidation:rule="self.address != ''",message="Address is required for hosted webhook configuration"
```
qiujian16: it seems cel validation is not added in the crd definition? Is it because we are using a lower version of crd gen?

qiujian16 (commented):
/approve

only issue is cel validation is not added in the crd.

operator/v1/types_clustermanager.go
bhperry: @qiujian16 Pushed change to inline common settings in separate Default and Hosted configuration structs. I didn't like that Address was left unused in Default mode. Also we get to keep the simple required Address validation for hosted (instead of CEL) as well as different default ports for each. What do you think?

operator/v1/types_clustermanager.go
```diff
@@ -232,19 +232,62 @@ const (
...
+	RegistrationWebhookConfiguration HostedWebhookConfiguration `json:"registrationWebhookConfiguration,omitempty"`
 
 	// WorkWebhookConfiguration represents the customized webhook-server configuration of work.
 	// +optional
-	WorkWebhookConfiguration WebhookConfiguration `json:"workWebhookConfiguration,omitempty"`
+	WorkWebhookConfiguration HostedWebhookConfiguration `json:"workWebhookConfiguration,omitempty"`
 }
 
-// WebhookConfiguration has two properties: Address and Port.
+// WebhookConfiguration represents customization of webhook servers
```
qiujian16: actually it is not WebhookConfiguration, I think we could also use this for controller to set healthiness and metrics endpoint?

operator/v1/types_clustermanager.go
qiujian16: I think it is ok. We could have a better name for it I think.

operator/v1/types_clustermanager.go
bhperry: True. Maybe `BindConfiguration`?

Realized there is an issue for Hosted mode. The Port field is only the external endpoint for k8s API to reach the validating webhook, but if you set HostNetwork then you would also want to control the internal port bind to avoid port collision on the host net.

Perhaps a better API would be something like this:
```yaml
spec:
  ...
  deployOption:
    hosted:
      workWebhookConfiguration:
        port: 443
        address: work.webhook.example.com
        bindConfiguration:
          port: 19443
          healthProbeBindAddress: ":18000"
          metricsBindAddress: ":18080"
          hostNetwork: true
      registrationWebhookConfiguration:
        ...
    mode: Hosted

---

spec:
  ...
  deployOption:
    default:
      workWebhookConfiguration:
        bindConfiguration:
          ...
      registrationWebhookConfiguration:
        ...
    mode: Default
```

bindConfiguration would be optional and nullable

bhperry: It looks like controller servers work very differently from the webhooks. They use the `openshift/library-go` controller builder, which only takes a single listen address:port. Whereas the webhooks use `controller-runtime` which has separate binds for each. So I don't think it makes sense to re-use this configuration for controllers (unless they get migrated to controller-runtime in the future)

qiujian16: I think we do not need to consider how the implementation is when considering this API. The controller-runtime allows bind endpoint separately does not mean we need to support that in the API level. Is there scenario that we need to let metrics/healthiness bind to the different address?

bhperry: Well it's not just that it can bind separately, they are separate listeners. No option to share port with the webhook server. As a result, when using hostNetwork mode, it is important to have control over the binds.

qiujian16: I understand the port should be different, I am wondering if it is necessary to bind different HOST/IP for metrics/healthz endpoint.

bhperry: I see. I figured if you can bind hostNet then it is reasonable to allow fully configuring the bind for metrics/healthz. Not strictly necessary though (I had no intention of using it, I'm just binding ":<port>" for all of these).

qiujian16: right, I think it is sufficient to specify port for each

qiujian16:
@bhperry would you like to rebase the PR and update accordingly? I think it is ready to be merged in general

qiujian16:
verify fails, you may need to rerun the make update

PR #390, opened 2025-08-18
✨ Adding managed namespace field in the managedclusterset

qiujian16:
## Related issue(s)

enhancement proposal https://github.com/open-cluster-management-io/enhancements/pull/147

## Summary by CodeRabbit

- **New Features**
  - ManagedClusterSet spec can declare managed namespaces via a new managedNamespaces list (DNS-1123 names, max length 63).
  - ManagedCluster status now reports managedNamespaces (clusterSet-scoped namespace entries with conditions).
  - New namespace configuration types expose per-namespace and cluster-set scoped settings and status.

- **Chores**
  - Removed a debug print to reduce console noise in a maintenance script.


cluster/v1/types.go
```diff
@@ -188,6 +188,23 @@ type ManagedClusterClaim struct {
...
 }
 
+// managedNamespaces defines a namespace on the managedclusters across the
+// clusterset to be managed by this clusterset.
+type ManagedNamespaceConfig struct {
+	// name is the name of the namespace.
+	// +required
+	// +kubebuilder:validation:MaxLength=63
+	// +kubebuilder:validation:Pattern=^[a-z0-9]([-a-z0-9]*[a-z0-9])?$
+	Name string `json:"name"`
```
elgnay, line 206: Is there any restriction on the name? for example, it cannot be one of 'kube-system', 'kube-public' and 'default'

cluster/v1/types.go
```diff
@@ -188,6 +188,23 @@ type ManagedClusterClaim struct {
...
+}
+
+type ClusterSetManagedNamespaceConfig struct {
```
elgnay, line 209: Has this struct been used anywhere?

qiujian16: updated, should be in the cluster status

cluster/v1/types.go
```diff
@@ -144,26 +144,33 @@ const (
...
 	// cluster. The set of claims is not uniform across a fleet, some claims can be
 	// vendor or version specific and may not be included from all managed clusters.
 	// +optional
 	ClusterClaims []ManagedClusterClaim `json:"clusterClaims,omitempty"`
+
+	// managedNamespaces are a list of namespaces managed by the clustersets the
+	// cluster belongs to.
+	// +optional
+	// +listType=map
+	// +listMapKey=clusterSet
```
elgnay, line 172: Since the 'clusterSet' is the key of the map, does that mean for each clusterset only one item is allowed in the list? in another world, only one managed namespace is allowed for a clusterset in the 'managedNamespaces' list?

qiujian16: correct, add name is another key

PR #391, opened 2025-08-26
:seedling: add auth type related consts

skeeey:
## Related issue(s)

Fixes #

## Summary by CodeRabbit

* Chores
  * Standardized authentication identifiers: AWS IRSA, CSR, and gRPC for consistent configuration.
  * Added CSR username metadata key to improve traceability of requests.
  * Added gRPC signing metadata and extended registration validation to support gRPC auth, improving reliability of cluster registration flows and future integrations.

### DIFF
--- a/operator/v1/types_clustermanager.go
+++ b/operator/v1/types_clustermanager.go
@@ -118,6 +118,25 @@ type RegistrationHubConfiguration struct {
 	RegistrationDrivers []RegistrationDriverHub `json:"registrationDrivers,omitempty"`
 }
 
+const (
+	// AwsIrsaAuthType represents the authentication type that uses AWS IRSA
+	AwsIrsaAuthType = "awsirsa"
+	// CSRAuthType represents the authentication type that uses Kubernetes CSR
+	CSRAuthType = "csr"
+	// GRPCAuthType represents the authentication type that uses gRPC.
+	GRPCAuthType = "grpc"
+)
+
+// GRPCAuthSigner is the signer name used when creating CSRs for gRPC authentication.
+const GRPCAuthSigner = "open-cluster-management.io/grpc"
+
+const (
+	// CSRUsernameAnnotation is added to a CSR to identify the user who requested the CSR.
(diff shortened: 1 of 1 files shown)


operator/v1/types_clustermanager.go
```diff
@@ -118,6 +118,26 @@ type RegistrationHubConfiguration struct {
...
+
+// GRPCAuthSigner is the signer name used when creating CSRs for gRPC authentication.
+const GRPCAuthSigner = "open-cluster-management.io/grpc"
+
+const (
+	// CSRUsernameAnnotation is added to a CSR to identify the user who requested the CSR.
+	CSRUsernameAnnotation = "open-cluster-management.io/csr/user"
+
+	// CSRAddOnAnnotation is added to a CSR to identify the name of the addon  that requested the CSR.
+	CSRAddOnNameAnnotation = "open-cluster-management.io/csr/addon"
```
qiujian16: I think this has been added on the label.

skeeey: we have this on the label, so we can 
- check whether the username is from the gRPC server.
- ~check that the `addon.Name` matches the CSR label.~ (we already did this)
- check the crs user annotation has the expected prefix

operator/v1/types_clustermanager.go
```diff
@@ -118,6 +118,23 @@ type RegistrationHubConfiguration struct {
...
+	CSRAuthType = "csr"
+	// GRPCAuthType represents the authentication type that uses gRPC.
+	GRPCAuthType = "grpc"
+)
+
+// GRPCAuthSigner is the signer name used when creating CSRs for gRPC authentication.
+const GRPCAuthSigner = "open-cluster-management.io/grpc"
+
+const (
+	// CSRUsernameAnnotation is added to a CSR to identify the user who requested the CSR.
```
qiujian16, line 134: let be add that this should only be honored when registration driver is grpc and the csr user name is service account of grpc server.

skeeey: updated

PR #394, opened 2025-09-08
✨ Add CleanUpCompletedManifestwork feature gate

qiujian16:
## Related issue(s)

Fixes #

## Summary by CodeRabbit

- New Features
  - Added an optional feature flag "CleanUpCompletedManifestWork" (Alpha, disabled by default) to enable automatic cleanup of completed ManifestWorks when enabled.

- Style
  - Minor formatting adjustments to feature gate definitions; no functional impact.

### DIFF
--- a/feature/feature.go
+++ b/feature/feature.go
@@ -88,6 +88,11 @@ const (
 
 	// ClusterImporter will enable the auto import of managed cluster for certain cluster providers, e.g. cluster-api.
 	ClusterImporter featuregate.Feature = "ClusterImporter"
+
+	// CleanUpCompletedManifestWork will delete manifestworks which have Completed status after a specified TTL seconds.
+	// When enabled, the work controller will automatically clean up completed manifest works based on the configured
+	// time-to-live duration to prevent accumulation of old completed resources.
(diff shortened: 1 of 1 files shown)


feature/feature.go
```diff
@@ -88,6 +88,11 @@ const (
 
 	// ClusterImporter will enable the auto import of managed cluster for certain cluster providers, e.g. cluster-api.
 	ClusterImporter featuregate.Feature = "ClusterImporter"
+
+	// CleanUpCompletedManifestWork will delete manifestworks which have Completed status after a specified TTL seconds.
+	// When enabled, the work controller will automatically clean up completed manifest works based on the configured
+	// time-to-live duration to prevent accumulation of old completed resources.
+	CleanUpCompletedManifestWork featuregate.Feature = "CleanUpCompletedManifestWork"
```
bhperry, line 95: Why does this need to be a feature flag? The cleanup behavior is already optional with the TTL on the manifestwork itself.

qiujian16: it will need to start a controller on the hub which does not start by default today. This is to ensure that it will not bring a new component to the existing environment.

bhperry: ah gotcha

zhujian7:
LGTM

PR #427, opened 2026-03-11
:seedling:  Add addon annotation key and value to beta1

haoqing0110:
## Related issue(s)

Fixes #

## Summary by CodeRabbit

* **Bug Fixes**
  * Added support for install namespace configuration via annotation in managed cluster addons, providing a migration path for previous installation methods.

### DIFF
--- a/addon/v1beta1/types_managedclusteraddon.go
+++ b/addon/v1beta1/types_managedclusteraddon.go
@@ -457,3 +457,9 @@ const (
 	// successful to set up rbac for the addon agent.
 	RegistrationAppliedSetPermissionApplied = "SetPermissionApplied"
 )
+
+const (
+	// InstallNamespaceAnnotation is the annotation key for storing installNamespace
+	// This is used because installNamespace field was removed in v1beta1
+	InstallNamespaceAnnotation = "addon.open-cluster-management.io/v1alpha1-install-namespace"
+)


addon/v1beta1/types_clustermanagementaddon.go
```diff
@@ -221,3 +221,18 @@ type ClusterManagementAddOnList struct {
 	// Items is a list of cluster management add-ons.
 	Items []ClusterManagementAddOn `json:"items"`
 }
+
+const (
+	// AddonLifecycleAnnotationKey is an annotation key on ClusterManagementAddon to indicate the installation
+	// and upgrade of addon should be handled by the general addon manager or addon itself. The valid values are
+	// addon-manager and self. If the annotation is not set, addon lifecycle is handled by addon itself.
+	AddonLifecycleAnnotationKey = "addon.open-cluster-management.io/lifecycle"
```
qiujian16: do we still nee this?

haoqing0110: This annotation is still used to check if CMA should be managed by addon-manager. If the annotation value is "self",  addon-manager won't handle it. 
https://github.com/open-cluster-management-io/addon-framework/blob/main/pkg/utils/helpers.go#L253
https://github.com/open-cluster-management-io/ocm/blob/main/pkg/addon/manager.go#L166

qiujian16: I mean shouldn't all addon move to managed by addon-manager in beta?

haoqing0110: Yes, we can remove this in beta and claim [supported-version](https://open-cluster-management.io/docs/developer-guides/addon/#supported-version) "The OCM [latest version] requires an addon developed with addon-framework v0.10.0 and above versions."

haoqing0110: Most addons are developed with at least v0.12.0 today.

haoqing0110: Log this issue to remove the annotation check in framework and ocm
https://github.com/open-cluster-management-io/ocm/issues/1428

qiujian16: should we just use alpaha annotation, instead of adding this to beta?

PR #430, opened 2026-03-17
🌱 Remove unsupported addon webhook hosted mode config

bhperry:
## Summary

Hosted mode is not supported for addon webhook

## Related issue(s)

https://github.com/open-cluster-management-io/ocm/pull/1420#discussion_r2944104057
https://github.com/open-cluster-management-io/ocm/issues/1442

## Summary by CodeRabbit

* **Chores**
  * Removed addon webhook configuration from Hosted mode ClusterManager. The addonWebhookConfiguration option and its associated settings (address, port configuration, health probes, and network parameters) are no longer available. This simplification reduces the configuration surface for Hosted deployments while other webhook options remain intact.

### DIFF
--- a/operator/v1/types_clustermanager.go
+++ b/operator/v1/types_clustermanager.go
@@ -402,10 +402,6 @@ type HostedClusterManagerConfiguration struct {
 	// WorkWebhookConfiguration represents the customized webhook-server configuration of work.
 	// +optional
 	WorkWebhookConfiguration HostedWebhookConfiguration `json:"workWebhookConfiguration,omitempty"`
-
-	// AddonWebhookConfiguration represents the customized webhook-server configuration of addons.
-	// +optional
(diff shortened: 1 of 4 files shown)


zhiweiyin318:
Hi @bhperry  thanks for your quick PR. 
I think we should mark the fields as **deprecated** instead of removing them, since the clusterManager API is already v1.

operator/v1/types_clustermanager.go
zhiweiyin318, line 397: I think we need to deprecate the whole HostedClusterManagerConfiguration

qiujian16: let's do it in separated PR. We cannot remove other field directly. but it is ok to remove this field.

bhperry: Yes I was just removing the hosted fields that I added in the other PR

PR #435, opened 2026-04-19
:bug: fix: omit PlacementDecision score field when zero-valued

mikeshng:
## Summary

Add omitzero json tag to ClusterDecision.Score so that the default zero value is not serialized to JSON. The field remains int64 (non-pointer) and is only included when explicitly set to a non-zero value.

## Related issue(s)

Fixes #

## Summary by CodeRabbit

* **Refactor**
  * Optimized JSON serialization to omit zero score values in cluster decision responses, reducing payload size.

### DIFF
--- a/cluster/v1beta1/types_placementdecision.go
+++ b/cluster/v1beta1/types_placementdecision.go
@@ -61,7 +61,7 @@ type ClusterDecision struct {
 
 	// Score is the computed score for the cluster based on configured prioritizers
 	// +optional
-	Score int64 `json:"score"`
+	Score int64 `json:"score,omitzero"`
 }
 
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object


mikeshng:
Discussed with original enhancement author @bhperry and he agreed this is the right approach. 

/assign @qiujian16

bhperry:
I misunderstood the problem. Does this actually fix https://github.com/argoproj/argo-cd/issues/27264 ? It seems like argo actually needs to handle non-strings or it will still break once the PR to set thr scores merges.

Score should always be set once the feature is complete.

bhperry:
@mikeshng would it be enough to just set omitempty to make argocd not break while they fix the underlying problem?

I think having it be a pointer is confusing for the api, when the intention is that it will always be set to a number.

mikeshng:
This doesn't directly fix the argo-cd issue. The argo-cd fix is still required. This current API change PR gives us enough time to address it properly.

qiujian16:
can we set omitzero instead? will it be good enough for serielization？

mikeshng:
@bhperry @qiujian16 thank you for the quick reviews. I've done the changes as you both suggested. Thanks!

qiujian16:
think about it more. I wonder if "not set" is meaningful when the feature is enabled. I guess not, when feature is enabled and score is zero, it does not mean the scheduler does not set the value, but just indicate the score is zero.

mikeshng:
@bhperry , could you please see @qiujian16 's comment? I think both me and him prefer if we use a pointer here. Thanks.

bhperry:
Once the feature in OCM is merged (I will rebase the [PR](https://github.com/open-cluster-management-io/ocm/pull/1436) tomorrow) score will never be "not set". When cluster decision is first created it has its initial score. That is why I don't think a pointer makes sense -- it will always be set to an integer, so it's an extra pointer dereference + nil check for no reason.

In my opinion the best way forward would be:
- Set omitzero as a temp fix for argo
- Wait for argo fix to be merged, then merge the scoring feature in OCM
- Remove omitzero from the API (as you mentioned @qiujian16 a score of zero means explicitly set to zero, so it should be included in serialized output)

qiujian16:
ok makes sense to me. 

/approve
/lgtm

PR #437, opened 2026-05-01
:sparkles: feat: add "random" placement prioritizer

dhaiducek:
## Summary

Add "random" placement prioritizer to Placement API description.

## Related issue(s)

- https://github.com/open-cluster-management-io/ocm/issues/1507
- https://github.com/open-cluster-management-io/ocm/pull/1506

## Summary by CodeRabbit

* **New Features**
  * Added Random prioritizer option for cluster placement. Users can now randomly select clusters when deploying workloads, providing an alternative distribution strategy for flexible resource management.


mikeshng (commented):
Would introducing randomness lead to unnecessary write operations and increased churn (ie excessive updates stored in etcd)?

I am assuming the list is no longer stabilized? Maybe the comment needs to explain that a bit futher.

PR #441, opened 2026-06-11
:seedling: chore: upgrade to go 1.26

tesshuflower:
Upgrades Go version from 1.25 to 1.26.

Changes:
- `go.mod` — go directive updated to 1.26.0
- `.github/workflows/go.yml` — GO_VERSION updated to 1.26
- `Dockerfile.build` — GO_VERSION updated to 1.26.0

Relates to: https://github.com/open-cluster-management-io/ocm/issues/1555

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -25,12 +25,12 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: checkout code
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
         with:
           fetch-depth: 1
           path: go/src/open-cluster-management.io/api
       - name: install Go
-        uses: actions/setup-go@v3
+        uses: actions/setup-go@v5
         with:
           go-version: ${{ env.GO_VERSION }}
       - name: verify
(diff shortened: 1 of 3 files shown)


dhaiducek (approved):
LGTM! Thanks @tesshuflower!

mikeshng:
@coderabbitai review

dhaiducek (changes requested):
I personally don't like a blanket commit that says "address suggestions". If you agree, could you make it more specific? In this case, probably separate it into two commits.

tesshuflower:
> I personally don't like a blanket commit that says "address suggestions". If you agree, could you make it more specific? In this case, probably separate it into two commits.

I guess I was trying to avoid too many commits so grouped them together? the commit details do explain exactly what the changes were - but you're right, it's not on the summary line.

I've updated to split into 2 commits so easier to see in the 1 ilner what the changes are.

dhaiducek (approved):
Thanks! Right, the summary/title is what I was poking at--your descriptions in the body were definitely spot on. Though now I see PRs are Squash merged rather than rebased, so maybe it doesn't matter quite so much since it all gets dumped into the body anyway. 😬

PR #181, opened 2022-09-30
Add spread policy

huaouo:
This PR is related to https://github.com/open-cluster-management-io/enhancements/pull/70.

PR #444, opened 2026-07-18
✨ feature: migrate structural validation to kubebuilder CEL markers (Phase 1)

Vaishnav88sk:
## Summary
This PR implements **Phase 1** of the CEL Validation Migration as outlined and approved in [Enhancement Proposal #181](https://github.com/open-cluster-management-io/enhancements/pull/181). 

It replaces legacy Go-based structural webhook validation with native Kubernetes `+kubebuilder:validation:XValidation` CEL markers directly on the Custom Resource structs.

**Validations Migrated:**
- `ManagedCluster`: Enforced HTTPS format for `Spec.ManagedClusterClientConfigs` URLs and regex format for `metadata.name`.
- `ManagedClusterSetBinding`: Enforced cross-field parity (`metadata.name == spec.clusterSet`).
- `ManifestWork`: Enforced `manifests` array is not empty.

*Note: The corresponding `ValidatingAdmissionPolicy` additions and FeatureGate operator updates (Phase 2) will be submitted in a follow-up PR to the `ocm` repository once this API change is merged.*

## Related issue(s)

Fixes phase 1 of https://github.com/open-cluster-management-io/ocm/issues/1566

## Summary by CodeRabbit

* **New Features**
  * Added an Alpha `CELValidation` feature gate for native Kubernetes API validation.

* **Bug Fixes**
  * Enforced non-empty manifests for add-on templates, manifest works, and manifest replica sets.
  * Strengthened `ManagedCluster` validation for names, client configuration limits, and secure HTTPS URLs.
  * Ensured `ManagedClusterSetBinding` names match the specified cluster set.

* **Tests**
  * Added integration coverage for valid and invalid `ManagedCluster` names, client URLs, and configuration limits.

### DIFF
--- a/feature/feature.go
+++ b/feature/feature.go
@@ -104,6 +104,10 @@ const (
 	// When enabled, a debug-server container will be added to the placement pod, providing
 	// /debug/placements/* endpoints for placement scheduling simulation and debugging.
 	PlacementDebugServer featuregate.Feature = "PlacementDebugServer"
+
+	// CELValidation replaces the legacy Go Validating Webhooks with native Kubernetes
+	// CEL validation rules and ValidatingAdmissionPolicy bindings.
+	CELValidation featuregate.Feature = "CELValidation"
 )
 
 // DefaultSpokeRegistrationFeatureGates consists of all known ocm-registration
@@ -127,6 +131,7 @@ var DefaultHubRegistrationFeatureGates = map[featuregate.Feature]featuregate.Fea
 	ResourceCleanup:            {Default: true, PreRelease: featuregate.Beta},
 	ClusterProfile:             {Default: false, PreRelease: featuregate.Alpha},
 	ClusterImporter:            {Default: false, PreRelease: featuregate.Alpha},
+	CELValidation:              {Default: false, PreRelease: featuregate.Alpha},
 }
 
 var DefaultHubAddonManagerFeatureGates = map[featuregate.Feature]featuregate.FeatureSpec{
@@ -147,6 +152,7 @@ var DefaultHubWorkFeatureGates = map[featuregate.Feature]featuregate.FeatureSpec
 	ManifestWorkReplicaSet:       {Default: false, PreRelease: featuregate.Alpha},
 	CloudEventsDrivers:           {Default: false, PreRelease: featuregate.Alpha},
 	CleanUpCompletedManifestWork: {Default: false, PreRelease: featuregate.Alpha},
+	CELValidation:                {Default: false, PreRelease: featuregate.Alpha},
 }
 
 // DefaultSpokeWorkFeatureGates consists of all known ocm work feature keys for work agent.
--- a/work/v1/types.go
+++ b/work/v1/types.go
@@ -67,6 +67,7 @@ type Manifest struct {
 type ManifestsTemplate struct {
 	// manifests represents a list of kubernetes resources to be deployed on a managed cluster.
 	// +optional
+	// +kubebuilder:validation:XValidation:rule="size(self) > 0",message="manifests should not be empty"
 	Manifests []Manifest `json:"manifests,omitempty"`
 }
 
--- a/addon/v1alpha1/0000_03_addon.open-cluster-management.io_addontemplates.crd.yaml
+++ b/addon/v1alpha1/0000_03_addon.open-cluster-management.io_addontemplates.crd.yaml
@@ -424,6 +424,9 @@ spec:
                           x-kubernetes-embedded-resource: true
                           x-kubernetes-preserve-unknown-fields: true
                         type: array
+                        x-kubernetes-validations:
+                        - message: manifests should not be empty
+                          rule: size(self) > 0
                     type: object
                 type: object
               registration:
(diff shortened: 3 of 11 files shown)


work/v1/types.go
```diff
@@ -66,8 +66,9 @@ type Manifest struct {
 // ManifestsTemplate represents the manifest workload to be deployed on a managed cluster.
 type ManifestsTemplate struct {
 	// manifests represents a list of kubernetes resources to be deployed on a managed cluster.
-	// +optional
-	Manifests []Manifest `json:"manifests,omitempty"`
+	// +required
```
mikeshng: This is an API change that is no longer about using CEL as validation. I wonder if this is truly needed?

qiujian16:
please add integration tests to test all the validations.

work/v1/types.go
Vaishnav88sk: Resolved

feature/feature.go
```diff
@@ -104,6 +104,10 @@ const (
 	// When enabled, a debug-server container will be added to the placement pod, providing
 	// /debug/placements/* endpoints for placement scheduling simulation and debugging.
 	PlacementDebugServer featuregate.Feature = "PlacementDebugServer"
+
+	// CELValidation replaces the legacy Go Validating Webhooks with native Kubernetes
+	// CEL validation rules and ValidatingAdmissionPolicy bindings.
+	CELValidation featuregate.Feature = "CELValidation"
```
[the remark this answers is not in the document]
Vaishnav88sk: We can't gate the CRD rollout in this repository because `open-cluster-management-io/api` only generates the static CRD YAMLs via Kubebuilder. The actual runtime logic to conditionally strip `x-kubernetes-validations` and manage webhooks based on the `CELValidation` feature gate is handled downstream by the installer operators (like `registration-operator`), which consume this flag.

Vaishnav88sk:
@mikeshng @qiujian16 This has been pending for 2 mo. Any chnages in the code?

mikeshng (commented):
> @mikeshng @qiujian16 This has been pending for 2 mo. Any chnages in the code?

Sorry, this could be easily a production breaking change that's why we are being more careful with this change. 

I am looking at this PR right now and it looks like the CRD change will take effect immediately and there is no way of turning it off. So I don't really understand the feature gate. Is that a true statement?

Vaishnav88sk:
> Sorry, this could be easily a production breaking change that's why we are being more careful with this change.
> 
> I am looking at this PR right now and it looks like the CRD change will take effect immediately and there is no way of turning it off. So I don't really understand the feature gate. Is that a true statement?

Agreed. However, this is the intended behavior as outlined and approved in Enhancement Proposal https://github.com/open-cluster-management-io/enhancements/pull/181.

Specifically, in the Phase 4: The Feature Gate section of the EP, it explicitly notes: "CRD-level CEL markers will always be present, as they are compiled into the CRD."

The `CELValidation` feature gate is designed only to govern the deployment of `ValidatingAdmissionPolicy` resources (for complex/authorization checks) and to toggle the registration of the legacy Go `ValidatingWebhookConfiguration` downstream in the operators. Structural CRD CEL markers (`x-kubernetes-validations`) are meant to be statically compiled and always active, aligning with standard Kubebuilder and Kubernetes best practices for CEL migration.

If a cluster administrator truly needs to disable these structural rules, they would need to manually strip the `x-kubernetes-validations` block from the CRD before applying, but from the OCM component perspective, shipping the complete schema is the intended design.

*(Text formatted by AI as my English not that good. 🙂)*

mikeshng (commented):
> Agreed. However, this is the intended behavior as outlined and approved in Enhancement Proposal https://github.com/open-cluster-management-io/enhancements/pull/181.

Yes. I think I over looked this critical issue.

It will be great if we can somehow gate the CRD changes as well. The manual edit afterwards most likely won't work because the clustermanager and klusterlet operator will just revert it.

I am going to ask my colleagues to chime in. I am not ready to go ahead with this one yet.

/hold

CC @tesshuflower @rokej @dhaiducek

Vaishnav88sk:
> It will be great if we can somehow gate the CRD changes as well. The manual edit afterwards most likely won't work because the clustermanager and klusterlet operator will just revert it.

Hi @mikeshng, You're exactly right.

To prevent this, the gating logic will be implemented directly inside the `clustermanager` and `klusterlet` operators. Before applying the CRD manifests, the operators will check the `CELValidation` feature gate:

* **If `CELValidation=true`:** Apply the CRD as-is.
* **If `CELValidation=false`:** Programmatically strip the `x-kubernetes-validations` block from the in-memory CRD object *before* reconciling it against the cluster.

This guarantees the operators will respect the feature gate and won't force CEL rules onto the cluster when disabled. We can coordinate merging this PR with the companion operator PRs to ensure a safe rollout.

I'll wait for the rest of the team to chime in and review before we proceed!

mikeshng (commented):
@Vaishnav88sk Yes, that sounds much safer. I want to enhance your logic a bit: instead of gate==false, strip CEL, it should be gate nil or gate==false the default CRD doesn't have the CEL at all. If the gate is present and ==true then insert the CEL validations. That way is even safer and it matches the feature gate naming to the behavior.

PR #450, opened 2026-09-18
:sparkles: feat: add replicaConfigs to AddOnDeploymentConfig Allow typed int32

amancse:
add replicaConfigs to AddOnDeploymentConfig Allow typed integer replica overrides for addon workloads. 
Supports Deployment/StatefulSet workloadIDs with optional wildcards.

Related: open-cluster-management-io/ocm#1695

## Related issue(s)

Fixes #

## Summary by CodeRabbit

* **New Features**
  * Added support for configuring desired replica counts for add-on agent workloads.
  * Replica settings can target deployments and statefulsets using workload IDs.
  * Wildcards are supported for flexible workload matching.
  * Replica counts must be zero or greater.
  * When multiple settings match a workload, the last matching entry applies.
  * Available in both v1alpha1 and v1beta1 configurations.
  * Replica configuration settings are optional and can be specified as an ordered list.


addon/v1alpha1/0000_02_addon.open-cluster-management.io_addondeploymentconfigs.crd.yaml
```diff
@@ -181,6 +181,38 @@ spec:
...
+                      pattern: ^(deployments|statefulsets|\*):.+$
+                      type: string
+                  required:
+                  - replicas
+                  - workloadID
+                  type: object
+                type: array
+                x-kubernetes-list-map-keys:
+                - workloadID
+                x-kubernetes-list-type: map
```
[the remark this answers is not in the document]
amancse: addressed

addon/v1alpha1/types_addondeploymentconfig.go
```diff
@@ -69,6 +69,33 @@ type AddOnDeploymentConfigSpec struct {
...
+	// +required
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:Pattern=`^(deployments|statefulsets|\*):.+$`
+	WorkloadID string `json:"workloadID"`
+
+	// Replicas is the desired number of replicas.
+	// +required
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:Minimum=0
+	Replicas int32 `json:"replicas"`
```
mikeshng, line 98: As I mentioned on Tuesday's meeting, I like things that are following existing convention or prior arts so we are not re-inventing the wheels and gives me confidence to know the change is the right approach. Please explain where you derive this idea from, what existing convention you are following, why just this field is enough and not more fields, why we have to define our own int32 fields and not use something from k8s core api, etc etc.

Also explain how is backward compatible, migration not needed, no risk to breaking anything existing to upstream OCM-io and downstream consumers (not just limited to RHACM but all the other consumers that might not be watching the code changes and will upgrade). 

Note: my comment is not just about this line but the whole change in general. Cheers.

amancse: **The problem is:**

customizedVariables only supports string values
Deployment.spec.replicas requires int32
String substitution (replicas: "{{.replicas}}") produces invalid YAML - the quotes make it a string, not an integer.

**Why just this field:**
replicas is the only commonly-needed int32 field that customizedVariables (string-only) cannot handle.

**Why not k8s core types:**
No k8s type exists for "workload → replica count" mapping; same reason AODC defines its own ContainerResourceRequirements.

**Backward compatible:**
Yes, +optional field with omitempty; empty/missing replicaConfigs triggers no code path changes.

**Migration needed:**
None, existing AODC CRs work unchanged; users opt-in when needed.

**Risk to consumers:**
Zero, purely additive, no existing fields modified, no behavior change unless replicaConfigs is explicitly used.

mikeshng: You just said the problem is `customizedVariables only supports string values` so the natural thinking is get customizedVariables to support int32 value right? 

Explain how come the API change is not at customizedVariables level (or close to where that is) where it's pointing at the specific KEY to set the value instead now it's at the top level spec where you have to use a "workload ID" to point at it. 

As I mentioned, please how me the prior arts or the convention where this change should be done this way.

amancse: The template substitution is fundamentally string-based.

Even if we added intValue to CustomizedVariable:
```yaml
customizedVariables:
  - name: replicas
    intValue: 3  #  int32 in API
```
The template rendering uses **fasttemplate** which does string substitution:
```yaml
# Template
replicas: {{.replicas}}
# After substitution (still a string in YAML context!)
replicas: 3  #  **looks like int, but fasttemplate just inserted "3" as text**
```
YAML parsing happens BEFORE we can apply types. The template engine doesn't understand YAML structure it just does text replacement.

ans to your question: **prior arts or the convention where this change should be done this way.**
**resourceRequirements works**, it also bypasses customizedVariables:

```yaml
spec:
  customizedVariables: [...]     # <- string substitution
  resourceRequirements:          # <- separate typed field
    - containerID: "deployments:foo:container"
      requests:
        memory: "64Mi"           # <- typed ResourceList
```
It uses **containerID** to identify targets, not template keys

mikeshng: resourceRequirements is a much bigger set vs replicas that is for 1 field.

I don't buy the int32 is the only field that matters argument because addonTemplate can include any CR and any CRDs can define their fields to have other non string values as well. If we are going to take special handling for int32 what about all the other different types? 

> YAML parsing happens BEFORE we can apply types. The template engine doesn't understand YAML structure it just does text replacement.

So why can't we consider post processing? I rather do that in the code and it's much user friendly as well. We KNOW for `replicas` specifically its an int so just strip the quotes. Because if we go the path of adding this special case for int32, then later another type then later another type when does it end?

mikeshng: I think we need to come up with a solution to address all types and not just in32. That's an acceptable API change. 

OR

Since your proposal is "hardcoding" the replicas anyway then just do the processing in the code. Whether that is stripping the quotes like you said or something else then no API change and easy to use for users.

mikeshng: Does the `boolean` have this template problem as well? That is a fairly common field and if we can't support that, it's an issue as well. So I rather we address them all.

mikeshng: @amancse let's step back and don't even think about OCM for a second let's just talk about templating. 
If we provide a system for the user to do templating, the user comes in and do something like t his

```
replicas: {{ replicasCount }}

replicasCount = 3
```

This some how doesn't work for whatever reason. 
Did the user did something wrong? Maybe? Maybe needed the extra casting like `replicas: {{ replicasCount | int }}` or whatever but it's hard to argue that the user shouldn't be allow to do this.

So if the user did the right thing, it's just our code can't process it properly.
By making some new additional API change just to support this case, we are basically saying, hey user you did the wrong thing, you actually have to do this instead. Which I don't think it's the right call.

